PR #4, opened 2019-04-15
3D scene for subsurface data

HansKallekleiv:
Relates to equinor/webviz-subsurface#10


HansKallekleiv:
Closing this for now. Can revisit if/when we want to implement 3D views.

PR #7, opened 2019-05-28
Reduce data size Map component

anders-kiaer:
Resolves https://github.com/equinor/webviz-subsurface/issues/17 on the component side.

The final data reduction achieved (on a sem-realistic benchmark model) is **82%**. There are reasons to believe comparable reduction in size will be the case also for real models.

**Approach:**
- Coordinate values have been changed from using nanometer (!) resolution to something more practical for subsurface purposes: Coordinate values are scaled to the range 0-1000, then converted to integers before passed to a JSON object. The linear conversion back to "physical" values is done by passing the corresponding two numbers, scale and offset (`ax + b`), to the `d3`-component. Same principle done for the map grid cell values (and flow).
- Change from dictionary with grid cell indices as keys (which are not used anyway) to array.


examples/example_subsurface_map.py
```diff
@@ -0,0 +1,17 @@
+import json
+import pandas as pd
```
HansKallekleiv: Pandas can be removed as it is not used

PR #36, opened 2019-09-09
Make leaflet draw optional

### LINKED ISSUE #34: Leaflet-draw toolbar and buttons should be optional
Currently the user can edit markers, polylines and polygons.
Often one or none of these are relevant for the visualization.
=>
- [ ] Set available edit options as props (default no options?)

- [ ] Remove toolbar if no options are active

HansKallekleiv:
Closes #34 

Added 3  new bool props to LayeredMap

`draw_toolbar_marker`, `draw_toolbar_polyline`, `draw_toolbar_polygon`.

When all are `false` the Leaflet Draw toolbar is not shown.

### DIFF
--- a/src/lib/components/LayeredMap.js
+++ b/src/lib/components/LayeredMap.js
@@ -25,8 +25,8 @@ class LayeredMap extends Component {
     }
 
     render() {
-        const {draw_marker, draw_polygon, draw_polyline, setProps} = this.props
-        const showDrawControls = (draw_marker || draw_polygon || draw_polyline) ? true : false
+        const {draw_toolbar_marker, draw_toolbar_polygon, draw_toolbar_polyline, setProps} = this.props
+        const showDrawControls = (draw_toolbar_marker || draw_toolbar_polygon || draw_toolbar_polyline) ? true : false
         return (
                 <Map id={this.props.id} style={{height: this.props.height}}
                      ref={this.mapRef}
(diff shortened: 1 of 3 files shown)


src/lib/components/LayeredMap.js
```diff
@@ -47,20 +49,28 @@ class LayeredMap extends Component {
...
-                    </FeatureGroup>
+                    { showDrawControls  && (
+                        <FeatureGroup>
+                            <DrawControls
+                                showMarker={show_marker}
+                                showPolygon={show_polygon}
+                                showPolyline={show_polyline}
+                                lineCoords={(coords) => setProps({'line_points': coords})}
+                                markerCoords={(coords) => setProps({'marker_point': coords})}
+                                polygonCoords={(coords) => setProps({'polygon_points': coords})} />
```
anders-kiaer: Put /> on separate line and same indent as opening tag.

src/lib/components/LayeredMap.js
```diff
@@ -47,20 +49,28 @@ class LayeredMap extends Component {
...
+                                </FeatureGroup>
```
anders-kiaer: Same indent as opening tag

src/lib/components/LayeredMap.js
```diff
@@ -47,20 +49,28 @@ class LayeredMap extends Component {
...
     }
 }
 
 LayeredMap.defaultProps = {
     height: 800,
-    hillShading: true
+    hillShading: true,
+    show_marker: false,
+    show_polygon: false,
+    show_polyline: false
```
anders-kiaer: Should the props be prefixed with e.g. draw_toolbar instead of show, just to avoid any chance of confusion with layer data input?

HansKallekleiv:
thx. should be fixed now :+1:

PR #48, opened 2019-09-12
Remove regl dependency

### LINKED ISSUE #28: Use WebGL directly instead of regl
Currently `regl` is used as a wrapper around WebGL in the map component. This is a bit too high level in order optimize e.g. by compiling shaders once (regardless of data input). The high level functional approach of `regl` makes this difficult/impossible.

anders-kiaer:
Resolves #28.

- [X] Use WebGL (texture) instead of `regl` for adding image data to canvas directly (i.e. greyscale image without colormap).
- [X] Port the shader code for `greyscale → colored image` conversion.
- [X] Port the shader code for direct light effect (aka. hill shading).


src/lib/private_components/layered-map-resources/alter_image.js
```diff
@@ -1,39 +1,166 @@
...
+         * @param {image element} image - The image element to use as input data for the texture
+         */
+
+        const program = gl.getParameter(gl.CURRENT_PROGRAM)
+
+        gl.activeTexture(gl.TEXTURE0 + textureIndex)
+        gl.bindTexture(gl.TEXTURE_2D, gl.createTexture())
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE)
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE)
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST)
```
HansKallekleiv: Should this be a prop?

src/lib/private_components/layered-map-resources/alter_image.js
```diff
@@ -1,39 +1,166 @@
...
+                const vectorLength = Math.sqrt(light_direction[0]**2 + light_direction[1]**2 + light_direction[2]**2)
+        
+                gl.uniform3f(
+                    gl.getUniformLocation(program, 'u_light_direction'),
+                    light_direction[0] / vectorLength,
+                    light_direction[1] / vectorLength,
+                    light_direction[2] / vectorLength
+                )
+
+                const elevationScale = 0.03
```
HansKallekleiv: Should this be a prop?

src/lib/private_components/layered-map-resources/alter_image.js
anders-kiaer: Yes, good idea. We could have a property (maybe on the `image` layer element) which states if it should be linear og nearest data point interpolation. When thinking about it, might be that `gl.LINEAR` vs. `gl.NEAREST` will not make any difference (since it is one data point per pixel). However, the CSS property [`image-rendering`](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering) will (where e.g. `pixelated` could be a suitable choice in some scenarios). I'll make an issue of it 📃

src/lib/private_components/layered-map-resources/alter_image.js
anders-kiaer: Yes 💡  There are some properties related to hillshading which should probably be adjustable:
 - Light direction
 - Elevation scale
 - The normalization in the "dot product".
I'll make a separate issue out of it 👍

PR #52, opened 2019-09-16
Update colormap on prop change

### LINKED ISSUE #51: LayeredMap: Colormap does not update on new props

HansKallekleiv:
Closes #51


anders-kiaer:
Maybe we could use this opportunity to use `ReactDOM.createPortal` here as well (similar to how it is used in `Switch`), in order to get «back into the React machinery». Would probably fix the bug + simplify existing code at the same time.

HansKallekleiv:
Sounds good!

anders-kiaer:
Moved to #53.

PR #76, opened 2019-11-06
LayeredMap: Set default bounds if no layers are defined

HansKallekleiv:


src/lib/components/LayeredMap.js
```diff
@@ -32,7 +32,9 @@ class LayeredMap extends Component {
     calculateBounds() {
         const x_arr = [];
         const y_arr = [];
-
+        if (this.props.layers.length == 0) {
```
anders-kiaer: `===` instead of `==`.

src/lib/components/LayeredMap.js
```diff
@@ -116,7 +117,10 @@ class LayeredMap extends Component {
     }
 
     componentDidUpdate(prevProps) {
-        if (this.props.uirevision !== prevProps.uirevision) {
+        if (
+            this.props.uirevision === "" ||
```
anders-kiaer: If this is the intended/wanted behaviour (and not instead checking that we go from 0 to >0 layers), maybe update the [`uirevision` documentation](https://github.com/equinor/webviz-subsurface-components/blob/9dfa4e7a72f617087b286e1ec31f82e98bcb7879/src/lib/components/LayeredMap.js#L354) saying that if it is not set at all, the view will always reset.

Issue #25, opened 2019-09-02
Show colormaps only for selected layers


anders-kiaer:
Currently, all colormaps for input layers are shown. This could be changed to a probably more user friendly behaviour where only colormap for selected layers are shown.

You can listen to layer changes by e.g. doing

```
componentDidMount() {
    this.mapRef.current.leafletElement.addEventListener('overlayremove', (e) => {console.log(e)});
    this.mapRef.current.leafletElement.addEventListener('overlayadd', (e) => {console.log(e)});
    this.mapRef.current.leafletElement.addEventListener('baselayerchange', (e) => {console.log(e)});
    }
```
in the `LayeredMap` component.

When doing this, we can show the value (i.e. `z` value) of the "selected" point in each layer (and corresponding colormap).

anders-kiaer:
Moved to #110 for better overview. Closing this.

PR #81, opened 2019-11-11
Show x, y, z on canvas mouseover

### LINKED ISSUE #33: Show coordinates on LayeredMap
A user needs to see the coordinates the LayeredMap is set to.
Investigate what option is the best:
- Coordinate axis  ([example](http://www.physicalgeography.net/fundamentals/2b.html))
- Coordinate on mouseover
- Other...

### LINKED ISSUE #82: Show value of map on mouseover

anders-kiaer:
Resolves #82. Resolves #33.

- [X] Extract `z` value from e.g. canvas on click.
- [X] Add info box with `x`, `y` and "`z`" coordinates.
- [x] Style info box
- [x] Make z unit configurable.
- [x] Scale from `z in [0, 255]` to physical units.


anders-kiaer:
Yes, that should be possible to add.

> highlighting of the selected cell?

Could adding a [circlemarker](https://react-leaflet.js.org/docs/en/components#circlemarker) be a simple solution, where we then `onclick` just update the `x, y` state of that marker?

HansKallekleiv:
> Could adding a [circlemarker](https://react-leaflet.js.org/docs/en/components#circlemarker) be a simple solution, where we then `onclick` just update the `x, y` state of that marker?

Sounds good 👍

anders-kiaer:
Latest clicked location is now marked with a circle. Other synced `LayeredMap` components get a circle corresponding to the same location, in addition to an infobox stating the same `x` and `y`. 

Getting other components to calculate their corresponding "`z` value" for the same location appears to be somewhat trickier, as we need to know what is the topmost layer at the time of the click in the ther components, and also if that layer is a canvas/image layer.

anders-kiaer:
Showing `z` value in "synced" components is moved to #25.

PR #143, opened 2020-06-25
Sync zoom

eolweus:


src/lib/components/NewLayeredMap/LayeredMap.js
```diff
@@ -31,10 +31,14 @@ const stringToCRS = (crsString) => {
...
 class LayeredMap extends Component {
 
+    static mapReferences = {};
+
     constructor(props) {
         super(props);
         
         this.state = {
+            id: props.id,
+            syncedMaps: props.syncedMaps,
```
Andorr, line 41: This does not necessarily need to be in _this.state_? 🤔

PR #190, opened 2020-07-21
refactor CML and add control folder in components

eolweus:


src/lib/components/NewLayeredMap/NewLayeredMap.jsx
```diff
@@ -7,7 +7,7 @@ import './layers/L.imageWebGLOverlay';
 import './layers/L.tileWebGLLayer';
 
 // Components
-import Controls from './components/Controls';
+import Controls from './components/controls/Controls';
```
Andorr, line 10: Call the folder for **Controls** and then have a index.js inside of that folder which exports the Controls.js.
```
export { default } from './Controls';
```
With this, the end result will be the following import:
```
import Controls from './components/Controls';
```
...which is more convenient than `./components/controls/Controls` 😛

src/lib/components/NewLayeredMap/components/leafletUtils.js
```diff
@@ -0,0 +1,118 @@
+export const getShapeType = layer => {
```
Andorr, line 1: Hmmm, does this file really belong in the _components_-directory though? 🤔 I really don't feel like it is a component, and finding it in the root of the components-directory would be unexpecting, will it not?

I suggest that you rather make a _utils_ directory inside _NewLayeredMap_-directory, and put it there instead 😛 Thoughts? 😅

PR #196, opened 2020-07-22
Dash app

dsp0011:


src/lib/components/NewLayeredMap/NewLayeredMap.jsx
```diff
@@ -372,6 +372,11 @@ NewLayeredMap.propTypes = {
      * Map coordinates of a mouse click
      */
     click_position: PropTypes.array,
+
+    /**
+     * Allows to choose between replacing the layers or updating them
+     */
+    layerChangeMethod : PropTypes.string,
```
Andorr, line 379: Can we maybe call this one _updateStrategy_ or something? 🤔

src/lib/components/NewLayeredMap/components/Controls/components/MousePosition.js
```diff
@@ -87,7 +87,7 @@ const MousePosition = (props) => {
         const onScreenCanvas = focusedImageLayer.getCanvas ? focusedImageLayer.getCanvas() : null;
         const minvalue = (focusedImageLayer.options || {}).minvalue;
         const maxvalue = (focusedImageLayer.options || {}).maxvalue;
-        if(!url || !onScreenCanvas || !minvalue || !maxvalue) {
+        if(!url || !onScreenCanvas || minvalue == null || !maxvalue) { // TODO fix minvalue = 0
```
Andorr, line 90: Should be `===` and also need to check for undefined. Maybe a type-check if enough:

```
typeof minvalue === "number"
```

src/lib/components/NewLayeredMap/layers/L.imageWebGLOverlay.js
```diff
@@ -94,6 +94,17 @@ L.ImageWebGLOverlay = L.Layer.extend({
 		return this;
     },
     
+    reUseColorscale: function (newOptions) {
+        const newColorScaleOptions = newOptions.colorScale;
+        const updatedColorScale = this.options.colorScale;
+
+        if(this.options.colorScale) {
```
Andorr, line 101: Can easily be replaced with `Object.assign({}, ...)`.

In addition, the colorScale-property is not necessarily an object. It might be an object and might be null.

src/lib/components/NewLayeredMap/layers/L.imageWebGLOverlay.js
```diff
@@ -111,9 +122,14 @@ L.ImageWebGLOverlay = L.Layer.extend({
     updateOptions: function(options) {
         const promisesToWaitFor = [];
 
+        if(this.options.colorScale) {
+            const newColorScale = this.reUseColorscale(options)
+            options.colorScale = newColorScale
```
Andorr, line 127: This removes the possibility to remove the colorScale.  You should instead merge with the other colorScale-checks in this function:

```
if(!options.colorScale) {
            this._colormapUrl = null;
} else {
            if(options.colorScale instanceof Object && this.options.colorScale instanceof Object) {
                  options.colorScale = Object.assign(this.options.colorScale, options.ColorScale)
            }
            this._initColormap();
}
```
Something like that. If may have to move the if-statements around in the function, but something like above is enough.

src/lib/components/NewLayeredMap/NewLayeredMap.jsx
eolweus: I'm working on this, and I use the name updateMode, but I thought I'd just change it when I merge this into my branch

src/lib/components/NewLayeredMap/layers/L.imageWebGLOverlay.js
```diff
@@ -139,6 +122,9 @@ L.ImageWebGLOverlay = L.Layer.extend({
         if(!options.colorScale) {
             this._colormapUrl = null;
         } else {
+            if(options.colorScale instanceof Object && this.options.colorScale instanceof Object) {
```
Andorr, line 125: I probably was not clear enough (😅), but I think I mentioned that you have to move this if-statement, as options.colorScale and this.options.colorScale has already been merged here, and is the same. 

Move the if-statement above the `Util.setOptions(...)` and add the following:
```
if(options.colorScale && options.colorScale instanceof Object && this.options.colorScale instanceof Object) {
 options.colorScale = Object.assign(this.options.colorScale, options.colorScale)
}
```

examples/example_new_layered_map_v3.py
```diff
@@ -0,0 +1,764 @@
...
+							html.H6("Minimum, maximum map values", id='min-max-output-container'),
+							dcc.RangeSlider(
+								id='min-max-slider',
+								min=min_value - 1000,
+								max=max_value + 1000,
+								step=0.5,
+								value=[min_value, max_value],
+                                marks={
+                                    str(min_value - 1000): str(min_value - 1000),
+                                    str(max_value + 1000): str(max_value + 1000)
```
eolweus, line 260: what's going on with the indenting in this segment?

src/lib/components/NewLayeredMap/layers/L.imageWebGLOverlay.js
Andorr: Can you also add this logic to the shader as well? 😮

PR #197, opened 2020-07-23
Update mode

eolweus:


.vscode/settings.json
```diff
@@ -1,4 +1,4 @@
 {
-    "python.pythonPath": "venv\\Scripts\\python.exe",
+    "python.pythonPath": "c:\\Users\\erol1\\git\\equinor\\original_component\\env\\Scripts\\python.exe",
```
Andorr, line 2: Amazing

eolweus: hahaha, ups. wonder how that was pushed. Should be in gitignore 😅

PR #205, opened 2020-07-27
🗻 Fix colorbar bug where it was not displaying correct ColorScale-data

Andorr:


setup.py
```diff
@@ -41,7 +41,8 @@
     install_requires=INSTALL_REQUIRES,
     tests_require=TESTS_REQUIRE,
     extras_require={"tests": TESTS_REQUIRE, "dependencies": INSTALL_REQUIRES},
-    setup_requires=["setuptools_scm>=3.2"],
+    #setup_requires=["setuptools_scm>=3.2"],
+    version='0.0.29',
```
eolweus, line 45:
```suggestion
    setup_requires=["setuptools_scm>=3.2"],
```

PR #206, opened 2020-07-27
🚲 Tilelayer hillshading issue - added the possibility to draw multiple tiles at the same time

Andorr:


src/lib/components/NewLayeredMap/webgl/commands/drawWithAdvancedHillShading.js
```diff
@@ -300,7 +299,9 @@ function calcN(width, height) {
     }
     else if(resolution <= 810000) { // 900x900
         return 18;
+    } else if(resolution < 2293760) { // 1792x1280
+        return 8;
     }
 
-    return 8; // resoltion > 900x900
+    return 1; // resoltion > 900x900
```
eolweus, line 306: Doesn't this make it so that the default return is when resolution > 1792x1280?

Andorr: Yup, correct x)

PR #207, opened 2020-07-27
Switch bugfix

dsp0011:


src/lib/components/NewLayeredMap/components/Controls/components/MousePosition.js
```diff
@@ -156,13 +163,39 @@ const MousePosition = (props) => {
         const red = ctx.getImageData(screenX, screenY, 1, 1).data[0]; // TODO: store this locally 
         const z = mapZValue(red);
 
+        // Used to disable map.on("click") events when pressed on the controls
+        const controlsList = document.getElementsByClassName("leaflet-custom-control leaflet-control")
```
Andorr, line 167: Hmmm... this code will run this one on every mouse click and generate the same result of _forbiddenCoordinates_ 🤔 If we would want to run this code as few times as possible, when for example the controls changes, it would maybe complicate things too much. What do you guys think? It is not heavy either, so it should be good to just keep it this way.

eolweus: hmm, I think it's fine. Not too heavy as you said, and I think I don't think you save that many operations if you have to run it every time the map moves/zooms/pans anyways

PR #212, opened 2020-07-28
Make shapes emit JSON object when clicked

eolweus:


src/lib/components/NewLayeredMap/NewLayeredMap.jsx
```diff
@@ -378,6 +379,14 @@ NewLayeredMap.propTypes = {
      */
     updateMode: PropTypes.string,
 
+    /**
+     * syncMapSize makes the map listen for changes in width and height and automatically recalculates
+     * the map dimensions when changes occur.
+     * 
+     * // TODO: maybe give this a different name?
+     */
+    syncMapSize: PropTypes.bool,
```
Andorr, line 388: Then we should do it asap, tbh. Any suggestions? 🤔

PR #216, opened 2020-07-29
🌋Suggestion for restructuring on component file-structure

Andorr:
## Introduction
The current file-structure for the components are not following the best practices for optimal file structure for React applications. Currently, there are two directories - _components_ and _private_components_, where each component in _components_ have their in a specific folder in the _private_components_-directory. This means that the logic for a single components is "spread out in different directories" around the project, making it not obvious where to find the belonging files for a given component.

I think the reason for the current structure is due to dash, as _dash-generate-components_ searches for all JS-files inside the _components_-directory. You only want dash to extract the React-components inside of the _components_-directory, not other utiliy files. This is probably why the "utility files" was put inside the _private_components_ directory.

## Suggestion
Put all component-specific files inside the _components_-directory, inside a specified directory for each component:
```
src
 |----components
           |----LayeredMap
                     |-----components
                     |-----shaders
                     |-----webgl
                     |-----layered-map.css
                     |-----LayeredMap.jsx
                  
```
This is a cleaner file-structure as all the logic needed for the _LayeredMap_-component can be found inside the _LayeredMap_-directory and not all over the place. 

However, this makes _dash-generate-components_ try to generate a Dash-component for each Javascript file inside of _components_, which is not wanted. Therefore, the solution is to change the file-extension for all wanted dash-components from `.js` to `.jsx`, and make _dash-generate-components_ only look for `.jsx`-files. Using the `.jsx`-extension is also quite normal for React-files.

So to activate these changes we had to do the following:
image
image


anders-kiaer:
LGTM 👍 Should we add a sentence in `README.md` regarding which file extensions (`.jsx`) `dash` will generate Python components from?

Andorr:
> LGTM 👍 Should we add a sentence in `README.md` regarding which file extensions (`.jsx`) `dash` will generate Python components from?

Think that is a good idea, yeah 👍 Will do ^^

PR #217, opened 2020-07-29
Scale bugfix

dsp0011:


src/lib/components/NewLayeredMap/NewLayeredMap.jsx
```diff
@@ -338,6 +339,11 @@ NewLayeredMap.propTypes = {
         position: PropTypes.string,
     }),
 
+    /**
+     * UnitScale is a box that displays the the current unit scale on the map.
+     */
+    unitScale: PropTypes.bool,
+
```
eolweus, line 346: sorry, I changed this. Should be like this:
```suggestion
    /**
     * UnitScale is a box that displays the the current unit scale on the map.
     */
    unitScale: PropTypes.shape({
        position: PropTypes.string,
    }),

```

src/demo/NewLayeredMapDemo.js
```diff
@@ -121,6 +122,7 @@ const NewLayeredMapDemo = () => {
                     syncedMaps={["NewLayeredMap-4"]}
                    // syncDrawings={true}
                     layers={layers}
+                    unitScale = "bottomright"
```
eolweus, line 125:
```suggestion
                    unitScale={{ position: "bottomright" }}
```

PR #221, opened 2020-07-30
Eslint fix

dsp0011:


src/lib/components/LeafletMap/layers/L.tileWebGLLayer.js
```diff
@@ -168,12 +168,7 @@ L.TileWebGLLayer = L.GridLayer.extend({
         return Util.template(this._url, Util.extend(urlData, this.options));
     },
 
-    updateOptions: function (options) {
-        options = Util.setOptions(this, {
```
Andorr, line 172: You still need this one! `Util.setOptions(this, ...)`. Take it back.

Andorr: You can remove `òptions = ` but `Util.setOptions(this, ...)` needs to stay.

Issue #230, opened 2020-08-26
Option to recalculate map bounds when using updateMode="update"


HansKallekleiv:
Either automatically based on changed in bounds of baselayer or as a prop, e.g. `uirevision`.

PR #231, opened 2020-08-27
Calculate bounds when first receiving layers

HansKallekleiv:
Relates to #230


anders-kiaer (changes requested):
As discussed, only add a sentence in `README.md` regarding the new behaviour, otherwise LGTM! 🚀

anders-kiaer (approved):

PR #234, opened 2020-09-03
Shifted legend colors for simulated values too high and too low

### LINKED ISSUE #233: HistoryMatch Legend colors wrong
Simulated values too high and simulated values too low should be shifted.

Bar colors are defined [here](https://github.com/equinor/webviz-subsurface-components/blob/9147efa4dcb0428c29d6d538c7fec361dd4a6778/src/lib/components/HistoryMatch/components/history_matching_plot.js#L27-L33) with "Negative"  getting `#34A037` and positive getting `#40537D`. Instead of using these properties in the legend, they are hard coded [here](https://github.com/equinor/webviz-subsurface-components/blob/9147efa4dcb0428c29d6d538c7fec361dd4a6778/src/lib/components/HistoryMatch/components/history_matching_plot.js#L161-L177), with opposite definition (also the stroke is not the same for one of them).

asnyv:
Closes #233

### DIFF
--- a/src/lib/components/HistoryMatch/components/history_matching_plot.js
+++ b/src/lib/components/HistoryMatch/components/history_matching_plot.js
@@ -162,17 +162,17 @@ export default class HistoryMatchingPlot extends Component {
             {
                 label: "Simulated values too high",
                 box: {
-                    colour: "#34A037",
-                    stroke: "#050",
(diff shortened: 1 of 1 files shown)


anders-kiaer (approved):
:tada: LGTM!

PR #243, opened 2020-09-30
Dynamic Tree

### LINKED ISSUE #47: Add Eclipse GRUPTREE visualization

anders-kiaer:
Closes #47.

### DIFF
--- a/src/demo/App.js
+++ b/src/demo/App.js
@@ -2,17 +2,18 @@
 import React, { Component } from "react";
 
 // Components
+import DynamicTreeDemo from "./DynamicTreeDemo";
 import HistoryMatchDemo from "./HistoryMatchDemo";
-import MorrisDemo from "./MorrisDemo";
-import SubsurfaceMapDemo from "./SubsurfaceMapDemo";
 import LayeredMapDemo from "./LayeredMapDemo";
-import PriorPosteriorDistributionDemo from "./PriorPosteriorDistributionDemo";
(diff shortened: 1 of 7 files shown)


asnyv:
Haven't looked at it yet, but do you also support the extended network option, or just regular GRUPTREE/GRUPNET combo?

anders-kiaer:
> Haven't looked at it yet, but do you also support the extended network option, or just regular GRUPTREE/GRUPNET combo?

The main assumption by the component itself is only that the network has a "tree structure" (i.e. any node has only one branch connected to it in the direction of the terminal node at the top of the tree). The tree structure can change through time.

asnyv:
Was thinking more if it made assumptions on keywords and etc, but would think that is handled by how we wrap it and give json-input?

E.g. in Eclipse you can either use the standard network option with `GRUPTREE` and `GRUPNET`, or the extended option with `BRANPROP` and `NODEPROP` which is freed from your grupnet, allowing you to have independent production group reports and actual flow networks. Still a tree structure, just a bit different input.

anders-kiaer:
Yes, I think the dependendices of keywords will first enter on the `webviz-subsurface` plugin. The component itself will probably be agnostic and independent of keyword (and instead have parameter names indicating what it is used for in the visualization).

anders-kiaer:
Closing the prototype PR. To be replaced with a new component.

PR #244, opened 2020-10-06
Set default props in LeafletMap

HansKallekleiv:
- Added defaultProp for `layers` to allow initialization of the component without layer data
- Added defaultProp for `syncedMaps` to simplify the code


anders-kiaer (approved):
LGTM! 🚀👏

Issue #242, opened 2020-09-30
Hillshading and shadows "follows the colormap"


anders-kiaer:
Hillshading/shadows currently "follow the colormap", in the sense that instead of adding black/grey (vith varying degree of transparency), it instead tunes colors with the existing colormap.

Makes more sense to add amount of black wrt. shadows, instead of going down on data colormap?

thor85:
I would vote for black shadows as default at least. Should not be dependent on colormap. Add a seperate parameter for shadow color if necessary (probably not?).

PR #254, opened 2020-12-03
Enable hillshading by default in the Leaflet Map demo

alexandruandrei91:
I've enabled the hillshading by default in the Leaflet Map demo to match the switch button state.

 Although it looks nice, it's pretty slow to load because of the shadow calculation. I can turn hillshading off by default, so we don't stare at a blank screen when first loading the demo, or I can turn the shadows off, which makes it load a lot faster, but looses the eye-candy. Any suggestion?

[Toy pull request to test the setup]


HansKallekleiv (commented):
Nice start @alexandruandrei91 :)

Discussed a bit with @anders-kiaer and a suggestion is to reduce the number of iterations as described here:
https://github.com/equinor/webviz-subsurface-components/blob/master/src/lib/components/LeafletMap/README.md#-hillshading

Another alternative is perhaps to show hillshading as default in only one of the maps?

anders-kiaer:
Agree with @HansKallekleiv - nice start @alexandruandrei91 🎉 

**Background info:**

The demo I think is using the default number of iterations. The summer interns did a great job earlier this year when creating `LeafletMap`. Since full shadow is costly for high number of iterations / ray tracing on large pictures, the default number of iterations is currently set here as a function of `width * height` of the tile/canvas to be calculated for:
https://github.com/equinor/webviz-subsurface-components/blob/946e12f3d67c0d390bd2b0afbd297a2dd3fb630b/src/lib/components/LeafletMap/webgl/commands/drawWithAdvancedHillShading.js#L285-L315

As stated in the function comment, the default values could probably benefit from being tuned a bit more when it comes to default shadow details vs. rendering time 🙂

I would expect the hill shading rendering time to be even more consistent across different data sets when using tile server, since the tiles being hill shaded is then at a reasonable + ~more constant resolution in terms of pixels, regardless of zoom level.

alexandruandrei91:
I've changed the shadow iteration from using the default (in this case it was 8) to an explicit 2 iterations. It seems to load alright.

In the end I chose to do it in the demo only, and not change calcN because it might be a bit early to do tweaking, considering that the coloring and shadows are not correct (#242) and might change.

anders-kiaer (approved):
LGTM 🚀

Issue #238, opened 2020-09-16
Set min/max values for color scale in LeafletMap layers


HansKallekleiv:
It is currently not possible to specifiy min/max values for the color scale in Leaflet Map layers.
It is possible to specify  `cutpointmin` and `cutpointmax` but these only set values outside the limits transparent.

An option to set the min / max points for the colorscale should be added.

PR #260, opened 2020-12-14
Add hillshading shader in one pass

### LINKED ISSUE #242: Hillshading and shadows "follows the colormap"

alexandruandrei91:
I've merged the behavior of the previous hill shading approach into one shader, to avoid iterating over the framebuffers multiple times. This had the side effect of making it easier to add new parameters and rearrange the order of the applied operations, so in this review I've also fixed #242 and #238.
 
  One thing I didn't add was the soft shadows, as it added a big overhead without having a clear benefit for the visualization imo. The other topological maps that I've seen don't usually cast shadows. They only have the shadows caused by the direct illumination.

 I've also changed the demo so that it's using this new shader.


src/demo/LeafletMapDemo.js
```diff
@@ -6,29 +6,11 @@ import LeafletMap from "../lib/components/LeafletMap";
...
-        "#f0f921",
-    ],
-    prefixZeroAlpha: false,
-    scaleType: "linear",
-    cutPointMin: 3400,
-    cutPointMax: 3500,
-};
-
 const LeafletMapDemo = () => {
     const [layers, setLayers] = useState(exampleData.layers.slice(1, 5));
```
ahboujelben, line 10: The code is useState is executed each render (even though it's assigned once to the state vars). Using a callback instead is a minor optimisation to prevent the execution
```suggestion
    const [layers, setLayers] = useState(() => exampleData.layers.slice(1, 5));
```

HansKallekleiv (approved):

PR #263, opened 2020-12-18
Add mapbox component

### LINKED ISSUE #262: Add basic MapBox component
Add a simple [mapboxgl](https://docs.mapbox.com/mapbox-gl-js/api/) component for evaluation purposes.

alexandruandrei91:
I've added a simple MapBox component that currently just renders data coming from the mapbox tile servers.
The plan is to continue evaluating it as an alternative for Leaflet.

In this review I've also added a new npm script, `validate`, which triggers linting and type checking of typescript files. I've added it to the github workflow as well.

Closes #262.


package.json
```diff
@@ -18,18 +18,22 @@
...
     "license": "LGPL",
     "dependencies": {
         "@material-ui/core": "^4.3.3",
+        "@types/mapbox-gl": "^2.0.0",
         "@types/react": "^16.9.35",
         "@types/react-dom": "^16.9.8",
         "d3": "^5.11.0",
         "leaflet": "^1.6.0",
         "leaflet-draw": "^1.0.4",
+        "mapbox-gl": "^2.0.0",
```
nicktalbot, line 36: MapBox JS 2.x is no longer BSD licensed. Every map load is billable.
Perhaps use https://github.com/maplibre/maplibre-gl-js forked from MapBox JS 1.x?

alexandruandrei91: Nice catch, Nick. I assumed it's a permissive license, so I didn't check.. Probably Mapbox is out of question then. I would be reluctant to go with a fork, considering it doesn't seem very well established.

Maybe we should do similar investigations of other alternatives. Some that I found after a quick search would be [OpenLayers] (https://openlayers.org/),  [Tangram from Mapzen](https://www.mapzen.com/products/tangram/), [HarpGL from HereMaps](https://www.harp.gl) and [DeckGL from Uber](https://deck.gl).
We can also continue building on top of Leaflet, of course.

@asnyv, @HansKallekleiv, what do you think, should I just drop this pull request, or move it to the libre fork? Did you already looked into any of the listed alternatives?

alexandruandrei91:
Close due to Mapbox license change.

Issue #264, opened 2021-01-06
Investigate alternative map technologies


alexandruandrei91:
Some that I found after a quick search would be [OpenLayers](https://openlayers.org/),  [Tangram from Mapzen](https://www.mapzen.com/products/tangram/), [HarpGL from HereMaps](https://www.harp.gl) and [DeckGL from Uber](https://deck.gl).
We can also continue building on top of Leaflet, of course.

alexandruandrei91:
https://github.com/equinor/webviz-subsurface-components/wiki/Map-Component

Issue #251, opened 2020-12-02
Map component


HansKallekleiv:
Scope out requirements of a map component

PR #275, opened 2021-02-02
Add DeckGL map component

alexandruandrei91:
Adding a DeckGL map component with support for color maps.

#264 #251


src/lib/components/DeckGLMap/DeckGLMap.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+    return (
+        <div style={{ height: "100%", width: "100%", position: "relative" }}>
+            {jsonProps && <DeckGL id="serialized-deck" {...jsonProps} />}
+        </div>
+    );
+}
+
+DeckGLMap.propTypes = {
+    jsonData: PropTypes.object,
```
HansKallekleiv, line 26: `id` should be a prop. Required for the Dash component

src/demo/DeckGLMapDemo.js
```diff
@@ -0,0 +1,144 @@
...
+
+// Component used to catch DeckGL errors an display a message.
+class ErrorBoundary extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = { hasError: false };
+    }
+
+    componentDidCatch(error, errorInfo) {
+        console.log(error, errorInfo);
```
HansKallekleiv, line 23: use console.warn or console.error instead of console.log?

HansKallekleiv (changes requested):
Small change and a question, otherwise LGTM 👍

src/lib/components/DeckGLMap/configuration.js
```diff
@@ -0,0 +1,50 @@
...
+    OrbitView,
+    OrthographicView,
+} from "@deck.gl/core";
+import * as Layers from "@deck.gl/layers";
+import * as AggregationLayers from "@deck.gl/aggregation-layers";
+import * as GeoLayers from "@deck.gl/geo-layers";
+import * as MeshLayers from "@deck.gl/mesh-layers";
+import * as CustomLayers from "./layers";
+
+import { COORDINATE_SYSTEM } from "@deck.gl/core";
```
aprilzhang, line 15: Just a very minor thing, the vs code has a short cut to organize import, (it is Shift+Alt+O on Windows, not sure about Mac...)

aprilzhang (commented):
Looks good to me!

src/lib/components/DeckGLMap/configuration.js
alexandruandrei91: I've just tried that, but it organizes everything alphabetically, combining imports from different libraries.
Now I've manually arranged them in alphabetical order, while keeping the imports grouped: React stuff, then DeckGL stuff, then other libraries, then local stuff. For me it's easier to visually "parse" them, but for consistency I can also use that shortcut. Let me know what you think :)

src/lib/components/DeckGLMap/DeckGLMap.jsx
```diff
@@ -0,0 +1,29 @@
...
+import DeckGL from "@deck.gl/react";
+import { JSONConverter, JSONConfiguration } from "@deck.gl/json";
+
+import PropTypes from "prop-types";
+
+import JSON_CONVERTER_CONFIGURATION from "./configuration";
+
+function DeckGLMap(props) {
+    const configuration = new JSONConfiguration(JSON_CONVERTER_CONFIGURATION);
+    const jsonConverter = new JSONConverter({ configuration });
```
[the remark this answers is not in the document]
alexandruandrei91: I've just tried that but for some reason it doesn't work when it the jsonData prop changes, so I've moved them in the `useEffect`. Not sure if that's the best place though.

src/lib/components/DeckGLMap/configuration.js
aprilzhang: Combining imports from different libraries sounds like an odd behavoir, I was expecting it would organise the imports by merging the ones from the same library. But yeah I think this doesnt matter. :)

HansKallekleiv (approved):

PR #280, opened 2021-02-03
Basic version of Well Completions component

aprilzhang:


examples/example_well_completions.py
```diff
@@ -0,0 +1,25 @@
...
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#
+# Copyright (C) 2020 - Equinor ASA.
+
+import json
+import webviz_subsurface_components
+import dash
+import dash_html_components as html
+
+with open('../src/demo/example-data/well-completions.json', "r") as json_file:
```
alexandruandrei91, line 12: I think ideally in these python examples, we process the real data into the format the component expects. In this case, converting from a given csv to the json you need for the react component (I guess using well-completions-csv2json.py)

alexandruandrei91 (commented):
Looks good to me 👍

HansKallekleiv (approved):
LGTM 👍

Issue #276, opened 2021-02-03
Add hillshading in DeckGL map


alexandruandrei91:
Could be done in the pixel shader, like for Leaflet, or create a 3D mesh, like DeckGL's terrain layer.

#251

PR #294, opened 2021-02-16
Add hillshading support to the DeckGL map

alexandruandrei91:
This PR adds a hill shading layer that can be overlapped on top of the existing layers.

#276


HansKallekleiv (approved):
Lgtm 🥳

PR #304, opened 2021-03-01
Switch the DeckGL map component to typescript

alexandruandrei91:


package.json
```diff
@@ -21,11 +21,13 @@
...
         "format": "eslint --fix *.js *json 'src/**/*.+(ts|tsx|js|jsx|json|css)'",
         "linting": "eslint *.js *.json 'src/**/*.+(ts|tsx|js|jsx|json|css)'",
-        "validate": "npm run typecheck && npm run linting"
+        "validate": "npm run typecheck && npm run linting",
+        "setup_deckgl_types": "cd ./node_modules/@danmarshall/deckgl-typings/ && indefinitely-typed --folder deck.gl__aggregation-layers --folder deck.gl__core --folder deck.gl__extensions --folder deck.gl__geo-layers --folder deck.gl__google-maps --folder deck.gl__json --folder deck.gl__layers --folder deck.gl__mapbox --folder deck.gl__mesh-layers --folder deck.gl__react --folder deck.gl --folder luma.gl__constants --folder luma.gl__core --folder luma.gl__gltools --folder luma.gl__webgl-state-tracker --folder luma.gl__webgl --folder math.gl__core --folder math.gl"
     },
     "author": "R&T Equinor <akia@equinor.com>",
     "license": "LGPL",
     "dependencies": {
+        "@danmarshall/deckgl-typings": "^4.5.6",
```
HansKallekleiv, line 30: Is it necessary to have these as dependency or do they belong in dev-dependencies?

package.json
```diff
@@ -66,6 +68,7 @@
         "eslint-plugin-prettier": "^3.1.3",
         "eslint-plugin-react": "^7.19.0",
         "file-loader": "^6.2.0",
+        "indefinitely-typed": "^1.1.0",
```
HansKallekleiv, line 71: Same as above

HansKallekleiv (approved):
LGTM.
Just a small question regarding dependencies

package.json
alexandruandrei91: I think the deckgl types need to be in dependencies, as they are used in parts exposed to the user, eg our ColormapLayerProps extend deckgl's BitmapLayerProps.

https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies

package.json
alexandruandrei91: This is actually in devDependencies, but it's probably not obvious when code-reviewing because of code collapse.

PR #309, opened 2021-03-05
Added layer for displaying wells as GeoJson.

hkfb:


src/lib/components/DeckGLMap/DeckGLMap.jsx
```diff
@@ -19,7 +19,15 @@ function DeckGLMap(props) {
 
     return (
         <div style={{ height: "100%", width: "100%", position: "relative" }}>
-            {jsonProps && <DeckGL id={props.id} {...jsonProps} />}
+            {jsonProps && (
+                <DeckGL
+                    id={props.id}
+                    {...jsonProps}
+                    getTooltip={({ object }) =>
```
alexandruandrei91, line 26: We'll have to change this soonish, but for now I think it's ok.

src/demo/example-data/deckgl-map.json
```diff
@@ -62,6 +62,11 @@
                     "valueRange": 731.0,
                     "opacity": 1.0,
                     "image": "https://i.imgur.com/Dx8eseE.png"
+                },
+                {
+                    "@@type": "WellsLayer",
+                    "id": "wells-layer",
+                    "data": "src/demo/example-data/emerald_wells.json"
```
alexandruandrei91, line 69: These wells are not overlapping with the surface, so maybe you can get some example wells from the same project as the surface above (the surface will be properly placed after #308 is merged). I think @Midtveit posted the RMS project somewhere.

hkfb: @alexandruandrei91 replaced Emerald wells with wells from Volve project in the latest commit.

alexandruandrei91 (approved):
lgtm 👍

HansKallekleiv (commented):
I can't see any wells in the demo. Are they supposed to be there now ? :)

alexandruandrei91:
@HansKallekleiv, they weren't overlapping with the map because the map wasn't positioned at the proper place. Now that #308 got merged, you should see the wells.

HansKallekleiv (approved):
Lgtm

hkfb:
Will create a new pull request with latest changes.

PR #313, opened 2021-03-11
Added layer for displaying wells as GeoJson

hkfb:


src/lib/components/DeckGLMap/layers/wells/wellsLayer.js
```diff
@@ -0,0 +1,31 @@
+import { CompositeLayer } from "@deck.gl/core";
+import { GeoJsonLayer } from "@deck.gl/layers";
+
+export default class WellsLayer extends CompositeLayer {
+    renderLayers() {
+        var properties = {
```
alexandruandrei91, line 6: `var` has some weird scoping rules, so better to stick to `let`

src/lib/components/DeckGLMap/layers/wells/wellsLayer.js
```diff
@@ -0,0 +1,31 @@
...
+            id: "wells-shadow",
```
alexandruandrei91, line 7: You can use `getSubLayerProps` (See [the doc](https://deck.gl/docs/developer-guide/custom-layers/composite-layers#forwarding-properties)) to propagate common attributes such as `opacity`, `pickable` etc and concatenate the `id` with the parent `id`.

src/lib/components/DeckGLMap/layers/wells/wellsLayer.js
```diff
@@ -0,0 +1,31 @@
...
+            data: this.props.data,
+            opacity: this.props.opacity,
+            pickable: true,
+            pointRadiusUnits: "pixels",
+            lineWidthUnits: "pixels",
+            pointRadiusScale: this.props.pointRadiusScale,
+            lineWidthScale: this.props.lineWidthScale,
+        };
+
+        const shadow = new GeoJsonLayer(properties);
```
alexandruandrei91, line 17: Not sure how expensive this is, but it could be good to be able to toggle the shadows through a prop, from the json, in case there are a lot of wells and we don't want to render them twice.

src/lib/components/DeckGLMap/layers/wells/wellsLayer.js
```diff
@@ -0,0 +1,31 @@
+import { CompositeLayer } from "@deck.gl/core";
+import { GeoJsonLayer } from "@deck.gl/layers";
+
+export default class WellsLayer extends CompositeLayer {
```
alexandruandrei91, line 4: Doesn't need to be in the same review, it could be ported later, but ideally all new layers are in typescript.

alexandruandrei91 (approved):
lgtm :)

HansKallekleiv (approved):
lgtm !

PR #326, opened 2021-03-24
Display the property map value on hover

alexandruandrei91:
Currently it only displays a tooltip with the Z value from a single layer.


src/demo/DeckGLMapDemo.js
```diff
@@ -71,6 +71,7 @@ const DeckGLMapDemo = () => {
 
     const [jsonData, setJsonData] = React.useState(null);
     const [colormaps, setColormaps] = React.useState([]);
+    const [showCoords, setShowCoords] = React.useState(true);
```
aprilzhang, line 74: Looking at the usage here, it might be easier to use useMemo()?
const showCoords = useMemo(()=>typeof jsonData.showCoords === "undefined"
                ? true
                : jsonData.showCoords,[jsonData])
Btw, what does typeof jsonData.showCoords === "undefined" do? Is it for checking if a field is present in the json?

aprilzhang (approved):
Looks good to me, just one comment

src/demo/DeckGLMapDemo.js
alexandruandrei91: Yes, if the field doesn't exist in the json, showCoords should be true.

alexandruandrei91:
@aprilzhang Thanks for the feedback and for showing me useMemo :D. Now that I know about it, I've refactored things a bit in the demo.

PR #329, opened 2021-03-29
Pacakge updates

alexandruandrei91:
Update minimum node version to the latest LTS release (14).
Update to Webpack 5.
Update most other packages to their latest versions, except react and d3.
Fix all the new linting errors that popped up after the update. Some warnings remained.

This PR is not necessarily meant to go in as is, but it can be a platform for discussing the packages we want to update, and to what versions.


anders-kiaer (approved):
LGTM! 👍 🚀

aprilzhang (approved):
Looks good to me

PR #331, opened 2021-03-29
Update to React 17

alexandruandrei91:


src/lib/components/WellCompletions/components/Settings/HideZeroCompletionsSwitch.tsx
```diff
@@ -19,7 +19,6 @@ const HideZeroCompletionsSwitch: React.FC = React.memo(() => {
         <Switch
             label="Filter by completions"
             size="small"
-            enterKeyHint="Only show wells completed in at least one of the selected layers"
```
alexandruandrei91, line 22: This is the text that would appear instead of the `enter` key on a virtual keyboard, if the component was a text field, and it has to be a value from a specific set. See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint

aprilzhang: Thanks! I might have mistaken it with a tooltip hint. LGTM now.

aprilzhang (approved):

PR #333, opened 2021-03-30
Add test coverage for well completions component

aprilzhang:


inputSchema/wellCompletions.json
```diff
@@ -0,0 +1,64 @@
+{
```
alexandruandrei91, line 1: I guess the location for this file is ok for now, but we probably need to change it if we split the component. We'll probably have to see if we will have it with the component, with the backend or in a separate package.

alexandruandrei91 (commented):
When running `npm run verify` there's already 66 warnings. Probably not in this review, but we should try and get the number down since it's becoming hard to see the newly added ones.

src/lib/components/WellCompletions/components/Settings/__snapshots__/HideZeroCompletionsSwitch.test.tsx.snap
```diff
@@ -0,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
aprilzhang: This is a good idea. The snapshots are generated when running the test script in package.json, but .json file doesnt seem to support adding comments. Maybe I could add it somewhere in the developer wiki.

inputSchema/wellCompletions.json
aprilzhang: yes, I agree

package.json
```diff
@@ -59,12 +67,20 @@
...
         "@types/jest": "^26.0.20",
         "@typescript-eslint/eslint-plugin": "^4.19.0",
         "@typescript-eslint/parser": "^4.19.0",
         "ace-builds": "^1.4.12",
+        "addon-redux": "^1.1.0",
         "babel-loader": "^8.1.0",
         "copyfiles": "^2.0.0",
+        "core-js": "^3.9.1",
         "css-loader": "^5.2.0",
         "css-minimizer-webpack-plugin": "^1.3.0",
```
alexandruandrei91, line 85: This can be removed, since now we're back to optimize-css-..blabla

package.json
```diff
@@ -59,12 +67,20 @@
...
         "cssnano": "^4.1.11",
```
alexandruandrei91, line 86: You can remove cssnano as an explicit dependency, since that was there for the css minimizer.

package.json
```diff
@@ -75,19 +91,25 @@
...
+        "react-resize-panel": "^0.3.5",
+        "react-test-renderer": "^17.0.1",
+        "redux-mock-store": "^1.5.4",
+        "source-map-loader": "^0.2.4",
         "typescript": "^4.2.3",
         "url-loader": "^4.1.1",
-        "webpack": "^5.28.0",
-        "webpack-cli": "^4.5.0",
-        "webpack-dev-server": "^3.11.2"
+        "webpack": "^4.41.2",
```
alexandruandrei91, line 110: Maybe create an issue to switch back to webpack5, when Storybook properly supports it.

alexandruandrei91 (commented):
Building the storybook creates a folder called `storybook-static` with some files in it. I guess we don't want to commit that, so better put it in .gitignore.

alexandruandrei91 (approved):
LGTM 👍

PR #334, opened 2021-04-06
Fix the bandit command in the CI script

alexandruandrei91:
Bandit wasn't running on the `examples` folder because of a leading `/`.


anders-kiaer (approved):
👍

PR #335, opened 2021-04-06
Update cssnano to get rid of is-svg

alexandruandrei91:
cssnano is a dependency of css-minimizer-webpack-plugin, but since that hasn't yet been updated, I've explicitly added the latest cssnano package as a dev dependency.


anders-kiaer (approved):
👍

PR #345, opened 2021-04-19
Add kH display, open/close status and some other small fixes

aprilzhang:


src/lib/components/WellCompletions/utils/sort.ts
```diff
@@ -12,9 +12,9 @@ const createSortKeyFunction = (
             return (well) => well.name;
         case SORT_BY_STRATIGRAPHY_DEPTH:
             return (well) =>
-                well.zoneIndices.find(
-                    (_, index) => well.completions[index] > 0
-                );
+                well.completions.find(
+                    (_, index) => well.completions[index]!.open > 0
```
alexandruandrei91, line 16: This triggers an `Forbidden non-null assertion` warning.

alexandruandrei91 (commented):
It looks good, but I have some questions about the behavior, so I'll give you a call on teams.

src/lib/components/WellCompletions/utils/sort.ts
aprilzhang: Yes I saw that too. Not sure what's the best way to fix the warning as well.completions[index] is obviously not undefined.

aprilzhang: Added a comment to disable the warning for that line

alexandruandrei91 (approved):
👍

Issue #279, opened 2021-02-03
Drawing layer for the DeckGL map


alexandruandrei91:
Add support for drawing polygons/lines/circles on map and getting back the geometry in some format.

Evaluate: https://nebula.gl/

#251

PR #346, opened 2021-04-23
Add a drawing layer and setProps to the map component

alexandruandrei91:
The `setProps` property is used by Dash to get callbacks from the map component.

#279


src/demo/example-data/deckgl-map.json
```diff
@@ -1,110 +1,112 @@
...
-                {
-                    "@@type": "Hillshading2DLayer",
-                    "id": "hillshading-layer",
-                    "bounds": [432205, 6475078, 437720, 6481113],
-                    "valueRange": [2782, 3513],
-                    "opacity": 1.0,
-                    "image": "https://i.imgur.com/Dx8eseE.png"
-                },
-                {
-                    "@@type": "WellsLayer",
```
hkfb, line 69: The WellsLayer has been removed. Was this intentional?

alexandruandrei91: No, I was testing without it and forgot to add it back. I'll add it now.

hkfb (approved):

PR #348, opened 2021-04-26
Update deckgl

alexandruandrei91:
This also updates the `y18n` and `ssri` packages newer versions, reported as safe by Dependabot.


src/lib/components/WellCompletions/components/Settings/__snapshots__/HideZeroCompletionsSwitch.test.tsx.snap
```diff
@@ -2,26 +2,26 @@
 
 exports[`test hide zero completions switch snapshot test 1`] = `
 <label
-  className="sc-dUrnRO eauNHr"
```
aprilzhang, line 5: I am wondering why this gets updated each time. It might be to do with how styled component generates hashed className. I will try to fix this in another PR

aprilzhang (approved):
LGTM

PR #356, opened 2021-04-29
Better align bin `x` positions with input data in `PriorPosteriorDistribution`

### LINKED ISSUE #274: Rightmost bar lost in histogram in PriorPosteriorDistribution.
PriorPosteriorDistribution:
Rightmost bar lost in histogram. X-axis should be wide enough to fit all bars.
image

tnatt:
Bug fix to avoid that histogram bins are cut off in the PriorPosteriorDistribution component.  This is easiest accomplished by increasing the plots x range compared to the data range. 

Closes #274

### DIFF
--- a/src/lib/components/PriorPosteriorDistribution/utils/prior_posterior_distribution.js
+++ b/src/lib/components/PriorPosteriorDistribution/utils/prior_posterior_distribution.js
@@ -56,14 +56,9 @@ class D3PriorPosterior {
             ...this.data.values.map((values) => Math.max(...values))
         );
 
-        this.xscale_addition = (this.global_max - this.global_min) * 0.05;
-
         this.x_scale = d3
             .scaleLinear()
-            .domain([
-                this.global_min - this.xscale_addition,
-                this.global_max + this.xscale_addition,
-            ])
+            .domain([this.global_min, this.global_max])
             .range([0, this.width_plot]);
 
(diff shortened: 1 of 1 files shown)


src/lib/components/PriorPosteriorDistribution/utils/prior_posterior_distribution.js
```diff
@@ -56,9 +56,14 @@ class D3PriorPosterior {
             ...this.data.values.map((values) => Math.max(...values))
         );
 
+        this.xscale_addition = (this.global_max - this.global_min) * 0.05;
```
anders-kiaer, line 59: I'm wondering if we can avoid the magic addition constant `0.05` by trying changing this line
https://github.com/equinor/webviz-subsurface-components/blob/292791116c7afc2a8e5f24242e845b40dbdef120/src/lib/components/PriorPosteriorDistribution/utils/prior_posterior_distribution.js#L94

to something like:
```javascript
 .thresholds(d3.range(this.global.min, this.global.max, (this.global.max - this.global.min) / this.number_bins)); 
```

**Reference:** https://stackoverflow.com/questions/41396211/how-to-create-histogram-with-fixed-bin-width-in-d3-js-v4

Would that also solve the original issue? :thinking:

anders-kiaer (approved):

PR #358, opened 2021-04-29
Entered values for empty well attributes

lindjoha:
Fixes the problem of empty strings in the well attributes.


aprilzhang (approved):
LGTM

PR #359, opened 2021-04-29
Range mode improvement and add space in between value and unit

aprilzhang:


src/lib/components/WellCompletions/components/Settings/TimeRangeSelector.tsx
```diff
@@ -53,15 +58,28 @@ const TimeRangeSelector: React.FC = React.memo(() => {
...
-            <Slider
-                value={timeIndexRange}
-                onChange={onChange}
-                ariaLabelledby="range-slider-label"
-                min={0}
-                max={times.length - 1}
-                step={1}
-                outputFunction={outputFunction}
-            />
+            {rangeDisplayMode === "Off" && (
```
aprilzhang, line 61: This part was a bit tricky because the slider from the Equinor design system does not accept changing a single-step slider to a range slider once it's created. Also, it only accepts value/value range changes from user interaction (ie not from an internal state update). Therefore it has to be a complete rerender every time the range mode changes. Luckily this is not a rendering performance-critical component.

alexandruandrei91 (approved):
lgtm

src/lib/components/WellCompletions/components/Settings/TimeRangeSelector.tsx
anders-kiaer: Thanks @aprilzhang. The changes looks good. I could only find one further improvement suggestion: When time aggregation is `None` there are still two handles and a blue line between them (even though only one of the handles can be moved). Would it be better to only show one handle (or at least not the blue line in between) when time aggreation is `None`? If the EDS slider doesn't support this - we can create an issue in their repo and/or consider moving to another open source slider :slightly_smiling_face:

PR #360, opened 2021-04-30
Highlight and select wells

hkfb:
This will allow well selection to trigger updates in other components. Geometry from the selected well trajectory can be used to calculate intersections.


src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -8,6 +8,7 @@ export interface WellsLayerProps<D> extends CompositeLayerProps<D> {
     pointRadiusScale: number;
     lineWidthScale: number;
     outline: boolean;
+    selectedFeature: string;
```
alexandruandrei91, line 11: I think this is a GeoJSON feature object, not a string. I've added a package with geojson types, you can see it used in the drawing layer. For this case, you can probably use the `Feature` type from there. Also, you can replace the `WellDataType` with a type from that package (maybe FeatureCollection?), and just create a type for the custom properties you've added to the geojson.

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -49,9 +60,16 @@ export default class WellsLayer extends CompositeLayer<
             this.getSubLayerProps(properties)
         );
 
-        if (this.props.outline) return [outline, colors];
-        else return [colors];
+        // Highlight the selected well.
+        properties.data = this.props.selectedFeature;
+        properties.pointRadiusScale = properties.pointRadiusScale + 2;
+        properties.lineWidthScale = properties.lineWidthScale + 2;
+        const highlight = new GeoJsonLayer<WellDataType>(properties);
```
alexandruandrei91, line 67: You need to do `this.getSubLayerProps(properties)`, otherwise the inherited props will not get propagated (e.g. opacity).

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -19,6 +20,16 @@ export interface WellDataType {
     };
 }
 
+function handleClick(event) {
+    this.setLayerProps("wells-layer", { selectedFeature: event.object });
+    return true;
+}
+
+const defaultProps = {
+    onClick: handleClick,
```
alexandruandrei91, line 29: We probably need a mechanism to toggle the selection on and off, so we can disable it when drawing is active. Not necessarily in this PR.

alexandruandrei91 (approved):

PR #368, opened 2021-05-19
Stratigraphic levels and tree-filtering functionality in WellCompletion component

aprilzhang:


alexandruandrei91 (commented):
The code looks good to me.

alexandruandrei91:
A few things I've noticed while playing around with the component:
* __If I type an invalid attribute in the Filter by Attributes box, such as `name:`, the page will go blank.__
* Just a comment, no action point for this review, but I find it a bit weird to have two fields for filtering wells: one for well names and one for everything else. 
* `Wells selection criteria:` is a bit redundant in my opinion, as it just spells out what's in the Attributes box. I propose removing it, but check also with the others if they find it useful or not.
* If `Wells selection criteria:` stays, it would be nice to also show the well name you type in the search box as well.
* There's no indication on how to close the drawer, maybe have the filter button selected when the drawer is open?
* When opening the zone selection tree, it's just `All` that's showing up. I think it would be more useful to expand the first level by default.
* The zone tree component moves up and down when selecting/deselecting zones, based on what's in the box above, making it hard to quickly click on stuff. Not sure what to do about that though.. Also, just nitpicking, but clicking + or -, the nodes move left to right a bit (__very__ unimportant issue).
* There's some behavior that I guess is by design, but I find it a bit unintuitive and was wondering if we can customize it: When opening the zone selection tree and typing something like `.3` in the search box there, clicking on `All` selects/deselects all the zones, not just the filtered ones. Same for all its child nodes.

Let me know if you prefer to get this PR in and address some of the points here in later PRs.

aprilzhang:
> A few things I've noticed while playing around with the component:
> 
> * **If I type an invalid attribute in the Filter by Attributes box, such as `name:`, the page will go blank.**
> * Just a comment, no action point for this review, but I find it a bit weird to have two fields for filtering wells: one for well names and one for everything else.
> * `Wells selection criteria:` is a bit redundant in my opinion, as it just spells out what's in the Attributes box. I propose removing it, but check also with the others if they find it useful or not.
> * If `Wells selection criteria:` stays, it would be nice to also show the well name you type in the search box as well.
> * There's no indication on how to close the drawer, maybe have the filter button selected when the drawer is open?
> * When opening the zone selection tree, it's just `All` that's showing up. I think it would be more useful to expand the first level by default.
> * The zone tree component moves up and down when selecting/deselecting zones, based on what's in the box above, making it hard to quickly click on stuff. Not sure what to do about that though.. Also, just nitpicking, but clicking + or -, the nodes move left to right a bit (**very** unimportant issue).
> * There's some behavior that I guess is by design, but I find it a bit unintuitive and was wondering if we can customize it: When opening the zone selection tree and typing something like `.3` in the search box there, clicking on `All` selects/deselects all the zones, not just the filtered ones. Same for all its child nodes.
> 
> Let me know if you prefer to get this PR in and address some of the points here in later PRs.

Thank you @alexandruandrei91 , your comment is very helpful! 👍 
1) I can reproduce this as well and it seems from the error log, the crash is from the SmartNodeSelector. Wondering if there is any place that I could put a catch for this error. @rubenthoms Any thoughts?
2) Yes, I agree. Name after all can be considered as another attribute of the wells. 
3) This was an improvement request (from Ovyind I think), and I think the emphasis was to explain the "AND" and "OR" logic in between attribute keys and among attribute values. How about I change change "and" and "or" into the capital case or change their color? Or maybe move that into a tooltip?
4) and 5) Good idea, done 💯 
6) Yes, I agree too. I think in general I am not very keen on displaying all the selected values in that input box as it can grow very clumsy if we select a lot of sub-zones from different zones. This is one of the reasons that I was hoping that we could switch to use our dashboard tree 😄. Maybe there is some way to tweak the css style sheet to trim the space down but it is not very obvious how to get that done easily. 
7) Yes, I have the exact same confusion and I found there is already an issue opened in this repo that people don't like that design too: https://github.com/dowjones/react-dropdown-tree-select/issues/476. Not sure if this issue will be selected for development soon. It feels to me it's another reason to use the dashboard tree.

I think I can bring up 2) and 3) in our next meeting with Equinor on Friday. 3) Should be a pretty small change that can get squeeze into the release. Depend on what we decide to do with 2), 2) might need to wait until the next update.

alexandruandrei91:
@aprilzhang , sounds like a good plan, better to get this PR in then. For 2) and 3), if you have time to work on it before Friday you can call for a meeting with Equinor or ask on Slack.

alexandruandrei91 (approved):
👍

PR #371, opened 2021-05-21
Add piechart layer

nilscb:


src/lib/components/DeckGLMap/Map.tsx
```diff
@@ -104,6 +108,12 @@ const Map: React.FC<MapProps> = (props: MapProps) => {
                     }}
                     ref={deckRef}
                     onHover={onHover}
+                    onViewStateChange={  ({viewState}) => {
```
alexandruandrei91, line 111: I don't think you need to manually do this. See my comment below.

src/lib/components/DeckGLMap/layers/piechart/pieChartLayer.ts
```diff
@@ -0,0 +1,96 @@
...
+export interface PieChartLayerProps<D> extends CompositeLayerProps<D> {
+    pieRadius: number;
+}
+
+const defaultProps = {};
+export default class PieChartLayer extends CompositeLayer<
+    Feature,
+    PieChartLayerProps<Feature>
+> {
+    renderLayers(): GeoJsonLayer<Feature>[] {
```
alexandruandrei91, line 15: You can get the viewport parameters from the layer context (`this.context.viewport`) and use that for computing the radius.
Check the viewport parameters here: https://deck.gl/docs/api-reference/core/viewport

src/lib/components/DeckGLMap/layers/piechart/pieChartLayer.ts
```diff
@@ -0,0 +1,96 @@
...
+            coordinates.push([xx, yy]);
+        }
+        const rad = (end_a * (2.0 * Math.PI)) / 360.0;
+        const xx = R * Math.cos(rad) + x;
+        const yy = R * Math.sin(rad) + y;
+        coordinates.push([xx, yy]);
+        coordinates.push([x, y]);
+
+        const col = colors[i % colors.length]; // wraps around if not enough colors
+        json.features.push({
```
alexandruandrei91, line 87: Better create geojson object in this function and return it instead of passing by reference and fill it in here. Javascript is garbage collected and everything passes by reference, so you will not make any copies by returning the object.

src/demo/example-data/piechart.json
```diff
@@ -0,0 +1,8 @@
+{
+    "pies": [ {"x": 433600, "y": 6477600, "R": 99, "fractions": [21, 15, 45, 32]},
+              {"x": 434556, "y": 6478951, "R": 99, "fractions": [18, 11, 15]},
+              {"x": 437086, "y": 6477198, "R": 99, "fractions": [12, 28, 22, 9]}
+            ],
+
+     "colors": [[255, 0, 0], [0, 255, 0], [0, 0, 255], [255, 255, 0], [0, 255, 255], [255, 0, 255]]
```
alexandruandrei91, line 7: I guess this is already discussed, but instead of just colors, this could include labels as well, and maybe any kind of data. 
```
"fractionAttributes"= [{"color": [255, 0, 0], "label":"blabla"}, {"color...
```
Also, the mapping between the fractions and the fraction color is based on the order, I guess? So it needs to work in cases where a fraction is 0, to make it possible to have cases like this, for example: 3 slices and 5 colors, where slice 1 maps to first color, slice 2 to the third color and slice 3 to the last color. You could also change the format to specify explicitly which fraction attribute to use for which fraction, up to you.

nilscb:
Abandoned

PR #379, opened 2021-05-31
New piechart layer

nilscb:


src/lib/components/DeckGLMap/layers/piechart/pieChartLayer.ts
```diff
@@ -0,0 +1,130 @@
...
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    updateState({ props, oldProps, context, changeFlags }) {
+        super.updateState({ props, oldProps, context, changeFlags });
+
+        const zoom = context?.viewport?.zoom ?? 1;
+
+        if (props.data?.pies !== undefined) {
+            // Data is read and it is in proprietary format
+            // We need to store this data. Zoom is realtive to it.
+            // Store in new prop 'data_orig'
```
alexandruandrei91, line 30: This approach will not work anymore and it's not ideal to set new props from inside the prop anyway. Not sure yet what would be a good alternative.

src/lib/components/DeckGLMap/layers/piechart/pieChartLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+        const pie_polygons = makePie(pie, data.properties);
+        for (const polygon of pie_polygons) {
+            polygons.push(polygon);
+        }
+    }
+    return polygons;
+}
+
+// return array of one pie's polygon's
+function makePie(pie, properties) {
```
alexandruandrei91, line 69: To add types to this function arguments, you would need to extract `Pie` and `PieSliceProps` from `PieData` into separate types, and reuse them both in `PieData` and here. A bit more involved, so I guess it can be changed later.
We can easily add the return type:
```suggestion
function makePie(pie, properties): Position[] {
```

alexandruandrei91 (commented):
Just a few minor changes to go :)

alexandruandrei91 (approved):

PR #386, opened 2021-06-03
Update well completion drawer resize

aprilzhang:


lindjoha (approved):
I have tested it on the WellCompletions plugin and it looks fine to me.

PR #389, opened 2021-06-07
Refactor map state management

alexandruandrei91:
* Add a react hook for updating the map state in the deckgl map component
* Add a map state class in the deckgl python example


aprilzhang (approved):
The javascript part looks good to me 👍

Issue #361, opened 2021-05-05
Visualization of simulation group tree 


anders-kiaer:

anders-kiaer:
**Maybe useful references:**
https://d3-wiki.readthedocs.io/zh_CN/master/Tree-Layout/
https://www.d3-graph-gallery.com/dendrogram.html

asnyv:
Comment: I would prefer to BHP, BP9 and etc to be their own leaves (we have all the information: rates (from well rates), pressures and so on. Actually it might only be relevant from the backend / data input, the component doesn't have to care about this at all?

lindjoha:
Other things we can include as options:
- Display VFP table number on the edges
- Display Well control mode (WMCTL) on the well nodes
- Split gas rates in reservoir gas and gas lift

nicktalb:
*Relative v Absolute values*
Would it be useful to be able to show relative allocations rather than absolute for each parent and children set? For large networks you could end up with very thick lines to the left and thin to the right. For ensemble based visualizations where stats are required it would be near impossible to see the variation over the ensemble if percentiles were visualized as bands.

*OWG combined*
You could have RGB lines with relative thickness. This might only make sense for history and/or single realization. For ensembles it is probably too much information?

aprilzhang:
Just to copy some comments from Knut:

There are many use cases for such data tree:
- For visualization components selecting which wells,
- For creating sections need  not the only list of wells but also the order of wells
- Creating well lists for use in modelling.

1) Most modelling create a well list called Used in model not used in model
2) Many history matching projects have similar list wells used in history match, wells not used in history match

There are some important features of such trees one could consider:
- Can wells be in several branches at the same time, i,e can wells be in two groups at the same time 
- does the well groups support ordering of the wells

Issue #388, opened 2021-06-07
WellCompletions timeslider too small with many time steps


lindjoha:
When there is many time steps in the well completions component (let say >20), it can be difficult to see which time step is selected, and to scroll one timestep at the time. The date is written in text, but that is very small and can be hard to see. 

I suggest a drop down with all dates, making it possible to select a specific time step.

PR #391, opened 2021-06-08
Basic version of Group Tree Visualization component

aprilzhang:
This is the very first version of the group tree component and it contains three basic settings: iteration selector, rate selector, and time step slider. The input data is randomly generated by group-tree.py and it is unrealistic dummy data.

group-tree.py is originally copied (with slight modification) from https://github.com/equinor/webviz_petech/blob/master/visualizations/dynamic_tree/examples/dynamic_tree_example.py.
Also, the core rendering part of the current component is using grouptree.js which is from https://github.com/anders-kiaer/webviz-subsurface-components/blob/dynamic_tree/src/lib/components/DynamicTree/group_tree.js

The full design of further features is listed in https://github.com/equinor/webviz-subsurface-components/issues/361.

At the current stage, the three flow rate options are hardcoded in the input data. As the next step, I think it would be good if we could use a map to store data at each node so that it offers the flexibility of providing a more generic data display.

Also, Nick is quite keen on using WebGL to do the tree render so that it minimizes lag for displaying large data. Tempest Dashboard has a quite mature library to simplify the experience of setting up WebGL rendering from scratch (because almost all our components are using WebGL).


src/lib/components/GroupTree/components/GroupTree.jsx
```diff
@@ -0,0 +1,149 @@
+import * as d3 from "d3";
+import Slider from "../../../shared/slider";
+import GroupTree from "./Plot/group_tree";
+
```
alexandruandrei91, line 4: Is this file used anywhere? If so, why is it jsx?

src/lib/components/GroupTree/components/GroupTree.jsx
```diff
@@ -0,0 +1,149 @@
...
+<script>
+    webviz_petech.initDynamicTree(
+        {'iterations': {{element.get_json_dump('iterations')}},
+         'iteration_names': {{element.get_json_dump('iteration_names')}} },
+        "#{{element.containerId}}grouptree_slider",
+        "#{{element.containerId}}iteration_select",
+        "#{{element.containerId}}flowrate_select",
+        "#{{element.containerId}}grouptree_tree");
+</script>
```
alexandruandrei91, line 49: Dead code or example?

src/lib/components/GroupTree/components/Plot/group_tree.js
```diff
@@ -0,0 +1,464 @@
+import * as d3 from "d3";
```
alexandruandrei91, line 1: As you've mentioned, it would be good to have this ported to typescript and use WebGL. Not for this review, though..

src/lib/components/GroupTree/components/Settings/DateTimeSlider.tsx
```diff
@@ -0,0 +1,79 @@
+import {
```
alexandruandrei91, line 1: This sort of components are pretty generic and can be useful for many things, so in the future we should probably have them in webviz core components.

alexandruandrei91 (commented):
A few other notes from playing around with the demo:
* When hovering over a node, the cursor changes to the `insert text` mode instead of `hand`, not properly indicating that the node is clickable (not sure if `insert node` and `hand` are proper cursor mode names, but I hope you get the point 😄)
* The collapsed nodes have no indication that they are collapsed and there's more data if you click on them
* When changing the `Flow Rate`, the line colors change, but not the node colors (by design?)
* When changing the `Flow Rate`, the lines width changes, but not the values on the lines or in the nodes (by design?)

Depending on the difficulty of adding this stuff and if they are bugs or by design, it's up to you if you want to do the changes in this review or later.

src/lib/components/GroupTree/components/GroupTree.jsx
aprilzhang: Well spotted, removed.

src/lib/components/GroupTree/components/Settings/DateTimeSlider.tsx
aprilzhang: Yes, I agree too. Most of the UI components are directly using eds or material UI components, but for this DateTime selector, I think it makes sense to move it into the webviz-core-component, especially because we now have other improvements requests (e.g. https://github.com/equinor/webviz-subsurface-components/issues/388), and therefore it might grow into a more complex component.

src/lib/components/GroupTree/components/GroupTree.jsx
aprilzhang: Done

src/demo/example-data/group-tree.py
```diff
@@ -0,0 +1,50 @@
...
+def create_node(depth, child_no):
+    is_inactive = random.randint(0, 10)
+    return {
+        "name": "{}_{}".format(depth, child_no),
+        "children": [],
+        "pressure": random.randint(8, 16) * is_inactive,
+        "oilrate": random.randint(1, 200) * is_inactive,
+        "waterrate": random.randint(1, 20) * is_inactive,
+        "gasrate": random.randint(1, 20) * is_inactive,
+        "grupnet": grupnetchars[random.randint(0, len(grupnetchars) - 1)]
```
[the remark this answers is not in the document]
aprilzhang: Thank you, done. I have changed the script to calculate the sum of the children's values.

aprilzhang:
Thank you for your review @alexandruandrei91.
Yes the tree plot itself is waiting for a huge refactoring and the features you saw at the moment are by design (from group_tree.js which I haven't made any changes). I imagine the plot will look quite different once we change to WebGL.

alexandruandrei91 (approved):
Looks great 👍

PR #395, opened 2021-06-11
Added log curve to Map componet

shadab-skhan:


src/lib/components/DeckGLMap/Map.tsx
```diff
@@ -82,8 +83,15 @@ const Map: React.FC<MapProps> = ({
...
                         isDragging ? "grabbing" : "default"
                     }
-                    getTooltip={(info: PickInfo<unknown>): string | null => {
-                        return (info.object as Feature)?.properties?.name;
+                    getTooltip={(
+                        info: PickInfo<unknown> | WellsPickInfo
+                    ): string | null | undefined => {
+                        const well_info = info as WellsPickInfo;
+                        if (well_info) {
+                            return well_info.logName;
```
alexandruandrei91, line 91: `info as WellPickInfo` will not return `null` if `info` is not a `WellPickInfo`. I think it should be something like this:
```suggestion
                        if ((info as WellsPickInfo)?.logName) {
                            return (info as WellsPickInfo).logName;
```

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -14,18 +16,104 @@ export interface WellsLayerProps<D> extends CompositeLayerProps<D> {
     outline: boolean;
     selectedFeature: Feature;
     selectionEnabled: boolean;
+    logData: string;
```
alexandruandrei91, line 19: This could come either as an url (so string), or directly as the object data. You can move the `LogCurveDataType` above `WellsLayerProps<D>` and have the `logData: string | LogCurveDataType`.

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -14,18 +16,104 @@ export interface WellsLayerProps<D> extends CompositeLayerProps<D> {
...
+export interface WellsLayerProps<D> extends CompositeLayerProps<D> {
+    pointRadiusScale: number;
+    lineWidthScale: number;
+    outline: boolean;
+    selectedFeature: Feature;
+    logData: string;
+    logName: string;
+    logRadius: number;
+    logCurves: boolean;
+}
```
alexandruandrei91, line 45: Duplicated code, the interface is already defined above.

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -83,8 +171,62 @@ export default class WellsLayer extends CompositeLayer<
...
+        }
+        return layers;
+    }
+
+    getPickingInfo({ info }) {
+        if (info.object == null || info.object.data == undefined) return info;
+
+        const trajectory = info.object.data[0];
+
+        let min_d = 99999;
```
alexandruandrei91, line 210: Use `Number.MAX_VALUE` instead of `99999`, as that could be too small, considering it's squared distance.

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -83,8 +171,62 @@ export default class WellsLayer extends CompositeLayer<
...
+        let vertex_index = 0;
+        for (let i = 0; i < trajectory.length; i++) {
+            const d = squared_distance(trajectory[i], info.coordinate);
+            if (d > min_d) continue;
+
+            vertex_index = i;
+            min_d = d;
+        }
```
alexandruandrei91, line 218: This function will be called for every mouse move, so I'm a bit concerned that this approach will be too slow. We can keep it for now and profile later.

alexandruandrei91 (commented):
Looks nice :)

Except for the comments below, when visualizing wells there are some portions of the log that are black. Is it because there's no log value there? And if so, could it be transparent instead of black?

Also, I think that after this review we should have some benchmark of the layer, checking how big the log data is for something like 1000 wells and how many wells with logs can be visualized at an interactive FPS. Not for this review, though.

alexandruandrei91:
Oh, and I forgot to mention, there's two new warnings in the wells layer:
```
/home/aalexandru/dev/equinor/webviz-subsurface-components/shadab/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
  205:5   warning  Missing return type on function          @typescript-eslint/explicit-module-boundary-types
  205:20  warning  Object pattern argument should be typed  @typescript-eslint/explicit-module-boundary-types
```

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
shadab-skhan: Done

src/lib/components/DeckGLMap/Map.tsx
shadab-skhan: Done

src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -99,7 +99,7 @@ export interface WellsPickInfo extends PickInfo<Feature | LogCurveDataType> {
 }
 
 export default class WellsLayer extends CompositeLayer<
-    Feature | LogCurveDataType,
```
shadab-skhan: Used 'unknown' type for data object of WellsLayer, as in layerTools.tsx pathLayerProps function uses Layer<unknown,P> type and it's causing type incompatibility.

hkfb:
LGTM. In the future it probably would be good to have a look at validating the well log data before parsing. Eg. https://github.com/jquense/yup

hkfb (approved):

PR #397, opened 2021-06-15
New fault polygon layer

nilscb:


hkfb (changes requested):
It's supposed to be a fault polygons layer, and I think the name should reflect that. I think the distinction is important.

src/lib/components/DeckGLMap/layers/fault_lines/faultLinesLayer.ts
```diff
@@ -0,0 +1,28 @@
...
+const getColor = (d: Feature): RGBAColor => {
+    const c: RGBAColor = d?.properties?.color;
+    const r = c[0] ?? 0;
+    const g = c[1] ?? 0;
+    const b = c[2] ?? 0;
+    return [r, g, b, 30]; // make fill color transparent
+};
+
+const defaultProps = {
+    pickable: true,
```
hkfb, line 15: Maybe use `filled: true`

shadab-skhan (approved):
Looks good to me

hkfb (approved):

PR #398, opened 2021-06-16
Adjusted stratigraphy height and added dropdown(s) for the time slider

aprilzhang:
When the time aggregation is switched on, there will be two dropdowns for selecting the start and end date.
One long-term possibility from @anders-kiaer: the time label is shown above/below each handle (useful on its own to show selected time), and that the time label can be clicked on in order to open one of these typical date picker react component. At some point, we could move the slider to another repository ( webviz-core-componetnsor other) and make it a general component and invest more time in it, based on learnings now.


anders-kiaer:
Saved the (future improvement) suggestion as https://github.com/equinor/webviz-core-components/issues/146. 🙂

hkfb (approved):

PR #406, opened 2021-06-22
Well completion unit tests

avanti-em:
Updated the previous snapshots and improved the code coverage.


src/lib/components/WellCompletions/components/Settings/WellAttributesSelector.test.tsx
```diff
@@ -18,13 +21,14 @@ describe("test well attributes selector", () => {
         render(<WellAttributesSelector />, {
             wrapper: Wrapper,
         });
-        
-        const attributeFilter = screen.getByRole('textbox')
-        userEvent.type(attributeFilter, 'type:Injector{enter}')
-        await waitFor(() => expect(screen.getByText(/well selection criteria: "type" is "injector"/i)).toBeVisible())
```
aprilzhang: This is not going to happen because it requires the state to be updated in the redux store to be able to see the criteria text update, but this testStore.dispatch is just a mocked-up function.

src/lib/components/WellCompletions/components/Settings/ZoneSelector.test.tsx
```diff
@@ -31,16 +31,39 @@ describe("test Zone Selector", () => {
...
-        expect(testStore.dispatch).toBeCalledWith({
-            payload: ['zone1'],
+        expect(testStore.dispatch).toHaveBeenNthCalledWith(1, {
+            payload: [],
+            type: "ui/updateFilteredZones",
+        });
+        expect(testStore.dispatch).toHaveBeenNthCalledWith(2, {
+            payload: [],
+            type: "ui/updateFilteredZones",
+        });
```
aprilzhang, line 49: I actually dont understand why the handler get called three times, it looks like it's when the click, {down}, and {enter} actions happen on the SmartNodeSelector. It feels like a small bug to me as it might drag down the performance, might worth asking @rubenthoms.

src/lib/components/WellCompletions/test/TestWrapper.tsx
```diff
@@ -13,5 +16,9 @@ export const Wrapper = ({
 }: {
     children: JSX.Element;
 }): JSX.Element => {
-    return <Provider store={testStore}>{children}</Provider>;
+    return (
```
aprilzhang, line 19: I realized another layer is needed to provide the essential test input data for some of the components that depend on that data.

aprilzhang (approved):

src/lib/components/WellCompletions/components/Settings/ZoneSelector.test.tsx
avanti-em: Yes ....It seems like that the keyboard action of down and enter had made the call two times. Thanks for bringing that to notice .

PR #436, opened 2021-06-30
Drogon datasets for the GroupTree component.

lindjoha:
Have added two data sets for the GroupTree component based on Drogon:

1. grouptree_exampledata_drogon.json: dataset with the current data format
2. grouptree_suggested_format_drogon.json: The same dataset, but with a suggested new data format

The reason for the suggested format is that the tree normally changes much less often than the data.


nilscb (approved):
Looks ok to me

PR #438, opened 2021-07-02
Redesigned information card

shadab-skhan:
Redesigned information card
Used backend generated colors for discrete logs


react/src/lib/components/DeckGLMap/layers/utils/propertyMapTools.ts
```diff
@@ -1,7 +1,7 @@
-import { PickInfo } from "@deck.gl/core/lib/deck";
 import { RGBColor } from "@deck.gl/core/utils/color";
+import { LayerPickInfo } from "../../components/InfoCard";
```
alexandruandrei91, line 2: In the future, the map component could be packaged by itself, so we shouldn't have dependencies on other internal components unless they are really used in the map.

You can move the needed classes (LayerPickInfo and PropertyDataType) into a file in layers/utils.

react/src/lib/components/DeckGLMap/layers/utils/propertyMapTools.ts
```diff
@@ -1,7 +1,7 @@
...
 
-export interface BitmapPickInfo extends PickInfo<unknown> {
+export interface BitmapPickInfo extends LayerPickInfo {
```
alexandruandrei91, line 4: Nothing from the LayerPickInfo is being filled in in the bitmap layer, so probably it doesn't need to extend it.
From the current usecase, LayerPickInfo includes information needed only for the Wells layer, so you can maybe add the fields to the WellPickInfo struct directly.

react/src/lib/components/DeckGLMap/components/InfoCard.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { Button, Icon } from "@equinor/eds-core-react";
+import { arrow_drop_up, arrow_drop_down } from "@equinor/eds-icons";
+
+import { PickInfo } from "deck.gl";
+import { PropertyMapPickInfo } from "../layers/utils/propertyMapTools";
+
+Icon.add({ arrow_drop_up, arrow_drop_down });
+
+export interface LayerPickInfo extends PickInfo<unknown> {
+    properties?: PropertyDataType[];
```
alexandruandrei91, line 20: Does this have to be an array? Afaik, the multipicking method already returns an array of PickInfo objects.

react/src/lib/components/DeckGLMap/layers/utils/propertyMapTools.ts
shadab-skhan: I planned to keep the common PickInfo type for all the layers required by the new info card.
But true, from the current use case only the wells layer needs this.

alexandruandrei91 (approved):
👍

PR #440, opened 2021-07-02
Add `npm` deploy to CI/CD workflow

anders-kiaer:
Setup copied from `webviz-core-components`. A bit difficult to test before doing a tag in GitHub, so there might be a follow up PR if it doesn't work on first try. :laughing:


alexandruandrei91 (approved):
👍

PR #445, opened 2021-07-07
Add full state prop to the map component

alexandruandrei91:
Because the patch-only system would require to keep the map state in both js and python land, I've added the map state back as a prop. The patch prop is also useful to have because it makes it easier to work with from the python side, so now the map component has both props: `deckglSpecBase` and `deckglSpecPatch`. The displayed map is formed from the base + patch.

In a following PR I will also update the python example.


anders-kiaer (approved):
LGTM! Looking forward to seeing this reflected in the Python examples :rocket:

PR #449, opened 2021-07-08
Fix postinstall on windows command prompt

t0oF:
Fixes #449


anders-kiaer (approved):
Thanks @t0oF, LGTM! 👍 

Apparently forward slashes in paths (instead of `\`) is not a problem on Windows command prompt?

t0oF:
`/` work fine with `copyfiles`
BTW, `copyfiles my/file.txt some/dir/` does also create the directory if missing.

Issue #448, opened 2021-07-08
npm run postinsall fails on windows


t0oF:
```
npm run postinstall
```
fails on windows using a command prompt as terminal:
```
> postinstall
> cp ./package.json ../webviz_subsurface_components/ && npm run setup_deckgl_types

'cp' is not recognized as an internal or external command,
operable program or batch file.
```

anders-kiaer:
Closed by #449.

PR #450, opened 2021-07-09
Remove copy step from `postinstall`

anders-kiaer:
Related to #448.

`postinstall` is a special script name called by `npm` automatically when installing this package from `npm` registry.

Therefore, split out the command only useful for Dash component building (which is not needed when using this repository as a pure JavaScript dependency).

Also updated related steps in `README.md` (and added `npm` badge).


alexandruandrei91 (approved):
LGTM

PR #457, opened 2021-07-13
Simple Well Log View based on videx-wellog example

Vladimir-Kokin:
- Modify videx-wellog example to work as REACT component.
- Add Demo react component
- Convert code from JavaScript to TypeScript.
- Add real data files in JSONWellLogFormat.org format
- Add DualScaleTrack to show MD and TVD scales simultaneously
- Add example data file selector to Demo component


Vladimir-Kokin:
Closed due to some lint messages

PR #458, opened 2021-07-14
added a button to toggle the code panel

amroahmed90:
Saw that this task is left to-do


alexandruandrei91:
Hi Amro, thank you for contributing!
The js linting fails in the CI, could you please update with the fixes? You can check it locally by running `npm run validate --prefix=react`. If you get a lot of DeckGL type errors it means that the DeckGL types are missing and you need to run `npm run postinstall --prefix=react`.

amroahmed90:
Hello Alex,
Thanks for the tip. I will look into it. Have a great weekend.

hkfb (commented):
This demo is not relevant any longer, as we are now using Storybook.

PR #460, opened 2021-07-16
Spline refine well path

nilscb:


react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -70,10 +71,12 @@ export default class WellsLayer extends CompositeLayer<
     }
 
     renderLayers(): (GeoJsonLayer<Feature> | PathLayer<LogCurveDataType>)[] {
+        const data = splineRefine(this.props.data);
```
alexandruandrei91, line 74: Maybe add a prop to toggle spline interpolation on and off (default on), in case the trajectories have enough sample points or for performance reasons.

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +352,109 @@ function getLogProperty(
...
+function CatmullRom(
+    p0: number,
+    p1: number,
+    p2: number,
+    p3: number,
+    t: number
+): number {
+    const t2 = t * t;
+    const t3 = t * t * t;
+    return  0.5 *(2*p1 + (-p0 + p2) * t + (2*p0 - 5*p1 + 4*p2 - p3) * t2 + (-p0 + 3*p1- 3*p2 + p3) * t3); // eslint-disable-line
```
alexandruandrei91, line 365: Why disable eslint? If it's because the line is too long, better split the formula on multiple lines.

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +352,109 @@ function getLogProperty(
...
+
+                newCoordinates.push([x, y]);
+                newMds[0].push(md);
+            }
+        }
+
+        newCoordinates.push(coords[n - 1]);
+        newMds[0].push(mds[n - 1]);
+
+        data[well_no]['geometry']['geometries'][1]['coordinates'] = newCoordinates; // eslint-disable-line
```
alexandruandrei91, line 456: Why disable eslint for this line?

alexandruandrei91 (commented):
The result looks nice. 
Can we change the script that extracts the trajectories into geojson to have less points? Not in this review, ofc.

nilscb:
Abondonthis for an improved version

PR #464, opened 2021-07-20
Catmull 3D spline interpolation

nilscb:


react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +353,159 @@ function getLogProperty(
...
+        const P_first: Position3D = [x0, y0, z0];
+
+        // Point after last.
+        const xn = coords[n - 1][0] - coords[n - 2][0] + coords[n - 1][0];
+        const yn = coords[n - 1][1] - coords[n - 2][1] + coords[n - 1][1];
+        const zn = coords[n - 1][2] - coords[n - 2][2] + coords[n - 1][2];
+        const P_n: Position3D = [xn, yn, zn];
+
+        //const md_first = 0.25 * (mds[0][0] - mds[0][1]) + mds[0][0];
+        //const md_n = 0.25 * (mds[0][n - 1] - mds[0][n - 2]) + mds[0][n - 1];
```
alexandruandrei91, line 439: Dead code? Either remove it or add a comment of what it does and when it should be enabled, if ever.

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +353,159 @@ function getLogProperty(
...
+        const newCoordinates: [Position3D?] = [];
+        const newMds: number[][] = [];
+        newMds.push([]);
+
+        for (let i = 0; i < n - 2; i += 1) {
+            let P0: Position3D, P1: Position3D, P2: Position3D, P3: Position3D;
+            //let md0: number;
+            let md1: number;
+            //let md2: number;
+            //let md3: number;
```
alexandruandrei91, line 450: Same as above

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +353,159 @@ function getLogProperty(
...
+            if (i === 0) {
+                P0 = P_first;
+                P1 = coords[i + 0];
+                P2 = coords[i + 1];
+                P3 = coords[i + 2];
+
+                //md0 = md_first;
+                md1 = mds[0][i + 0];
+                //md2 = mds[0][i + 1];
+                //md3 = mds[0][i + 2];
```
alexandruandrei91, line 461: Same as above

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +353,159 @@ function getLogProperty(
...
+            } else if (i === n - 3) {
+                P0 = coords[i + 0];
+                P1 = coords[i + 1];
+                P2 = coords[i + 2];
+                P3 = P_n;
+
+                //md0 = mds[0][i + 0];
+                md1 = mds[0][i + 1];
+                //md2 = mds[0][i + 2];
+                //md3 = md_n;
```
alexandruandrei91, line 471: Same as above

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +353,159 @@ function getLogProperty(
...
+            } else {
+                P0 = coords[i + 0];
+                P1 = coords[i + 1];
+                P2 = coords[i + 2];
+                P3 = coords[i + 3];
+
+                //md0 = mds[0][i + 0];
+                md1 = mds[0][i + 1];
+                //md2 = mds[0][i + 2];
+                //md3 = mds[0][i + 3];
```
alexandruandrei91, line 481: Same as above

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -348,3 +353,159 @@ function getLogProperty(
...
+
+    if (data["features"] === undefined) {
+        return;
+    }
+
+    const no_wells = data["features"].length;
+    for (let well_no = 0; well_no < no_wells; well_no++) {
+        const mds = data["features"][well_no]["properties"]["md"];
+
+        const coords = data["features"][well_no]["geometry"]["geometries"][1]["coordinates"]; // eslint-disable-line
```
alexandruandrei91, line 419: Is it possible that any of these fields are not there for some trajectories? Or maybe there aren't at least 2 geometries, so `...["geometries"][1]["blabla"]` will throw an error? If the format guaranties that you have all those fields and number of geometries, then there's no action to be taken here, otherwise you probably need to add some checks.

nilscb: The dead code will be fixed in the next pull request. I will also then add a test to ensure data["features"] is defined. As for the other fields they are guaranteed by the format I think. The number "1" there for instance refers to the trajectory itself and "0" is the well head both of which has to be there

alexandruandrei91: The well heads and trajectories together might not work in the future when adding other well markers (where the well intersects surfaces, for example), but that's a discussion for another time :)

alexandruandrei91 (approved):

Issue #466, opened 2021-07-21
Add npm scripts to generate npm component package


t0oF:
The package currently published on www.npmjs.com (as of version 0.4.4) can not be used by an npm project: it does currently not contain javascript code.

PR #467, opened 2021-07-21
Add npm scripts to generate npm components package

t0oF:
The package can be generated only with npm (no python script).
It has been used by an npm project.


react/package.json
```diff
@@ -26,7 +27,9 @@
         "storybook": "start-storybook -p 6006",
         "build-storybook": "build-storybook",
         "test": "jest --coverage",
-        "test:watch": "npm test -- --watch"
+        "test:watch": "npm test -- --watch",
+        "build:npm": "rimraf dist && tsc --declaration && npm run copy-files",
```
t0oF, line 31: build:npm and copy-files scripts are inspired from webviz-core-components

react/src/lib/components/WellCompletions/components/WellCompletionComponent.tsx
```diff
@@ -7,7 +7,8 @@ import DataProvider from "./DataLoader";
 import WellCompletionsViewer from "./WellCompletionsViewer";
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const inputSchema = require("../../../../../inputSchema/wellCompletions.json");
+const inputSchema = require("../../../inputSchema/wellCompletions.json");
```
t0oF, line 10: I couldn't find any other way to access this file both from webpack and when using the package as a library package.
When building an npm package, we want to include only the library part. Thus, inputSchema has to be at that level.

Note:
* in an ideal world, the demo and the library part should be in separate packages
* if there is a better solution I'll be happy to apply it, but I couldn't find any

react/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
     "compilerOptions": {
         "outDir": "./dist/", // path to output directory
+        "declaration": true,
```
t0oF, line 4: This could actually be removed: I added the flag in the npm script

t0oF:
Proposal to fix #466

anders-kiaer (approved):

PR #468, opened 2021-07-21
Spline:

nilscb:
* added spline interploations also for md values
 * added boolean property "refined"
 * spline functions in separate file


alexandruandrei91 (commented):
There are a few linting warnings, please fix them if possible.
On my side, linting takes about 1 minute. I'll do a PR now where I get that down to 20 seconds on my system. Let me know if you still have problems with eslint running for hours :D

alexandruandrei91 (approved):

PR #474, opened 2021-07-26
Map component tests

avanti-em:


react/src/lib/components/DeckGLMap/test/TestWrapper.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { Provider } from "react-redux";
+import { createStore } from "../redux/store";
+import { Operation } from "fast-json-patch";
+
+
+export const testStore = createStore({}, (patch: Operation[]) => void);
```
alexandruandrei91, line 7: `void` is just a type, it cannot be the body of that callback. Provide an empty callback `{}`, or something like below, so the linter doesn't complain that you're not using `patch`. After this, you have a bunch of other linting errors and warnings. Run `npm run format --prefix=react` to fix most of them.
```suggestion
export const testStore = createStore({}, (patch: Operation[]) => {
    patch;
});
```

avanti-em: Thank you @alexandruandrei91 for fixing it up, I will work on the other test related issues now.

PR #475, opened 2021-07-27
FixMdInterpolation

nilscb:


shadab-skhan (approved):
Tested it locally, I found no issue. Change is good to be in.

PR #495, opened 2021-08-12
Map component storybook

shadab-skhan:


react/package.json
```diff
@@ -69,8 +69,8 @@
         "lodash": "^4.17.21",
         "mathjs": "^9.4.2",
         "nebula.gl": "^0.22.3",
-        "react": "^17.0.2",
-        "react-dom": "^17.0.2",
+        "react": "^16.13.0",
```
shadab-skhan, line 72: is it okay to downgrade react? I had to do this install manager-webpack5.
It doesn't get installed with current react version.

react/src/lib/components/DeckGLMap/DeckGLMap.stories.jsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
```
shadab-skhan, line 1: Its WIP, you can ignore this file for now.

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -40,12 +39,14 @@ const Map: React.FC<MapProps> = ({
...
 
     React.useEffect(() => {
+        console.log("use effect create store")
         store.current = createStore(deckglSpec, setSpecPatch);
     }, [setSpecPatch]);
 
     const [specObj, setSpecObj] = React.useState(null);
 
     React.useEffect(() => {
+        console.log("Use effect sepec obj update");
```
shadab-skhan, line 49: This hook is not getting called when I do selection in the tool bar to hide a layer.

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -82,9 +83,28 @@ const Map: React.FC<MapProps> = ({
     );
 
     React.useEffect(() => {
+        console.log("Use effect dispatch");
```
shadab-skhan, line 86: This hook is not getting called when I do selection in the toolbar to hide a layer.

alexandruandrei91 (commented):
The MapWrapper component should either be a wrapper around the Map component, and provide exactly the same interface as the Map component, or a wrapper around DeckGLMap component, and provide exactly the same interface as the DeckGLMap component.
Now it does both, in a way merging the DeckGLMap component and the DeckGLMapDemo component.

You only need a very small class that holds the deckglSpec state and provides setSpecPatch. Something like this:
```js
function MapWrapper(props) {
    let [mapSpec, setMapSpec] = React.useState(null);

    React.useEffect(() => {
        setMapSpec(props.deckglSpec);
    }, [props.deckglSpec]);

    return (
        <Map
            {...props}
            deckglSpec={mapSpec}
            setSpecPatch={(patch) => {
                setMapSpec(_setPatch(mapSpec, patch));
            }}
        />
    );
}
```
We can take this discussion to Teams if you prefer.

react/package.json
alexandruandrei91: It's ok, we've updated react with the condition that if anything breaks we'll just downgrade again. Now that something broke, it's alright to get it back to 16.

react/src/lib/components/DeckGLMap/storybook/components/Map.stories.jsx
```diff
@@ -0,0 +1,79 @@
...
+            base,
+            normalizedPatch,
+            true,
+            false
+        ).newDocument;
+    } catch (error) {
+        console.error("Unable to apply patch: " + error);
+    }
+    return result;
+};
```
alexandruandrei91, line 56: Not for this PR, but in a future refactoring it would be good to extract these functions into some common place so you don't have to repeat them in multiple files.

alexandruandrei91 (approved):

shadab-skhan:
Map component storybook.

PR #498, opened 2021-08-16
Add all the needed files in the npm package

alexandruandrei91:


hkfb (approved):

anders-kiaer:
@alexandruandrei91 Should I do an alpha release now? https://github.com/equinor/webviz-subsurface-components/blob/3493151822fb476123758e96b3bd1cb703cb07de/.github/workflows/webviz-subsurface-components.yml#L91-L102 and https://github.com/equinor/webviz-subsurface-components/blob/3493151822fb476123758e96b3bd1cb703cb07de/.github/workflows/webviz-subsurface-components.yml#L55-L57 can be as is?

alexandruandrei91:
@anders-kiaer Yes, the commands should remain the same. It would be good to try an alpha release, to see how it goes :). I've tried locally with `npm pack` after setting the version and it seems to work ok.

PR #510, opened 2021-08-19
Fix Well Log size issues

alexandruandrei91:
The well log view didn't vertically scale to fill the parent. Also, horizontal scrolling was broken on chrome.


HansKallekleiv (approved):
Lgtm 👍

PR #516, opened 2021-08-25
Scale bar feature

shruthirai:


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -108,12 +111,14 @@ const Map: React.FC<MapProps> = ({
...
                     ref={refCb}
                     onHover={onHover}
+                    onViewStateChange={({viewState}) => setviewState({viewState})}
                 >
                     {children}
                 </DeckGL>
                 {showInfoCard ? <InfoCard pickInfos={hoverInfo} /> : null}
                 <Settings />
-            </ReduxProvider>
+                {viewState ? <DistanceScale zoomLevel={viewState} /> : null}
```
alexandruandrei91, line 120: It's better to send just what you need to a comp. You don't need the whole view state in DistanceScale, so here you can send just `viewState.zoom`.

alexandruandrei91 (commented):
You can run locally `npm run validate` to make sure you don't have eslint or tsc errors or warnings.

PR #520, opened 2021-08-30
Scale bar feature

shruthirai:


react/src/lib/components/DeckGLMap/DeckGLMap.jsx
```diff
@@ -193,6 +200,24 @@ DeckGLMap.propTypes = {
...
+        /**
+         * Toggle component visibility.
+         */
+        visible: PropTypes.bool,
+        /**
+         * Increment value for the scale.
+         */
+        incrementValue: PropTypes.number,
+        /**
+         * width per unit value to calculate width in units.
```
alexandruandrei91, line 216: It would be good to specify that this value will be in pixels, so maybe something like:
"Scale bar width in pixels per unit value"

react/src/lib/components/DeckGLMap/components/DistanceScale.tsx
```diff
@@ -1,26 +1,52 @@
 import React from "react";
 
 interface scaleProps {
+    // Needed the zoom value to calculate width in units
     zoom: number;
+    // Scale increment value
+    incrementValue: number;
+    // width per unit value to calculate width in units
```
alexandruandrei91, line 8: Same as above.

react/src/lib/components/DeckGLMap/components/DistanceScale.tsx
```diff
@@ -1,26 +1,52 @@
...
+    widthPerUnit: number;
 }
-const DistanceScale: React.FC<scaleProps> = ({ zoom }: scaleProps) => {
-    const [width] = React.useState(100);
-    const pixelPerUnit = width / Math.pow(2, zoom);
-    const scaleBarStyle = {
-        width: width,
+
+const roundOffvalues = function (num: number, step: number) {
```
alexandruandrei91, line 12: This name leads me to think that this function receives many values (maybe an array) and rounds them.
I think it should be called something like `roundToStep` since it tries to round a value up or down to a specified step.

react/src/lib/components/DeckGLMap/components/DistanceScale.tsx
```diff
@@ -1,26 +1,52 @@
...
+    return Math.floor(num / step + 0.5) * step;
+};
+
+const DistanceScale: React.FC<scaleProps> = ({
+    zoom,
+    incrementValue,
+    widthPerUnit,
+}: scaleProps) => {
+    const [rulerWidth, setRulerWidth] = React.useState<number>(widthPerUnit);
```
alexandruandrei91, line 21: This should probably be initialized to 0, since it will be rewritten immediately anyway.

react/src/lib/components/DeckGLMap/components/DistanceScale.tsx
```diff
@@ -1,26 +1,52 @@
...
         right: 0,
         bottom: 0,
     };
 
+    let scaleValue = 0;
+
+    scaleValue =
+        widthInUnits < incrementValue
+            ? Math.round(widthInUnits)
+            : roundOffvalues(widthInUnits, incrementValue);
```
alexandruandrei91, line 40: `scaleValue` can be `const`. Also it's redundant to initialize it to 0 and then immediately set it to another value. You can do it in one step:
```js
     const scaleValue = widthInUnits < incrementValue
            ? Math.round(widthInUnits)
            : roundOffvalues(widthInUnits, incrementValue);
```

PR #546, opened 2021-09-03
Ensure *.scss files are copied into dist directory.

### LINKED ISSUE #545: Missing .scss file in npm package
src/lib/components/WellLogViewer/components/styles.scss is not copied

hkfb:
Fixes #545.


anders-kiaer:
Haha 😄 Hmm... `\"src/lib/**/*\"` would that work or do we then include files we don't want to be copied (if so maybe we can use the [`exclude` input argument](https://www.npmjs.com/package/copyfiles))?

anders-kiaer (approved):
👍 LGTM

PR #549, opened 2021-09-06
TVD readout for wells

shadab-skhan:


hkfb (commented):
Looks like the TVD values is not interpolated like the MD value. Should it?

hkfb (approved):
Nice :)

PR #559, opened 2021-09-14
Fixed DeckGLMap python example and added test

### LINKED ISSUE #556: `DeckGLMap` python example does not work
The Deckgl python example in `examples/example_deckgl_map.py` does not render the map.
The Dash app successfully starts with no errors in shell or web console, but only an empty page is shown.

HansKallekleiv:
- [x] Added missing props to `DeckGLMap` python example, did some minor changes to layout and simplified imports
- [x] Added basic test for console errors

Closes #556


hkfb (approved):
Works fine now :+1:

anders-kiaer (approved):
Good job! :rocket: So the underlying issue was missing input data to the component? Should the component had given some console log errors in this case? And if so do we have an issue/task on getting some error from the component in case mandatory input data is missing?

hkfb:
> Good job! So the underlying issue was missing input data to the component? Should the component had given some console log errors in this case? And if so do we have an issue/task on getting some error from the component in case mandatory input data is missing?

Created issue #562

setup.py
```diff
@@ -25,7 +25,7 @@
     .replace("@", "")
 )
 
-INSTALL_REQUIRES = ["dash>=1.6"]
+INSTALL_REQUIRES = ["dash>=1.6", "jsonpatch>=1.32"]
```
anders-kiaer, line 28: Now I see what you mean @HansKallekleiv

anders-kiaer (changes requested):

anders-kiaer (approved):

PR #569, opened 2021-09-20
Redesign toolbox. Closes #423

### LINKED ISSUE #423: Redesign toolbox
Be able to specify per layer settings

shadab-skhan:


react/src/demo/example-data/deckgl-map.json
```diff
@@ -32,6 +32,7 @@
         {
           "@@type": "ColormapLayer",
           "id": "colormap-layer",
+          "name": "Surface map",
```
hkfb, line 35: Suggest "Property map" as default name for this layer.

Suggest also having default names for each layer, which will be used if not specified explicitly.

hkfb (changes requested):

react/src/demo/example-data/deckgl-map.json
shadab-skhan: Resolved

hkfb (approved):

PR #584, opened 2021-09-30
Discrete Color Legend For Well layers

shruthirai:


hkfb:
@shruthirai @shadab-skhan should we have a storybook example for this component also? We could handle that in a separate issue.

react/src/lib/components/DeckGLMap/components/DiscreteLegend.tsx
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+import legendUtil from "../utils/legend";
+import * as d3 from "d3";
```
shadab-skhan, line 3: Import only those modules from d3 which are being used.

react/src/lib/components/DeckGLMap/components/DiscreteLegend.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+interface ItemColor {
+    color: string;
+}
+
+interface colorLegendProps {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    discreteData: any;
```
shadab-skhan, line 11: Define the interface it cannot be set to any.

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -109,6 +110,30 @@ const Map: React.FC<MapProps> = ({
         setSpecObj(jsonConverter.convert(deckglSpec));
     }, [deckglSpec, resources]);
 
+    const [discreteData, setDiscreteData] = React.useState([]);
+    const [dataPresent, setDataPresent] = React.useState(false);
+
+    React.useEffect(() => {
+        if (deckglSpec) {
+            //eslint-disable-next-line
+            const logData = (deckglSpec["layers"] as any[])[2].logData;
```
shadab-skhan, line 119: use it as so there will not be multiple type casting 
const wells_layer = (deckglSpec["layers"] as any[])[2]);
const log_data = wells_layer.logData;
const log_data = wells_layer.logName;

also avoid using explicit indexing, use layer id to find the layer. From python dash side the layer ordering maybe different.

shadab-skhan (changes requested):

shadab-skhan:
> @shruthirai @shadab-skhan should we have a storybook example for this component also? We could handle that in a separate issue.
> 
> Another improvement could be to add outlines or shadows to the color boxes and make the text more readable on a dark background.
> 
> Also, most importantly, the color legend should be possible to hide, similar to the scale bar, eg. "visible = false".

We should have storybook for this component, I agree, it would be better to do in separate review.

Issue #586, opened 2021-10-01
Improve DeckGlMap interface - split Deck GL spec


hkfb:
Using the DeckGlMap component requires creating a complete DeckGL specification as a parameter. Suggested improvement is to split the DeckGl spec into smaller props that can be passed as arguments to the DeckGlMap component.

See also related issues #577 and  #562

Issue #577, opened 2021-09-26
Click event gives invalid patch


HansKallekleiv:
Selecting (mouse click) different well elements in the `DeckGLMap` component creates an invalid patch operation to a non-existing path.

Reproducable in the storybook:  https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/deckglmap--default

by:
1. Deactivate drawing edit mode
2. Highlight and click a few times on different wells
3. Observe the error in the console
 
Error:
```
       Unable to apply patch: OPERATION_PATH_UNRESOLVABLE: Cannot perform the operation at a path that does not exist
name: OPERATION_PATH_UNRESOLVABLE
index: 0
operation: {
  "op": "remove",
  "path": "/layers/2/selectedFeature/properties/md/0/2387"
}
tree: {
  "initialViewState": {
    "target": [
      434962.5,
      6478095.5,
      0
    ],
    "zoom": -3
  },
  "layers": [
    {
      "@@type": "ColormapLayer",
      "id": "colormap-layer",
      "bounds": [
        432205,
        6475078,
        437720,
        6481113
      ],
      "image": "@@#resources.propertyMap",
      "colormap": "https://cdn.jsdelivr.net/gh/kylebarron/deck.gl-raster/assets/colormaps/plasma.png",
      "valueRange": [
        2782,
        3513
      ],
      "pickable": true
    },
    {
      "@@type": "Hillshading2DLayer",
      "id": "hillshading-layer",
      "bounds": [
        432205,
        6475078,
        437720,
        6481113
      ],
      "valueRange": [
        2782,
        3513
      ],
      "opacity": 1,
      "image": "@@#resources.propertyMap",
      "pickable": true
    },
    {
      "@@type": "WellsLayer",
      "id": "wells-layer",
      "data": "https://raw.githubusercontent.com/equinor/webviz-subsurface-components/master/react/src/demo/example-data/volve_wells.json",
      "logData": "https://raw.githubusercontent.com/equinor/webviz-subsurface-components/master/react/src/demo/example-data/volve_logs.json",
      "opacity": 1,
      "lineWidthScale": 5,
      "logRadius": 6,
      "logrunName": "BLOCKING",
      "logName": "ZONELOG",
      "pointRadiusScale": 8,
      "outline": true,
      "logCurves": true,
      "refine": true,
      "selectionEnabled": true,
      "selectedFeature": {
        "type": "Feature",
        "geometry": {
[...]

PR #593, opened 2021-10-07
Multiple synced views in DeckGL map

shadab-skhan:
1. Removed views from deckGlSpecBase
2. Added syncedView prop to toggle between single and dual view
3. Updated example_deckgl_map.py to incorporate changes

### DIFF
--- a/examples/example_deckgl_map.py
+++ b/examples/example_deckgl_map.py
@@ -170,6 +170,7 @@ if __name__ == "__main__":
         coords={"visible": True, "multiPicking": True, "pickDepth": 10},
         scale={"visible": True},
         coordinateUnit="m",
+        views={"count": 1, "layout": [1, 1]},
         deckglSpecBase={
             "initialViewState": {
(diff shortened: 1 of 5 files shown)


HansKallekleiv:
There are use cases for having more than two views.
E.g. displaying two maps and a difference map between them. Demo using Leaflet here:
https://webviz-subsurface-example.azurewebsites.net/surface-viewer-fmu

Displaying multiple model realizations of the same surface for QC and to investigate how different model parameters affect the result:
Example:
image

Lets discuss it in the meeting tomorrow.

shadab-skhan:
@HansKallekleiv @hkfb 
I have created another PR to fix original issue #586. This change incorporated 2 things 
1. Improving interface for DeckGLSpec by separating views from the spec object
2. Display multiple views for for synced view, for which a different issue already exists #577 

Hence, we decided to create 2 different PRs for delivering fixes for both the issue #586 and #577.
Also, currently for synced views, it only shows multiple views, but feature is incomplete.

shadab-skhan:
New PR in progress

PR #598, opened 2021-10-13
Removed initialViewState from deckglSpec object

shadab-skhan:
1. Removed initialViewState from deckglSpec object
2. Added two properties bounds and zoom, where bounds is an array coordinates defines as
   [left, bottom, right, top] and default zoom is set to -3
3. Updated examples/example_deckgl_map.py


shadab-skhan:
Redundant

PR #601, opened 2021-10-14
Continuous color legend

shruthirai:


hkfb:
Also, is it possible to better contrast the text to the background? Eg. reverse of the background color or add a shadow?

PR #621, opened 2021-10-26
Performance testing demo

avanti-em:
There are a lot of commits since I was experimenting a lot of stuff with github workflow.


.github/workflows/webviz-subsurface-components.yml
```diff
@@ -140,3 +140,69 @@ jobs:
...
+        run: |
+          npm run validate --prefix ./react
+          black --check webviz_subsurface_components/ tests/ setup.py examples/
+          pylint webviz_subsurface_components/ tests/ setup.py examples/*.py --ignore well-completions-csv2json.py
+          bandit -r -c ./bandit.yml webviz_subsurface_components/ tests/ setup.py examples/
+
+      - name: 🤖 Run performance tests
+        if: github.event_name != 'release' # Related to https://github.com/equinor/webviz-subsurface-components/issues/409
+        run: |
+          npm run test_perf --prefix ./react
```
anders-kiaer, line 208: You could use the `matrix` `include` functionality perhaps, e.g.:
https://github.com/equinor/ecl2df/blob/0e30fb8046bf17fd338bb468584985c5d816e2f6/.github/workflows/ecl2df.yml#L25-L32
https://github.com/equinor/ecl2df/blob/0e30fb8046bf17fd338bb468584985c5d816e2f6/.github/workflows/ecl2df.yml#L54

So instead of `install-ert`, we here would have `run-performance-tests`.

Then you don't need to define a complete new job `performance-check` repeating the steps in the existing job :slightly_smiling_face:

anders-kiaer (changes requested):
Looks good @avanti-em. One suggestion on the `.github/workflow`. Why has the `package-lock.json` changed with +500 -34000 lines?

.github/workflows/webviz-subsurface-components.yml
avanti-em: Oh! I didn't realize the package-lock.json got so heavily modified, though it shouldn't be; I just added one package "@actions/core". I will take a look.
Also , working on adding matrix and 'include' approach in the workflow

.github/workflows/webviz-subsurface-components.yml
```diff
@@ -19,7 +19,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9']
+        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        include:
+          - run-performance-tests: true
```
anders-kiaer, line 24: It looks like this ends up with the following matrix:
`["3.6", true], ["3.7", true], ["3.8", true], ["3.9", true]`

While we want probably only one to be `true`. I can help research the GitHub workflow `matrix` syntax documentation a bit further.

anders-kiaer: One way that should work would be to e.g.:
```
matrix:
  include:
    - python-version: 3.6
       run-performance-tests: false
    - python-version: 3.7
       run-performance-tests: false
    - python-version: 3.8
       run-performance-tests: false
    - python-version: 3.9
       run-performance-tests: false
    - python-version: 3.9
       run-performance-tests: false
```
The only downside would be that it is a bit more verbose, but maybe the best solution? See also https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#using-environment-variables-in-a-matrix. What do you think @avanti-em?

anders-kiaer (changes requested):

.github/workflows/webviz-subsurface-components.yml
```diff
@@ -19,7 +19,16 @@ jobs:
...
+        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        include:
+          - python-version: 3.6
+            run-performance-tests: false
+          - python-version: 3.7
+            run-performance-tests: false
+          - python-version: 3.8
+            run-performance-tests: false
+          - python-version: 3.9
+            run-performance-tests: true
```
anders-kiaer, line 31:
```suggestion
          - python-version: 3.9
            run-performance-tests: false
          - python-version: 3.9  # This job is allowed to fail
            run-performance-tests: true
```
We want the standard Python 3.9 job to be required, but the performance test run to allow to fail (when that is true). One hacky solution from my side is to do 3.9 twice. We could look into later if there are better ways of defining the job matrix GitHub action workflow, but for now I think this is good enough.

anders-kiaer (commented):
Good work @avanti-em! 👍 One last comment from my side, otherwise LGTM.

.github/workflows/webviz-subsurface-components.yml
avanti-em: Oh! I am sorry!... I completely missed it . You had already mentioned that in the previous comment to this but I missed it. I have added another commit

anders-kiaer (commented):
We can remove
```
python-version: ["3.6", "3.7", "3.8", "3.9"]
```
I _think_ (since we mention it in `include` explicitly).

avanti-em:
> We can remove
> 
> ```
> python-version: ["3.6", "3.7", "3.8", "3.9"]
> ```
> 
> now I _think_ (since we mention it in `include` explicitly).

hi @anders-kiaer , I removed the "python-version: ["3.6", "3.7", "3.8", "3.9"]". But the original checks (webviz-subsurface-components (3.x) ) has got skipped and the new ones defined in this particular commit only, has run successfully. Is this expected?

anders-kiaer:
Indeed - that is expected :)

Should we then merge this from your point of view? 

I see GitHub has a new feature in public beta starting some weeks ago which enables "workflow reuse" (https://github.blog/2021-11-19-whats-new-from-github-changelog-october-2021-recap/#github-actions). Might be something we as a project could consider in the future.

avanti-em:
> Indeed - that is expected :)
> 
> Should we then merge this from your point of view?
> 
> I see GitHub has a new feature in public beta starting some weeks ago which enables "workflow reuse" (https://github.blog/2021-11-19-whats-new-from-github-changelog-october-2021-recap/#github-actions). Might be something we as a project could consider in the future.

Yes , this looks promising , we should definitely consider using this version of workflow re-use in future. 
For now , it looks good to me to be merged. I did some other small changes as well in the latest commit apart from what you suggested. Please take a look at them and let me know if it looks okay.

anders-kiaer:
> I did some other small changes as well in the latest commit apart from what you suggested. Please take a look at them and let me know if it looks okay.

LGTM 👍🙂 Thanks.

anders-kiaer (approved):

PR #634, opened 2021-11-05
Grid rotate

nilscb:


react/src/lib/components/DeckGLMap/layers/utils/layerTools.ts
```diff
@@ -58,3 +59,19 @@ export function patchLayerProps<
...
+export function getModelMatrix(deg: number, x: number, y: number): Matrix4 {
+    const rad = deg * 0.017453;
+    const IDENTITY = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1];
+
+    const m1 = new Matrix4(IDENTITY).translate([-x, -y, 0, 1]); // translate to origin
+    const mRot = new Matrix4(IDENTITY).rotateZ(rad); // rotate
+    const m2 = new Matrix4(IDENTITY).translate([x, y, 0, 1]); // translate back
+
+    // Make  m2*mRot*m1
+    mRot.multiplyRight(m1);
```
shadab-skhan, line 73: Not sure about the API, looks the result is not used. Is the multiplication result in-place?

nilscb: There is a call to getModelMatrix in colormapLayer.ts and also the hillshading layer. Thaay are in the "draw" function.

shadab-skhan (approved):

PR #644, opened 2021-11-12
Discrete legend using color table

shruthirai:


react/src/demo/example-data/color-tables.json
```diff
@@ -62,5 +62,119 @@
       [ 0.95, 0.45, 1.0, 0.0 ],
       [ 1.00, 1.0, 0.0, 0.0 ]
     ]
+  },
+  {
+    "name": "ZonelogData",
```
shadab-skhan, line 67: Color table name cannot be data specific, we need to make it generic.

shruthirai: I did not get any proper json colors link for discrete data, so for now I have renamed the name as color_set_1, color_set_2 and so on.

shadab-skhan (approved):

PR #645, opened 2021-11-12
Continuous legend using color table

shruthirai:


react/src/lib/components/DeckGLMap/components/ContinuousLegend.tsx
```diff
@@ -1,106 +1,96 @@
...
 
-            const legendaxis = axisRight(legendscale)
-                .scale(legendscale)
-                .tickValues(legendscale.domain());
-            const svg = select(selected_id)
-                .append("svg")
-                .attr("height", legendheight - 3 + "px")
-                .attr("width", legendwidth - 20 + "px");
+        //create tick marks
+        const xLeg = scaleLinear().domain([min, max]).range([10, 258]);
```
shadab-skhan, line 83: Why are we using hard coded range?

react/src/lib/components/DeckGLMap/utils/continuousLegend.ts
```diff
@@ -1,5 +1,86 @@
...
+interface colorTemplate {
+    name: string;
+    properties: propertiesArr;
+}
+
+export function colorsArray(
+    objectName: string
+): [number, number, number, number][] {
+    const properties = colorTemplate[0]["properties"];
+    const propertiesData = properties.filter(
```
shadab-skhan, line 32: we can use type insensitive comparison by using toLower

react/src/lib/components/DeckGLMap/utils/continuousLegend.ts
```diff
@@ -1,5 +1,86 @@
...
+}
+
+export function rgbValues(
+    objectName: string,
+    point: number
+): number[] | { r: number; g: number; b: number; opacity: number } | undefined {
+    const color_table = colorsArray(objectName);
+    const colorArrays = color_table.find(
+        (value: [number, number, number, number]) => {
+            return point.toFixed(2) == value[0].toFixed(2);
```
shadab-skhan, line 48: Comparing with 2 decimal points may not work for some log/property data.

react/src/lib/components/DeckGLMap/components/ContinuousLegend.tsx
shruthirai: Here range is used to adjust the position of tick line according to legend, which will always be same, so I feel that's fine to give hard coded range.

react/src/lib/components/DeckGLMap/utils/continuousLegend.ts
shruthirai: Changes as per the comment

react/src/lib/components/DeckGLMap/utils/continuousLegend.ts
shruthirai: Removed toFixed(2) because as per the logic however, it checks if point and value matches else it will do interpolation.

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -21,9 +23,14 @@ const Template = (args) => {
...
+            colorTables={colorTables}
         />
     );
 };
 
 export const Default = Template.bind({});
 Default.args = exampleData[0];
+
+export const templateData = Template.bind({});
+templateData.args = { template: colorTemplate, colorTables: colorTables };
```
shadab-skhan, line 36: this templateData won't work on storybook as it will need other props from exampleData too

shadab-skhan (commented):
Please confirm the storybook issue, rest of the changes are fine

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
shruthirai: I can make necessary changes when I submit separate PR for storybook . If there is no other issues, then may be you can accept this PR.

shadab-skhan (approved):

PR #653, opened 2021-11-18
Well log viewer tests

avanti-em:
The workflow run will fail as one of the tests is failing


Vladimir-Kokin:
```
      38 |         expect(mockFn).toHaveBeenCalledWith("tvd");
    > 39 |         expect(axis_selectors[1]).toBeChecked();
         |                                   ^
      40 |     });
```
It seems like a normal behaviour. This Radio only send notification but do not automatically becomes checked. The check state of the Radio buttons could be changed by changing prop "value" of the component.

Vladimir-Kokin (approved):

PR #654, opened 2021-11-18
Removed unwanted storybook code

shruthirai:


hkfb:
What is the purpose? Could you give a quick description?

shruthirai:
> What is the purpose? Could you give a quick description?

@hkfb This piece of code was written at start but since we already have separate Storybook for discrete and continuous legend now, I think this is not required now.

hkfb (approved):

PR #655, opened 2021-11-19
Well Log Viewer: Select current track; edit Plot with Dialog

Vladimir-Kokin:
- select current track
- edit plot options with Dialog
- dblclk handler


react/package.json
```diff
@@ -27,6 +27,7 @@
         "build": "npm run transpile && npm run build:js && npm run build:js-dev && npm run tidy-up && npm run build:py && npm run tidy-up",
         "typecheck": "tsc --noEmit",
         "format": "eslint --fix *.js *json \"src/**/*.+(ts|tsx|js|jsx|json|css)\"",
+        "formatWellLogViewer": "eslint --fix \"src/lib/components/WellLogViewer/**/*.+(ts|tsx|js|jsx|json|css)\"   ",
```
hkfb, line 30: Maybe you could create your own local script instead.

hkfb (changes requested):

react/package.json
Vladimir-Kokin: The only way to do that I know is insert it into the package.json

Vladimir-Kokin: I have restored package.json

hkfb (approved):

PR #658, opened 2021-11-22
Reversed order of layer toggles

### LINKED ISSUE #620: Reverse order of layer toggles in toolbox
The ordering of the layer toggles is unintuitive. The top-most layer in the map should be the top-most toggle.

hkfb:
Represents the deck.gl layer stack more intuitively.

Fixes #620


shadab-skhan (commented):
This works, but still the layer ordering is governed by the order in which it's defined in deckgl-map.json file. Do we always want layers to be in particular fixed order? If that being the case, we need to do some more changes.
Or for storybook we can change the ordering in deckgl-map.json file instead of dong it here.

hkfb:
> This works, but still the layer ordering is governed by the order in which it's defined in deckgl-map.json file. Do we always want layers to be in particular fixed order? If that being the case, we need to do some more changes. Or for storybook we can change the ordering in deckgl-map.json file instead of dong it here.

The purpose is only to change to a more intuitive representation in the layer selector. This does not and should not affect representation when drawing layers.

shadab-skhan (approved):

PR #661, opened 2021-11-22
Passing colortable and template as prop

shruthirai:


react/src/lib/components/DeckGLMap/DeckGLMap.jsx
```diff
@@ -1,6 +1,10 @@
 import PropTypes from "prop-types";
 import * as React from "react";
 import Map from "./components/Map";
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
hkfb, line 4: I think it should be possible to fix this by using import statement

hkfb (changes requested):

react/src/lib/components/DeckGLMap/DeckGLMap.jsx
shruthirai: Cannot use import statement in jsx file

hkfb (approved):

Issue #660, opened 2021-11-22
Polygon fills in DeckGLMap


jin-queen:
Is it possible to adjust the color opacity, the color intensitivity and/or choose a color for the infills in polygons?
This based on i.e. numeric values.

nilscb:
Which polygons would this be. Fault polygons? We may add properties for infill color and opacity here if requested.

jin-queen:
We use Leafletmap component as of today and here  there is a general polygon layer, since we plot closed polygons for other attributes than faults. Since Leafletmap is not updated any more, we will move to DeckGLMap and will need these properties. 
 I guess that the fault polygon layer might be used for other kinds of polygons as well?
 If not, than it would be good to have a general polygon layer.

hkfb:
Should be possible to add a polygon layer directly to the "layers" prop in the DeckGlMap. For example, you could add something like

```
"layers": [
    {
        "type": "GeoJsonLayer",
        "data": <url to geojson file>,
        "opacity": 0.5
    },
    ...
]
```

See https://deck.gl/docs/api-reference/layers/geojson-layer

We could try to make a Storybook example with an example polygon.

hkfb:
@jin-queen Please see this Storybook example: https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/deckglmap--custom-layer

Does this cover your use case?

PR #673, opened 2021-11-29
Storybook example to additional deckgl layer 

shadab-skhan:
Any standard deckgl layer can be added as a layer to map component.


hkfb (approved):

hkfb:
Should this be linked to #660?

hkfb:
Fixes #660

Issue #656, opened 2021-11-22
Grouptree Storybook examples not working


hkfb:
See Grouptree DateTimeSlider, FlowRateSelector, etc

[Eg.](https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/grouptree-components-settings-flowrateselector--default)

PR #679, opened 2021-11-29
Fix storybook for grouptree

nilscb:


hkfb (changes requested):
Should this be linked to #656?

hkfb (approved):

PR #681, opened 2021-11-29
Horizontal and vertical layout options for color legends

shruthirai:


hkfb (commented):
Maybe not related to this change, but the legend doesn't update when changing the "logName" in the wells layer

shruthirai:
> Maybe not related to this change, but the legend doesn't update when changing the "logName" in the wells layer

Yes, this issue is already discussed with Shadab earlier and he mentioned that its because of some store update issue, which will be looked by him.

hkfb (approved):

PR #683, opened 2021-12-02
WellLogViewer: Content area selection, add/edit track dialog

Vladimir-Kokin:
- Add/Edit Track Dialog
- TrackEvent for selection and track editting
- Content selection
- Set/get visible interval with props
- Set/get selected interval with props
- Storybook descriptions of props
- Color table color channel range [0,255] instead of [0.0,1.0]
- Some fixes


hkfb (approved):
Some observations:

* The selection range highlighting extends into header area:

image

* There are some debugging statements left on the console, which should be removed

* Toggling a plot also toggles the track selection

* Would it be possible to directly edit the plot which is clicked on rather than having to select it in the pop-up dialog?

If you like we can merge this change like it is and address the above in a follow up PR.

Vladimir-Kokin:
- some console output removed
- plot dialog could be directly called by double click in legend area. If it is the only plot in the track no additional context menu appear to select plot from the track

PR #687, opened 2021-12-07
Map performance test

avanti-em:
The test for map performance is terminating with error : "Cannot find module '!!raw-loader!./decoder.fs.glsl' from 'src/lib/components/DeckGLMap/layers/shader_modules/decoder.ts'"

May or may not be because of : 
[](https://github.com/equinor/webviz-subsurface-components/pull/627/files)

To run the test, please use: npm run test -- Map_performance.test.tsx


hkfb:
@avanti-em added a fix for the !!raw-loader! syntax. I believe the current failure is unrelated.

PR #690, opened 2021-12-08
Terrain layer

nilscb:


package.json
```diff
@@ -0,0 +1,8 @@
+{
```
hkfb, line 1: Is this file needed?

react/package.json
```diff
@@ -30,7 +30,7 @@
         "lint": "eslint *.js *.json \"src/**/*.+(ts|tsx|js|jsx|json|css)\"",
         "validate": "npm run typecheck && npm run lint",
         "setup-deckgl-types": "cd ./node_modules/@danmarshall/deckgl-typings/ && indefinitely-typed --folder deck.gl__aggregation-layers --folder deck.gl__core --folder deck.gl__extensions --folder deck.gl__geo-layers --folder deck.gl__google-maps --folder deck.gl__json --folder deck.gl__layers --folder deck.gl__mapbox --folder deck.gl__mesh-layers --folder deck.gl__react --folder deck.gl --folder luma.gl__constants --folder luma.gl__core --folder luma.gl__gltools --folder luma.gl__webgl-state-tracker --folder luma.gl__webgl --folder math.gl__core --folder math.gl",
-        "storybook": "start-storybook -p 6006",
+        "storybook": "start-storybook -p 6006 -s public",
```
hkfb, line 33: Doesn't look like you included the depth map file. Perhaps this should point to the 'example-data' directory.

hkfb (changes requested):

react/package.json
```diff
@@ -50,6 +50,9 @@
         "@equinor/eds-core-react": "^0.12.1",
         "@equinor/eds-icons": "^0.6.2",
         "@equinor/videx-wellog": "^0.4.5",
+        "@loaders.gl/core": "^2.3.13",
+        "@loaders.gl/images": "^2.3.13",
+        "@loaders.gl/terrain": "^3.0.13",
```
hkfb, line 55: @nilscb Are these needed explicitly? They are installed automatically on my system

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -0,0 +1,73 @@
+import { CompositeLayer } from "@deck.gl/core";
+import TerrainMapLayer, {
+    TerrainMapLayerProps,
+    TerrainMapLayerData,
+} from "./terrainMapLayer";
+import { ExtendedLayerProps } from "../utils/layerTools";
+import { layersDefaultProps } from "../layersDefaultProps";
+import { TerrainLoader } from "@loaders.gl/terrain";
```
hkfb, line 8: Got a typescript error from this line. Replacing with 
```
const terrain = require("@loaders.gl/terrain");
```
seems to work.

@shadab-skhan @shruthirai @Vladimir-Kokin any of you get an error?

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
nilscb: I think you have to install it.. npm install loaders.gl/terrain

hkfb (approved):

PR #707, opened 2021-12-20
Link property "view3d" to " layers "is3d"

nilscb:


shadab-skhan (approved):
We can get this in, we can optimize the store update on layer and view3D prop update.

PR #710, opened 2021-12-23
Multi view

shadab-skhan:
Improvised prop for multiple view.
views: {
  layout: [number, number],
 viewports: [
    {
         id: string,
         show3D: boolean,
         layerIds: string[]
    }
  ]
}

### DIFF
--- a/react/src/lib/components/DeckGLMap/components/Map.tsx
+++ b/react/src/lib/components/DeckGLMap/components/Map.tsx
@@ -3,7 +3,7 @@ import DeckGLWrapper, { ViewsType } from "./DeckGLWrapper";
 import React from "react";
 import { Provider as ReduxProvider } from "react-redux";
 import { createStore } from "../redux/store";
-import { setLayers } from "../redux/actions";
+import { setSpec } from "../redux/actions";
 import { getLayersWithDefaultProps } from "../layers/utils/layerTools";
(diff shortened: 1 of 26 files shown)


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -130,9 +126,11 @@ const Map: React.FC<MapProps> = ({
                 false,
                 false
             ).newDocument;
-            store.dispatch(setLayers(new_state));
+            const spec = { layers: new_state, views: views };
+            console.log(spec);
```
nilscb, line 130: Sould probably not be here?

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
```diff
@@ -88,7 +98,7 @@ export default class WellsLayer extends CompositeLayer<
             ? splineRefine(this.props.data as FeatureCollection) // smooth well paths.
             : (this.props.data as FeatureCollection);
 
-        const is3d = this.props.is3d;
+        const is3d = this.context.viewport.constructor.name === "OrbitViewport";
```
nilscb, line 101: This is nice!

nilscb (commented):
Good work.

nilscb (approved):

Issue #718, opened 2022-01-03
Error in scss file loader


hkfb:
Error loading *.scss files when the webviz-subsurface-components library is used as a 3rd party dependency

Similar issue was solved in DeckGLMap colormapLayer.ts using the `!!...-loader!` format

PR #719, opened 2022-01-03
Use inline loaders for .scss, .glsl and .css files.

### LINKED ISSUE #717: LayeredMap: error in glsl file loader
Error when the webviz-subsurface-components library is used as a 3rd party dependency:

```
ERROR in ./node_modules/@webviz/subsurface-components/dist/components/LayeredMap/shaders/vertexShader.vs.glsl 7:10
Module parse failed: Unexpected token (7:10)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
```

Similar issue was solved in DeckGLMap colormapLayer.ts using the `!!raw-loader!` format

hkfb:
Allows using react components as 3rd party libraries where consumers have not configured loaders for said files.

Fixes #717, #718


jorgenherje (approved):
Looks good to me!

Built and tested locally for `VectorCalcualtor`. Seems fine!

HansKallekleiv (approved):

PR #722, opened 2022-01-04
Exclude storybook and tests from generated package

t0oF:
In attempt to fix the dependency to addon-redux in the generated package, I started to exclude test and storybook related files.
Unfortunately, this does not prove to be enough, because the DeckGLMap, GroupTree and WellCompletions components still have an explicit dependency on addon-redux (see the files redux/store.ts in each of these components).


hkfb (approved):
Thanks Christophe! I made a small modification.

Added separate tsconfig for builds so that typecheck will still run on test files.

PR #735, opened 2022-01-07
Add color template and color table as deckgl context

shadab-skhan:
Add color template and color table as deckgl context instead of setting it as state


hkfb (approved):
Nice! didn't know about the LayerContext :)

PR #741, opened 2022-01-10
Use color table

nilscb:


hkfb (commented):
The colors are applied step-wise. Would be nice to be able to interpolate between color samples.

react/src/lib/components/DeckGLMap/layers/grid/gridLayer.ts
```diff
@@ -117,21 +162,46 @@ GridLayer.defaultProps = layersDefaultProps[
...
+    valueRange: [number, number],
+    colorMapRange: [number, number]
+): PolygonData[] {
     const polygons: PolygonData[] = data.map(function (
         cell: CellData
     ): PolygonData {
-        const v = cell.vs[ti];
+        const propertyValue = cell.vs[ti];
+
+        // console.log("colors: ", colors)
```
hkfb, line 177: Please clean up commented out code.

react/src/demo/example-data/deckgl-map.json
```diff
@@ -73,6 +74,21 @@
...
+        "colorMapName": "Rainbow",
+        "valueRange": [
+          0,
+          1
+        ],
+        "colorMapRange": [
+          0,
+          1
+        ],
+        "visible": true
```
hkfb, line 90: This leaves the grid layer hidden behind the property map layer in the default storybook example (with a small artifact protruding on the left side of the map). Would be better to create a separate storybook example for the grid layer.

hkfb (changes requested):

hkfb (approved):

PR #743, opened 2022-01-11
Fixed deckgl map to add and remove layers

### LINKED ISSUE #705: DeckGL layers cannot be added/removed
It is not possible to add or remove layers after the initial render.
This is an important use case, especially when multiple views are concerned.
Number of views, and therefore the number of layers may change dynamically based on user input.

### LINKED ISSUE #724: Removing Color and Hillshading layer when Wells layer is present throws error

shadab-skhan:
Closes #724 
Closes #705

### DIFF
--- a/react/src/demo/example-data/deckgl-map.json
+++ b/react/src/demo/example-data/deckgl-map.json
@@ -75,7 +75,6 @@
       },
       {
         "@@type": "Map3DLayer",
-        "id": "map3d-layer",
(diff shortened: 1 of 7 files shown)


hkfb (changes requested):
An empty toggle is left in the layer menu

Screenshot from 2022-01-11 09-22-32

hkfb (approved):

PR #746, opened 2022-01-11
Import color table repo

shruthirai:
Importing discrete and continuous legend from color table repo will go as another commit


react/src/lib/components/DeckGLMap/components/DeckGLWrapper.tsx
```diff
@@ -18,7 +18,7 @@ import { DrawingLayer, WellsLayer, PieChartLayer } from "../layers";
 import { Layer, View } from "deck.gl";
 import { DeckGLView } from "./DeckGLView";
 import { Viewport } from "@deck.gl/core";
-import { colorTablesArray } from "./ColorTableTypes";
```
hkfb, line 21: Can this file be removed now?

hkfb (changes requested):

react/src/lib/components/DeckGLMap/components/DeckGLWrapper.tsx
shruthirai: Yes, will test once and submit changes.

shruthirai: Since this file is used inside discrete and continuous legend currently, it is better to remove later once we use legend from color table repo.

hkfb (approved):

PR #749, opened 2022-01-12
Map3 d layer readout

nilscb:


react/src/lib/components/DeckGLMap/layers/terrain/terrainMapLayer.ts
```diff
@@ -65,6 +65,9 @@ export interface TerrainMapLayerProps<D> extends SimpleMeshLayerProps<D> {
 
     // Use color map in this range.
     colorMapRange: [number, number];
+
+    //If true readout will be z value (depth). Otherwise it is the texture property value.
+    isReadoutDepth: boolean;
 }
 
 const defaultProps = {
```
shadab-skhan, line 73: We should have created default  props in layersDefaultProps.ts
May be in another review.

shadab-skhan (approved):

PR #757, opened 2022-01-18
Added property map with transparent border and added a fix in fra…

nilscb:
…gment shader to discard transparent pixels.


react/src/lib/components/DeckGLMap/layers/terrain/terrainmap.fs.glsl
```diff
@@ -48,6 +48,12 @@ void main(void) {
 
    vec4 color = hasTexture ? texture(sampler, vTexCoord) : vColor;
 
+   // Discard transparent pixels.
+   if (!picking_uActive && color.w < 1.0) {
```
hkfb, line 52: Getting lots of these errors: 

> [.WebGL-0x55a5bbe1f9c0]RENDER WARNING: there is no texture bound to the unit 0

hkfb (changes requested):

hkfb (approved):

PR #776, opened 2022-01-24
Well logs viewer tests

avanti-em:


Vladimir-Kokin (approved):
It seems to be OK

Vladimir-Kokin:
```
FAIL src/lib/components/WellLogViewer/WellLogViewer.test.tsx
  ● Test Well Log Viewer Component › snapshot test

    TypeError: name.toLowerCase is not a function

      366 |     logController.scaleHandler = scaleHandler;
      367 |     logController.domain = minmaxPrimaryAxis;
    > 368 |     logController.setTracks(tracks);
          |                   ^
      369 | }
      370 |
      371 | function addTrackMouseEventListner(
      
```
      
This error is due to a BUG in videx-wellog  (file log-controller.ts, function _trackEnter):
```
    if (showLegend) {
      newtracks.append('div')
        .classed('track-legend', true)
        .classed('hidden', legendHeight <= 0)
        .style([attr.size], `${legendHeight}px`);

```

The last line should reads (without square brackets):
```
.style(attr.size, `${legendHeight}px`);
```

The same bug is in function _trackUpdate
```
    if (showLegend) {
      const legend = selection.select('.track-legend');
      legend
        .style([sizeAttr], `${legendHeight}px`)

```

It is unclear why these bugs leads to crash only in test environment

To avoid the crash it is possible to turn off legends in file WellLogView.tsx (function createLogViewer)
```
            // create new LogViewer
            this.logController = new LogViewer({
                showLegend: false,
```

But after that there are still some another errors concerning the test

Vladimir-Kokin:
Make an issue about he bug https://github.com/equinor/webviz-subsurface-components/issues/803
and make a nwe options to allow hige legend via props #803

PR #780, opened 2022-01-27
Fixed wells TVD readout in 2d view

### LINKED ISSUE #775: TVD readout for wells is broken

shadab-skhan:
Closes #775


hkfb (commented):
Maybe not related to this change, but it looks like TVD value is inverse?

nilscb (approved):

shadab-skhan:
Yes, the TVD values are inverted in data itself. We should take a call on fixing that,

PR #787, opened 2022-01-28
Float32 support for Map3D property texture.

nilscb:


hkfb (changes requested):
Can we create a new layer for this format? Would be good to compare performance characteristics.

Also, it would be good to optionally include a value range so that it is not necessary to calculate min and max from the input data.

react/src/lib/components/DeckGLMap/layers/map/mapLayer.ts
```diff
@@ -0,0 +1,351 @@
...
+                        );
+                        resolve(imageData);
+                    };
+                });
+            });
+
+        // Same as textur above. However this will not have colormap applied
+        // to it an will contain the property vaules directly (in Float32)
+        // May be more efficient methods to do this.
+        const propertyValuesImageData = fetch(this.props.propertyTexture)
```
hkfb, line 299: Should not be necessary to fetch the same data twice.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/map/mapLayer.ts
```diff
@@ -0,0 +1,351 @@
...
+    // Will calculate normals and enable phong shading.
+    enableSmoothShading: boolean;
+}
+
+export default class MapLayer extends CompositeLayer<
+    unknown,
+    MapLayerProps<unknown>
+> {
+    renderLayers(): [PrivateMeshLayer] {
+        let mesh = load(this.props.mesh, TerrainLoader, {
```
hkfb, line 247: The mesh needs the same encoding as the texture. The reason is that we don't know in advance which map is depth and which is another property.

hkfb (changes requested):

PR #789, opened 2022-01-31
Fixed broken pylint checks

shadab-skhan:


hkfb (approved):
Tested the Morris example, and it seems to work.

PR #800, opened 2022-02-03
Labels for views

### LINKED ISSUE #773: Labels for views
Optionally show label on each view

shadab-skhan:
"views": {
      "showLabel": false,
      "viewports": [
        {
          "id": "view_1",
          "name"?: "Main view",
          "show3D"?: false,
          "layerIds":? []
        }
      ]
    }

Default label is generated if not supplied. Label visibility is controlled by a global flag "showLabel".
Closes #773

### DIFF
--- a/react/src/demo/example-data/deckgl-map.json
+++ b/react/src/demo/example-data/deckgl-map.json
@@ -148,6 +148,7 @@
     },
     "views": {
       "layout": [1, 1],
+      "showLabel": true,
       "viewports": [
         {
           "id": "view_1",
--- a/react/src/demo/example-data/deckgl-map.json
+++ b/react/src/demo/example-data/deckgl-map.json
@@ -148,7 +148,7 @@
     },
     "views": {
       "layout": [1, 1],
-      "showLabel": true,
+      "showLabel": false,
       "viewports": [
         {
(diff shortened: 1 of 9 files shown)


hkfb (changes requested):
The label should be optional, in case it does not match the style of the consumer.

shadab-skhan:
> The label should be optional, in case it does not match the style of the consumer.

The label is optional, if it's not specified, the component will generate default labels like "View 1", "View 2", ...
Did you meant something else?

hkfb:
> > The label should be optional, in case it does not match the style of the consumer.
> 
> The label is optional, if it's not specified, the component will generate default labels like "View 1", "View 2", ... Did you meant something else?

Yes, I meant by default there should be no label at all. At least this is my opinion.

shadab-skhan:
> > > The label should be optional, in case it does not match the style of the consumer.
> > 
> > 
> > The label is optional, if it's not specified, the component will generate default labels like "View 1", "View 2", ... Did you meant something else?
> 
> Yes, I meant by default there should be no label at all. At least this is my opinion.

Done! Now if name is supplied, it will be used to label the in viewport. Visibility of name is controlled by a global flag.

Issue #292, opened 2021-02-16
User documentation for the DeckGL map component


alexandruandrei91:
Document the DeckGL component interface: props, deck.gl json etc.

#251

shadab-skhan:
Creating new issue to improvise document in storybook

PR #809, opened 2022-02-09
Added a minimal map example.

hkfb:
Work in progress for #292.


shadab-skhan (approved):

shadab-skhan:
We should add another example, where only required props are passes and map and other components are visualized, for rest default props will be used by the component.

Issue #754, opened 2022-01-17
Property 3D map disappears when zooming in


hkfb:
Using the Default Storybook example, in orthogonal view, enabling only the Map 3D layer, the map disappears when zooming in beyond around 300 m on the scale bar.

PR #815, opened 2022-02-10
Fix: Property 3D map disappears when zooming in #754

nilscb:


hkfb (commented):
Works well in orthogonal view. I can zoom in to the 34 m mark. But in 3D view I am not able to zoom as far as before.

hkfb (approved):
LGTM

PR #819, opened 2022-02-15
Further minimizing kh example map

hkfb:


react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
```diff
@@ -15,6 +15,7 @@ export const layersDefaultProps: Record<string, unknown> = {
             step: 0,
         },
         rotDeg: 0,
+        colorMapName: "Rainbow",
```
shadab-skhan, line 18: default colorMapRange and valueRange are not defined properly in ColorMapLayer and Hillshading2DLayer.
Need to be { type: "array", value: [0, 1]}

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -242,10 +242,11 @@ KhMapFlat.args = {
     bounds: [432150, 6475800, 439400, 6481500],
     layers: [
         {
-            ...colormapLayer,
+            "@@type": "ColormapLayer",
+            id: "property_map",
```
shadab-skhan, line 246: If we are using single layer of any type, id can be optional too.

PR #821, opened 2022-02-15
Cypress tests

avanti-em:


.github/workflows/webviz-subsurface-components.yml
```diff
@@ -94,6 +95,16 @@ jobs:
           npm run test_correctness --prefix ./react
           pytest ./tests --headless
 
+      # Run cypress tests on firefox browser
+      - name: E2E on firefox
+        if: github.event_name != 'release' && matrix.e2e-on-firefox
+        uses: cypress-io/github-action@v2
```
anderskiaer, line 101: Should we turn this into a SHA? I.e. instead of `@versionnumber` we use `@sha`

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

anders-kiaer, line 101: Should we turn this into a SHA? I.e. instead of `@versionnumber` we use `@sha`

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

avanti-em: Hi @anders-kiaer , according to the official documentation of Github:
[](https://github.com/actions/upload-artifact#retention-period)

Artifacts are retained for 90 days by default. We can specify a shorter retention period using the "retention-days" input.
for example we will keep files only for 5 days. So that it does not put stress on our storage capacity.
The retention period for a pull request restarts each time someone pushes a new commit to the pull request.

Right now, we store images only if a test fails. And videos are always recorded despite of pass or failure.
So we can stop recording videos to further save our storage

anders-kiaer: Sounds great @avanti-em  - thanks for the information. Then as you say, storage should not be a problem.

The new actions included in this PR are approved, it might be that the PR needs a new commit or something to restart the workflow/CI?

avanti-em: I have made the updates, and triggered the pipeline

anders-kiaer (approved):
LGTM when CI is 🟢👍

hkfb:
Blocked by #900

hkfb:
@avanti-em looks like the reference screenshots need to be updated?

avanti-em:
> @avanti-em looks like the reference screenshots need to be updated?

hi @hkfb , I have updated the screenshots. The tests are also passing.

PR #822, opened 2022-02-15
Use normalized data for both mesh and texture.

nilscb:


hkfb (approved):
I still see some artifacts around the edge of the map.

PR #838, opened 2022-02-19
Make color scale component available through Dash

shruthirai:
Throws error while building 
npm run build  -----> TypeError: Cannot convert undefined or null to object


hkfb:
@shruthirai made a fix for the continuous legend wrapper. Should be similar for discrete legend wrapper.

shruthirai:
> @shruthirai made a fix for the continuous legend wrapper. Should be similar for discrete legend wrapper.

@hkfb Thanks for the fix. Will do the same for discrete legend as well.

hkfb (approved):

PR #839, opened 2022-02-21
Float resolution update

nilscb:


react/src/demo/example-data/deckgl-map.json
```diff
@@ -89,6 +89,26 @@
         ],
         "visible": false
       },
+      {
```
hkfb, line 92: Let's not add this until it's ready.

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -265,10 +265,34 @@ KhMapFlat.parameters = {
     },
 };
 
-// Map3DLayer.
-const map3DLayer = exampleData[0].layers[3];
-export const Map3DLayer = EditDataTemplate.bind({});
-Map3DLayer.args = {
+// MapLayer. This is never Float32 resolution for properties.
```
hkfb, line 268: Let's not add this until we are certain that we want to replace the old version.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
```diff
@@ -37,6 +37,27 @@ export const layersDefaultProps: Record<string, unknown> = {
             step: 0,
         },
     },
+    MapLayer: {
+        "@@type": "MapLayer",
+        name: "Map",
+        id: "map3d-layer-float32",
+        pickable: true,
+        visible: false,
```
hkfb, line 45: Odd to have default visibility false. Normally if you add a layer you intend to see it.

hkfb (approved):

PR #852, opened 2022-02-25
Revert "Moved addon-redux to devDependencies and resolved dependency

shadab-skhan:
Revert "Moved addon-redux to devDependencies and resolved dependency on storybook (#832)"

Causes crash on github storybook, works fine locally

This reverts commit 53897d58cd5c3ca1d46845db1168e25a89034e53.


shadab-skhan:
Not required

Issue #850, opened 2022-02-24
Display XY axes for MapComponent


shadab-skhan:
Display XY axes in map viewer.
- [x] There should be 4 axes displayed (x bottom, x top, y left, y right)
- [x] Axes should be georeferenced to a coordinate system
- [ ] Axes should have primary ticks and secondary ticks
- [x] Primary ticks should be annotated with values.
- [ ] Annotations and Ticks should be adjusted according to zoom
- [ ] Annotations of primary tick should be round integers

shadab-skhan:
Original feature request from Viewers team
https://dev.azure.com/pdgm/EP%20Connect/_workitems/edit/11834

PR #854, opened 2022-02-25
Added axes layer.

nilscb:


shadab-skhan:
Just talked with @avanti-em she is telling that on the components are not getting rendered on the test, I misunderstood it with simple issue of component not being accessible. 

In this case we cannot disable the test, we need to find proper cause for it.
@nilscb I will try to spend some time on it.

hkfb:
Is this linked to #850?

react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
```diff
@@ -122,6 +124,12 @@ export const layersDefaultProps: Record<string, unknown> = {
         visible: true,
         selectedPie: "@@editedData.selectedPie", // used to get data from deckgl layer
     },
+    AxesLayer: {
+        "@@type": "AxesLayer",
+        name: "Axes",
+        id: "axes-layer",
+        visible: false,
```
hkfb, line 131: Default visibility should be true.

hkfb (changes requested):

hkfb (approved):

PR #870, opened 2022-03-07
Removed addon-redux dependency from DeckGlMap.

hkfb:
Causes compilation error when DeckGlMap is consumed in BLA Dashboard.


hkfb:
@shadab-skhan this is still causing compilation error with the latest release candidate

shadab-skhan (approved):

PR #873, opened 2022-03-08
Added 3D north arrow.

nilscb:


shadab-skhan:
Seems to be behaving incorrectly in 2d view, also sometime it fails to load.
image

react/src/lib/components/DeckGLMap/DeckGLMap.tsx
```diff
@@ -28,6 +28,10 @@ interface DeckGLMapProps {
         widthPerUnit?: number | null;
         position?: number[] | null;
     };
+    northArrow?: {
+        visible?: boolean | null;
+        position?: number[] | null;
```
hkfb, line 33: Would be nice also to optionally be able to override the GLTF file, in order to be able to a custom styled north arrow.

nilscb: Could we do that in a later commit?

react/src/lib/components/DeckGLMap/components/NorthArrow3D.tsx
```diff
@@ -0,0 +1,87 @@
...
+            getScale: [1, 1, 1],
+            sizeScale: 200, // 200
+            _lighting: "pbr",
+            pickable: false,
+            coordinateSystem: COORDINATE_SYSTEM.CARTESIAN,
+        }),
+    ];
+
+    return (
+        <DeckGL
```
hkfb, line 63: By adding another DeckGl component, you are consuming another GL context. In the web browser this is limited to 16. In the BLA dashboard we already consume several GL contexts (one per module). We should try to share the GL context within the map component. See https://deck.gl/docs/api-reference/core/deck#gl

hkfb (changes requested):

PR #874, opened 2022-03-08
Added gltf 3D north arrow files

nilscb:


hkfb (changes requested):
We will need these files bundled in the NPM package, so they should not be in the demo directory. Suggest moving to the "lib/assets" directory

hkfb (approved):

PR #889, opened 2022-03-16
Added SelectableGeoJsonLayer which extends functionality of GeoJsonLayer to return selected feature when clicked.

### LINKED ISSUE #688: Possible to get selected feature from a `GeoJsonLayer`
Is it/would it be possible to get the selected feature from a `Feature` in a geojson `FeatureCollection` added as a `GeoJsonLayer`?

E.g having a layer with points, it would should be possible to get the selected point through the `editedData` prop.
```json
{
    "@@type": "GeoJsonLayer",
    "data": {
        "type": "FeatureCollection",
        "features": [
            {
                "type": "Feature",
                "geometry": {
                    "type": "Point",
                    "coordinates": [420176, 7254218],
                },
                "properties": {
                    "name": "WellA",
                },
            },
            {
                "type": "Feature",
                "geometry": {
                    "type": "Point",
                    "coordinates": [421176, 7254218],
                },
                "properties": {
                    "name": "WellB",
                },
            },
        ],
    },
}

shadab-skhan:
Added SelectableGeoJsonLayer which extends functionality of GeoJsonLayer to return selected feature when clicked.
Recommended to use layer type SelectableGeoJsonLayer instead of GeoJsonLayer

Fix #688


hkfb (changes requested):
Would suggest a more descriptive name, eg. "SelectableFeatureLayer", "SelectableGeoJsonLayer" or "InteractiveFeatureLayer".

Also, would be helpful with a specific storybook example that demonstrates this layer.

hkfb (approved):

PR #908, opened 2022-03-28
Fixed issue: Unexpected cutoff opacity behavior

nilscb:
setting colorMapClampColor to false will now make this color completely transparent.

### DIFF
--- a/react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
+++ b/react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
@@ -276,8 +276,8 @@ const meshMapLayer = {
     meshValueRange: [2782, 3513],
     propertyTexture: "kh_netmap_25_m_normalized_margin.png",
     propertyValueRange: [2782, 3513],
-    colorMapRange: [2815, 3513],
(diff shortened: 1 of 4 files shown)


hkfb (changes requested):
Unable to get the behaviour where the clamp color is the edge of the color range.

Also, would be good to have a separate story for color clamping.

nilscb:
> Unable to get the behaviour where the clamp color is the edge of the color range.
> 
> Also, would be good to have a separate story for color clamping.

You get this behavior when the keyword is not used.

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -349,6 +347,48 @@ MapMaterial.parameters = {
...
+                show3D: false,
+                layerIds: [],
+            },
+        ],
+    },
+};
+MapClampColor.parameters = {
+    docs: {
+        description: {
+            story: "An example usage of ....",
```
hkfb, line 385: ... what?

hkfb (changes requested):

hkfb (approved):

PR #913, opened 2022-03-29
Dynamic tick marks and labels based on zoom

nilscb:


hkfb (commented):
It's an improvement, but I think the ticks labels are too dense, cf. the X axis:

image

hkfb (approved):

Issue #875, opened 2022-03-11
Callback for readout


hkfb:
Add a callback option to the coords prop, for example:

```
"coords": {
    "visible": false,
    "multiPicking": true,
    "pickDepth": 10,
    "callback": (params) => {renderTooltip(params)} 
}
```

- [x] The callback should be optional, and if omitted should do nothing.
- [x] `params` should on hover contain info on cursor position, including
  - [x] x, y coordinates
  - [x] hovered object name
  - [x] hovered MD if well trajectory
  - [x] hovered Log name, value if log curve
  - [x] hovered property value (if grid, property map)
  - [x] hovered depth (if depth surface)

shadab-skhan:
All the information's are provide by onHover and onMouseEvent callbacks of the component.

PR #917, opened 2022-03-31
onMouseEvent callback for DeckGL Map component

Vladimir-Kokin:
Initial version of callback for mouse events on DeckGL Map component
Get hover and click events.

See also https://github.com/equinor/webviz-subsurface-components/issues/875


hkfb (approved):
LGTM.

Issue #487, opened 2021-08-06
Create schemas for input data for Map component


hkfb:
Need schema for
- [x] Well trajectories
- [x] Well logs
- [x] Grid layers
- [x] Color tables
- [x] Pie charts
- [x] Fault polygons

- [x] Schema should be validated, and errors should be handled gracefully (ie. no crash, give a helpful warning message)

- [x] Schema validation should be optional

- [ ] Schema should have versions (should be validated by the component)

hkfb:
Check well log validation in the well log viewer. Schema file: [react/src/lib/inputSchema/WellLog.json](https://github.com/equinor/webviz-subsurface-components/commit/eee9cb68a452a22a38b150c172636ee4bf0360e7#diff-553d74a7c516f69a3170f6dd3f0d503ba475f3ca157df434e6f94333f1371f2d)

hkfb:
As the Json well log format is an array at the top level, and we don't have any use cases to modify or extend the format, we can drop versioning on the well log data.

shadab-skhan:
A new issue has been created for schema versioning #1077

PR #919, opened 2022-04-05
Create schemas for input data for Map component

shadab-skhan:
For  Well Logs #487


hkfb (commented):
Looks like the validation toggle is ignored:

image

hkfb (approved):

Issue #418, opened 2021-06-29
 Well name and symbol placed at top, bottom, @md


hkfb:
... in Map view

hkfb:
@Midtveit not sure if "@md" makes sense for well names. In this case perhaps it's better to automatically arrange labels according camera distance, similarly as contour labels?

shadab-skhan:
Created sub issues from it
#946 #947 #948 #949

hkfb:
Is this done? Is there a separate storybook example for the well labels?

PR #935, opened 2022-04-15
Well name at top and bottom of trajectory

shadab-skhan:
#418

### DIFF
--- a/react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
+++ b/react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
@@ -105,10 +105,10 @@ export const layersDefaultProps: Record<string, unknown> = {
         logCurves: true,
         refine: true,
         visible: true,
-        wellNameVisible: true,
+        wellNameVisible: false,
         wellNameAtTop: false,
         wellNameSize: 14,
-        wellNameColor: [0, 0, 255, 255],
+        wellNameColor: [0, 0, 0, 255],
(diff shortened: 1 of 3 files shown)


react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
```diff
@@ -105,6 +105,10 @@ export const layersDefaultProps: Record<string, unknown> = {
         logCurves: true,
         refine: true,
         visible: true,
+        wellNameVisible: true,
```
hkfb, line 108: Not sure if this should be default visible currently, because it looks cluttered in the examples

react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
```diff
@@ -105,6 +105,10 @@ export const layersDefaultProps: Record<string, unknown> = {
...
+        wellNameAtTop: false,
```
hkfb, line 109: Also not sure if this default makes sense. @Midtveit ?

react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
```diff
@@ -105,6 +105,10 @@ export const layersDefaultProps: Record<string, unknown> = {
...
+        wellNameSize: 14,
+        wellNameColor: [0, 0, 255, 255],
```
hkfb, line 111: Maybe subjective, but I think black is more reasonable default

hkfb (approved):

PR #938, opened 2022-04-19
Option to toggle well trajectories on/off 

### LINKED ISSUE #702: Option to toggle well trajectories on/off
Ie. ability to show only well heads

shadab-skhan:
Also added style prop to well layer, where individual/all well can be styled.
Available styles { color: [r, g, b, a], dash: [number, number]}, both color and dash can be optional.
It can take call backs too to have more specific control.

Fix #702


shadab-skhan:
> I think it would be better to expose the getLineColor prop from the GeoJsonLayer. Then you can override the opacity and color of individual trajectories, eg:
> 
> ```
> // Hide all trajectories
> getLineColor: d => {
>    return [0, 0, 0, 0];
> }
> 
> // Hide specific trajectory
> getLineColor: d => {
>   if (d.properties.name == "abc")
>      return [0, 0, 0, 0];
>   return d.properties.color;
> }
> 
> // Mark specific trajectory red
> getLineColor: d => {
>   if (d.properties.name == "abc")
>      return [255, 0, 0, 255];
>   return d.properties.color;
> }
> ```

Added style prop instead of dashed, a callback can be passed to style along with dash and color information for individual well or as an object for all the wells. Updated storybook example for the same.

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.stories.tsx
```diff
@@ -73,30 +73,35 @@ CustomDashedWells.args = {
...
 
 function callback(object: Record<string, Record<string, unknown>>) {
-    if (object["properties"]["name"] == "15/9-F-11 B") return [5, 3];
-    return [0, 0];
+    if (object["properties"]["name"] == "15/9-F-11 B")
+        return { color: [255, 0, 0, 255], dash: true };
+    else if (object["properties"]["name"] == "15/9-F-14")
+        return { dash: [10, 1] };
+    else if (object["properties"]["name"] == "15/9-F-10")
+        return { color: [0, 0, 0, 0] };
```
hkfb, line 98: I think the use case was to hide well trajectories and not well heads. If we can instead supply separate `trajectoryColor` and `headColor` then we can support hiding well head and/or trajectory separately

react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
```diff
@@ -109,6 +109,7 @@ export const layersDefaultProps: Record<string, unknown> = {
         wellNameAtTop: false,
         wellNameSize: 14,
         wellNameColor: [0, 0, 0, 255],
+        trajectoryVisible: true,
```
hkfb, line 112: Do we need this if we can set the opacity of a well trajectory to 0?

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.stories.tsx
```diff
@@ -105,16 +110,16 @@ CallbackDashedWells.args = {
         {
             "@@type": "WellsLayer",
             data: "@@#resources.wellsData",
-            dashed: callback,
+            style: callback,
```
hkfb, line 113: Maybe it would be better to have each style attributes (dashed, trajectory color, head color, etc) as optionally respective callbacks. This way you can implement only one attribute as callback and the other as values more easily and also filter on wells with different respective criteria.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.stories.tsx
shadab-skhan: yes, currently wellhead and trajectory color pic up color from same color prop, if we can separate the colors, it should be possible.

react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
shadab-skhan: Currently as wellhead and well trajectory uses same color, in order to hide only the trajectory, we will need this. This setting is global, it applies to all the wells.

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.stories.tsx
shadab-skhan: Nice suggestion! I will do the change.

react/src/lib/components/DeckGLMap/layers/layersDefaultProps.ts
hkfb: What I mean is if we add `trajectoryColor` and `headColor`, then we can hide only the trajectories. For example 
```
trajectoryColor: [0, 0, 0, 0]
```

hkfb (changes requested):
Getting this error in the Default map story:

```deck.jss:87 TypeError: update of PathLayer({id: 'wells-layer-log_curve'}): Cannot read properties of undefined (reading 'color')
    at getPath (wellsLayer.tss:275:1)
    at PathTesselator._forEachGeometry (tesselator.jss:124:1)
    at PathTesselator._rebuildGeometry (tesselator.jss:155:1)
    at PathTesselator.updateGeometry (tesselator.jss:55:1)
    at PathLayer.updateState (path-layer.jss:153:1)
    at PathLayer._updateState (layer.jss:718:1)
    at PathLayer._update (layer.jss:698:1)
    at LayerManager._updateLayer (layer-manager.jss:261:1)
    at LayerManager._updateSublayersRecursively (layer-manager.jss:214:1)
    at LayerManager._updateSublayersRecursively (layer-manager.jss:224:1)
```

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.stories.tsx
```diff
@@ -64,49 +64,80 @@ DashedWells.parameters = {
...
+    if ((object["properties"]["name"] as string).match("15/9-19"))
+        return object["properties"]["color"];
+    else if ((object["properties"]["name"] as string).match("15/9-F"))
+        return [0, 0, 0, 0];
+    else return [0, 0, 0, 255];
 }
 
-export const CallbackDashedWells = Template.bind({});
-CallbackDashedWells.args = {
+function dashCallback(object: Record<string, Record<string, unknown>>) {
```
hkfb, line 97: For the callback example, I think it would be nice to show all three use cases, ie. both hidden, dashed and unstyled. In this example all shown wells are dashed.

hkfb (approved):

PR #952, opened 2022-04-26
Update colortable version 

shruthirai:


shadab-skhan:
@hkfb We have "lockfileVersion" change with this PR to "lockfileVersion": 1

PR #957, opened 2022-04-29
DeckGLMap bounds are not updated when supplied with new bounds

### LINKED ISSUE #867: Automatically set initial rotation point to center of axis bounds in 3D (vertical and lateral)

### LINKED ISSUE #956: DeckGLMap bounds are not updated when suplied with new bounds
We render DeckGLMap while data is still being loaded with a default bounds. Once the data is loaded we send the real bounds of the data to the component, however these bounds are not updated and the area viewed at by the component remains empty (as the data is rendered elsewhere)

shadab-skhan:
1. Removed unnecessary viewState management, using initialViewState t…et deckGL manage the state
2. Set axes rotation along Y axis for 3D view
Fix #956
Fix #867


shadab-skhan:
@avanti-em One job is failing due to timeout issue, can you please check

avanti-em (approved):

hkfb:
> 1. Removed unnecessary viewState management, using initialViewState t…et deckGL manage the state
> 2. Set axes rotation along Y axis for 3D view

This feels awkward. It's difficult to rotate around the z axis now.

Also, the view now sometimes resets itself to the initial viewport. Not sure when it happens, but after zooming and panning a bit it always happens.

shadab-skhan:
I reverted my change for orbit rotation

PR #959, opened 2022-05-02
Map cypress tests

avanti-em:


avanti-em:
> Looks like the coverage stats are not affected?

adding some steps in workflow file to put coverage in codecov, (in a separete PR)

PR #967, opened 2022-05-05
Reverted orbit axis change

shadab-skhan:
Reverted change for orbit rotation #957
Handled view reset


hkfb (approved):
Still have the issue of the viewport resetting occasionally. Seems to happen when picking, eg a well trajectory, close to the bottom edge of the canvas. The tooltip causes the scrollbar to appear, and then the viewport is reset.

Issue #964, opened 2022-05-03
X axis labels overlap with the axis


hkfb:
Also, there are no tick marks:

![Screenshot from 2022-05-03 14-32-47](https://user-images.githubusercontent.com/19621822/166454250-2db15823-dce3-4db2-b683-53040666fc32.png)

PR #968, opened 2022-05-05
Fixed issue: X axis labels overlap with the axis #964

nilscb:


shadab-skhan:
image
The ticks and number are proper now, but I see X and Y axis label is missing.

hkfb (changes requested):
See @shadab-skhan 's comment

PR #971, opened 2022-05-05
Enforce minimum `npm` and `node` version

### LINKED ISSUE #640: Add `install-scripts` in a `.npmrc` file
...such that we "enforce" the usage (instead of only recommending it).

### LINKED ISSUE #877: Enforce NPM version >= 7
Optionally enforce package lock file version >= 2

anders-kiaer:
The comitted `./react/package-lock.json` is switching between `lockFileVersion` 1 and 2 (depending on which PR author the commits come from).

We should as a project stick to one version. Since `lockFileVersion` 2 has been around since `npm == 7` (`node 16`) it should be safe to require this version.

Something is not perfect with the current commited lock file in `master` (dependabot complains about not being able to read it). Hopefully this PR fixes that.

Note that an unmaintained `devDependency` needs to be removed in order to create a consistent `package.json` file, see #972.

Closes #877 and closes #640.


shadab-skhan:
Thanks @anders-kiaer this was much needed

shadab-skhan:
It breaks storybook
https://github.com/storybookjs/storybook/issues/18019
https://stackoverflow.com/questions/72016684/storybook-install-with-react

Possible solution is provided, need to try it out.

hkfb (approved):

hkfb:
> It breaks storybook, defining host in storybook script works ("storybook": "start-storybook -p 6006 -h localhost") [storybookjs/storybook#18019](https://github.com/storybookjs/storybook/issues/18019)

Works with node version 16

.github/workflows/webviz-subsurface-components.yml
```diff
@@ -41,10 +41,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Use Node.js 15.12
-        uses: actions/setup-node@v1
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
         with:
-          node-version: "15.12"
+          node-version: 18
```
hkfb, line 47: Should be < 18, until https://github.com/storybookjs/storybook/issues/18019 (as @shadab-skhan mentioned) is fixed

hkfb (changes requested):

.github/workflows/webviz-subsurface-components.yml
anders-kiaer: :+1: Changed to `node == 16` in CI.

anders-kiaer:
Perhaps I should also remove `--ignore-scripts` from `README.md` and CI commands (now that we enforce it through `.npmrc`).  Thoughts?

hkfb (approved):

PR #973, opened 2022-05-06
Measurement tool for map

### LINKED ISSUE #882: Measurement tool for map
Add a tool to measure distance between arbitrary points on a map

shadab-skhan:
Fixes #882
Measurement tool for map
Provided Escape and Delete key functionality to remove data points while drawing polyline and polygon


hkfb (changes requested):
I think I'm supposed to see a tooltip when drawing a polyline or polygon, but I don't see anything..

shadab-skhan:
> The tooltip doesn't always work.. I guess we can improve it later.

I observed this issue too, I tried checking it, it's not something with drawing layer but the getTooltip callback in Map.tsx, for some reason it returns empty PickInfo when I move cursor towards ride side of the view.

PR #975, opened 2022-05-06
New image comparison Tool - Cypress-Visual-Regression

avanti-em:
1. Use new snapshot comparison tool called cypress-visual-regression
2. Trigger an automated PR which contains new snapshots


hkfb (approved):

.github/workflows/webviz-subsurface-components.yml
```diff
@@ -97,23 +97,73 @@ jobs:
...
-      #   uses: cypress-io/github-action@33ec2f03412577367c7c6aac7b646c88e646e33c
-      #   with:
-      #     browser: firefox
-      #     working-directory: ./react
-      #     start: npm run storybook
-      #     wait-on: "http://localhost:6006"
-      # - uses: actions/upload-artifact@v2
+      - name: Cache Cypress
+        if: github.event_name != 'release' && matrix.e2e-on-firefox
+        id: cache-cypress
```
avanti-em, line 102: Cypress needs to install Cypress Binary in cache to function properly . Installation with  --ignore-scripts,  prevents Cypress installing the binary. 
Previously, "cypress-io/github-action" action was used to take care of this issue.
But recently I noticed this action is unable to handle the issue.
[Issue posted here](https://stackoverflow.com/questions/69439269/the-cypress-npm-package-is-installed-but-the-cypress-binary-is-missing-in-gi)

.github/workflows/webviz-subsurface-components.yml
```diff
@@ -97,23 +97,73 @@ jobs:
...
+        uses: actions/cache@v1
```
avanti-em, line 103: This action allows caching dependencies

.github/workflows/webviz-subsurface-components.yml
```diff
@@ -97,23 +97,73 @@ jobs:
...
+        with:
+          working-directory: ./react
+          path: ~/.cache/Cypress
```
avanti-em, line 106: restore / cache, the binary ourselves on Linux machine(pipeline)

hkfb (approved):

avanti-em:
@hkfb , please review it again
@anders-kiaer , can you please review and approve this third party action (for automated PR) :  peter-evans/create-pull-request@5a6b15373e5788c38d83296b674ad5abea28085c

cypress.sh
```diff
@@ -0,0 +1,32 @@
+git ls-files --others --exclude-standard > files.txt
```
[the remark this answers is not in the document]
avanti-em: Yes , it can be moved within cypress directory.
Pushed a commit with this change

hkfb (approved):

hkfb:
@avanti-em is it possible to strip the generated snapshots to only include the part of the screen which is being referenced?

avanti-em:
> @avanti-em is it possible to strip the generated snapshots to only include the part of the screen which is being referenced?

@hkfb , not sure if I understand. Can you please give an example

anders-kiaer:
LGTM - my only comment would be that I think I would prefer the generated snapshots that the CI compares against put in another branch (since they are basically build artifacts). Both to keep it clean in the `master` branch, but also not unnecessariy increase the size of the git history (of the `master` branch).

Would that be easy to do @avanti-em? I've approved `peter-evans/create-pull-request@5a6b15373e5788c38d83296b674ad5abea28085c`.

avanti-em:
> LGTM - my only comment would be that I think I would prefer the generated snapshots that the CI compares against put in another branch (since they are basically build artifacts). Both to keep it clean in the `master` branch, but also not unnecessariy increase the size of the git history (of the `master` branch).
> 
> Would that be easy to do @avanti-em? I've approved `peter-evans/create-pull-request@5a6b15373e5788c38d83296b674ad5abea28085c`.

Yes, I can try doing that

Issue #924, opened 2022-04-08
Improved bounds specification for maps


hkfb:
- [x] remove the global bounds, calculate automatically from union of all layers (include rotation)
- [x] should not move camera if bounds are changed
- [x] replace layer `bounds` (for colorMap, hillshading, Map3DLayer) with origin (x, y), (xinc, yinc) **or** (x-extent, y-extent). 
- [x] Should be backwards compatible (grace period)
- [x] Coordinate system handedness (left/right-handed)

Nice to have:
- [x] Rotation around same corner as Resqml
- [x] Same rotation direction as Resqml

hkfb:
Need to also handle different coordinate systems (right/left-handed)

hkfb:
Suggestions from Views team:
- Arbitrary rotation point

nilscb:
"Bounds" are changed to "Frame" property for MapLayer and Map3DLayer but not for colormap and hillshading layer as "bounds" are more suited for them, The given inout image already contains number of pixels and width height. Setting to "Done"

PR #976, opened 2022-05-06
Fixed zoom for view with axes

shadab-skhan:


react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -403,7 +403,7 @@ export const Axes = MinimalTemplate.bind({});
 Axes.args = {
     id: "axes",
     layers: [axes, meshMapLayer, north_arrow_layer],
-    bounds: [432150, 6475800, 439400, 6481500],
+    bounds: [432205, 6475078, 437720, 6481113],
```
shadab-skhan, line 406: This bound should be same as axis bounds.
Ideally, project bounds should be calculated based on data bounds of all the layers.

hkfb:
What is the problem we are solving with this? That the initial viewport is too small when the axes are shown?

shadab-skhan:
Before:
image

After: 
image

This change only corrects the bounds of storybook example for axes story.

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
hkfb: We have an issue for automatically calculating the bounds (https://github.com/equinor/webviz-subsurface-components/issues/924), but until we have that in place, then I think using the bounds itself to set the initial viewport is the best solution.

hkfb (changes requested):

shadab-skhan:
Discarding PR

PR #982, opened 2022-05-11
feat: Per well trajectory line width

### LINKED ISSUE #944: Per well trajectory line width
Would like to be able set line width for a specific well.

Similar feature to #928. Would be nice to provide similar interface, ie. callback to filter on well properties (name, etc.).

Split from #916.

shadab-skhan:
Added width in line style for wells layer
    - lineStyle: { color: [RGBA], dash: [number, number] | boolean, width: number }
    - Each property is also supported as a callback

Fixes #944


shadab-skhan:
I got tempted to keep the trajectory thickness option in the layers tool, may be that will be useful for Equinor. I have changed the name though to "Trajectory thickness scale".

hkfb:
> I got tempted to keep the trajectory thickness option in the layers tool, that may be useful for Equinor dashboard. I have changed the name though to "Trajectory thickness scale".

I think it would be better to provide an example story with a widget to control the props separate from the component. "Teach a man to fish and you feed him for a lifetime"

PR #993, opened 2022-05-13
Map readout

nilscb:


hkfb (changes requested):

hkfb (changes requested):
The kh netmap example yields incorrect readout values

shadab-skhan:
> The kh netmap example yields incorrect readout values

For me too the depth value seem to differ

hkfb:
> > The kh netmap example yields incorrect readout values
> 
> For me too the depth value seem to differ

The difference seems to be within the bounds of the `meshMaxError` threshold, which is expected.

hkfb (approved):
Tested in the PPA Dashboard and works fine with the readout callback. Would be nice to have a storybook example using a `onMouseEvent` callback.

PR #999, opened 2022-05-19
The layer should always be pickable for the readout panel to show val…

nilscb:
…ues when hovering.


react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -311,7 +311,7 @@ export default class Map3DLayer extends CompositeLayer<
             >({
                 mesh: this.state.mesh,
                 texture: this.state.texture,
-                pickable: this.props.pickable,
+                pickable: true,
```
hkfb, line 314: I think there is reason to disable picking in some circumstances. But we should be able to enable picking when the onMouseEvent prop is specified and the readout panel is hidden.

hkfb (changes requested):

PR #1006, opened 2022-05-20
Enabled web worker when loading data.

nilscb:


react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -133,6 +134,10 @@ async function load_mesh_and_texture(
         console.log("Error. One or both of texture and mesh must be given!");
     }
 
+    // Serve the worker as a local url. See: https://deck.gl/docs/developer-guide/loading-data#loaders-and-web-workers
+    const blob = new Blob([WORKER_SCRIPT], { type: "text/plain" });
+    const url = window.URL.createObjectURL(blob);
```
anders-kiaer, line 139: This would require `script-src blob:` in CSP, which is equivalent to allowing `unsafe-eval`, as confirmed by https://www.w3.org/TR/CSP2/#source-list-guid-matching:
> Especially for the [default-src](https://www.w3.org/TR/CSP2/#default_src) and [script-src](https://www.w3.org/TR/CSP2/#script_src) directives, policy authors should be aware that allowing "data:" URLs is equivalent to unsafe-inline and allowing "blob:" or "filesystem:" URLs is equivalent to unsafe-eval.

For deployed applications that want tight security settings, we could perhaps offer the component consumer to provide a `url` where the web worker script is hosted, and if not provided we do as done here (create a `blob` url` for `script` usage)?

react/src/demo/example-data/terrain-worker.js
```diff
@@ -0,0 +1,1254 @@
+(() => {
```
hkfb, line 1: Is this file needed?

hkfb (changes requested):

react/src/demo/example-data/terrain-worker.js
nilscb: Yes I think so. As an example of a not-inlined but externally served worker script

react/src/lib/components/DeckGLMap/layers/terrain/terrain-worker.ts
```diff
@@ -0,0 +1,1261 @@
+// Serve the worker as a local url.
+// See: https://deck.gl/docs/developer-guide/loading-data#loaders-and-web-workers
+// This is from the script terrain-worker.js but as a string. Some symbols are escaped.
```
anders-kiaer, line 3: According to https://deck.gl/docs/developer-guide/loading-data#loaders-and-web-workers, there exist bundles in `node_modules/@loaders.gl/<module>/dist/<name>-loader.worker.js`. Should we instead of comitting the terrain worker script to our code base, instead include it in the bundle through `raw-loader` in `webpack` (or `assets/source` in `webpack 5`) - https://v4.webpack.js.org/loaders/raw-loader/?

Pros: Reduces the amount of code we "own", and perhaps makes it easier to stay in sync with any `@loaders.gl` bumps we do in `package.json` (since they then stays in sync).

nilscb: And also we are not ablde/allowed to use raw-loader in webpack 5. (Håvard would know more. I will remove the script from example data.

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -274,6 +274,7 @@ const meshMapLayer = {
     bounds: [432205, 6475078, 437720, 6481113],
     meshMaxError: 100,
     mesh: "hugin_depth_25_m_normalized_margin.png",
+    // terrainWorkerUrl: "terrain-worker.js",  //  KEEP. Optional. Will default to an inlined version of the script.
```
hkfb, line 277: Maybe better to create a separate story using the example terrain worker script.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -419,6 +419,31 @@ Axes.args = {
     },
 };
 
+// Example using externally web worker script for loading map mesh.
```
hkfb, line 422: We should move this story to the map3DLayer.stories.tsx

PR #1010, opened 2022-05-24
Loosen engine requirements.

### LINKED ISSUE #1008: Engine requirements are too restrictive
Some projects depending on this module are using npm 6.14 and node 14.17.

hkfb:
Fixes #1008.


anders-kiaer:
LGTM.

Should we add a new one-liner in CI? E.g. prior to this line: https://github.com/equinor/webviz-subsurface-components/blob/9d4eb52b92e3d24c846bb51cff23ceccfa619160/.github/workflows/webviz-subsurface-components.yml#L52
checking if the string `"lockfileVersion": 2,` is in `package-lock.json`? Can for instance use `grep` (https://stackoverflow.com/a/4749368).

Why is the `package-lock.json` diff so big? Regenerated from scratch?

hkfb:
> LGTM.
> 
> Should we add a new one-liner in CI? E.g. prior to this line:
> 
> https://github.com/equinor/webviz-subsurface-components/blob/9d4eb52b92e3d24c846bb51cff23ceccfa619160/.github/workflows/webviz-subsurface-components.yml#L52
> 

Prefer to handle that in a separate PR.

> 
> checking if the string `"lockfileVersion": 2,` is in `package-lock.json`? Can for instance use `grep` (https://stackoverflow.com/a/4749368).
> Why is the `package-lock.json` diff so big? Regenerated from scratch?

Yes, it is.

anders-kiaer (approved):

PR #1021, opened 2022-05-25
Enable control on MapInContainer story and remove redundant file

shadab-skhan:


react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -644,11 +644,16 @@ const useStyles = makeStyles({
...
     return (
         <div className={classes.main}>
-            <DeckGLMap {...exampleData[0]} />
+            <DeckGLMap {...args} />
         </div>
     );
 };
+
+MapInContainer.args = {
+    ...exampleData[0],
```
hkfb, line 658: I think we should avoid using the exampleData, as it is a specific example of using a json file to create the map props, which most consumers probably won't do.

hkfb (approved):
I think for longer term we should write stories in tsx. It will make it easier to follow the examples for consumers using typescript.

Issue #1018, opened 2022-05-24
`ContinuousLegend` and `DiscreteLegend` should have an `id` prop


HansKallekleiv:
Using multiple instances of these components does not work as they will have the same id.
Should expose the `id` in these and the underlying components as props.

https://github.com/emerson-eps/color-tables/blob/main/react-app/src/component/Legend/ContinuousLegend.tsx#L198

PR #1025, opened 2022-05-25
Make legend to show for multiple color map layer

shruthirai:


shruthirai:
@HansKallekleiv Can you also test if its solve the issue(#1018), as I have made changes in latest colortable repo version(v0.2.33) by adding id as prop but have not tested it from my side, so not sure if that will work or not

HansKallekleiv:
Thanks @shruthirai . There is now a legend for each colormap/view .
It doesn't seem to update when the layers change though.
Try e.g. changing the `colorMapName` in the Multi Color Map storybook.

shruthirai:
> Thanks @shruthirai . There is now a legend for each colormap/view . It doesn't seem to update when the layers change though. Try e.g. changing the `colorMapName` in the Multi Color Map storybook.

Yes, even I noticed that. Working on that. Thanks

react/src/lib/components/DeckGLMap/layers/colormap/colormapLayer.ts
```diff
@@ -153,6 +153,18 @@ export default class ColormapLayer extends BitmapLayer<
             },
             moduleParameters: mergedModuleParams,
         });
+
+        const legendProps = [];
```
shadab-skhan, line 157: this should not be done in draw method, appropriate place to set this state would have been in onDataLoad or initializeState

react/src/lib/components/DeckGLMap/components/ColorLegend.tsx
```diff
@@ -92,27 +29,46 @@ const ColorLegend: React.FC<ColorLegendProps> = ({
...
-                (legend, index) =>
-                    legend.visible && (
-                        <div key={index}>
-                            {legend.discrete && (
+            {layers.map(
+                (layer, index) =>
+                    layer?.props?.visible &&
+                    layer?.state?.legend?.[0] && (
+                        <div style={{ marginTop: 30 }} key={index}>
+                            {layer?.state?.legend?.[0].discrete && (
```
shadab-skhan, line 37: I suggest to remove this indexing.

shadab-skhan (changes requested):

react/src/lib/components/DeckGLMap/components/ColorLegend.tsx
shruthirai: Will do this changes in another PR

shruthirai:
@shadab-skhan I have to do bit of refactoring but will raise another PR for that .

react/src/lib/components/DeckGLMap/layers/colormap/colormapLayer.ts
```diff
@@ -187,6 +187,54 @@ export default class ColormapLayer extends BitmapLayer<
...
+        const valueRangeMax = this.props.valueRange[1] ?? 1.0;
+
+        // If specified color map will extend from colorMapRangeMin to colorMapRangeMax.
+        // Otherwise it will extend from valueRangeMin to valueRangeMax.
+        const min = this.props.colorMapRange?.[0] ?? valueRangeMin;
+        const max = this.props.colorMapRange?.[1] ?? valueRangeMax;
+
+        legendProps.push({
+            discrete: false,
+            metadata: { objects: {} },
```
shadab-skhan, line 231: There is code duplicacy in initializeState and updateState, use legendProps as normal object instead of array, and just update coloeMapName in updateState so it can be made consise.

shadab-skhan (approved):

react/src/lib/components/DeckGLMap/layers/colormap/colormapLayer.ts
shruthirai: Yes, have noticed that. That will be separate PR (Code refcator).

PR #1026, opened 2022-05-27
Sync N LogViewers

Vladimir-Kokin:
- Synchronize variable number of well log views. A number of views is a number of **props.wellogs** items (0 or greater)
- Storybook example modified to 3 wells (Or 2 wells example should be remained?)


shadab-skhan (approved):
Works fine!

PR #1036, opened 2022-06-07
Color scale update

### LINKED ISSUE #1030: ColorLegend is not updated on first render when changing `ColormapLayer` props (e.g. `colorMapRange`, `colorMapName`)
https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/deckglmap--multi-color-map
1. Change `colorMapRange` in one layer
2. Note that there is no change in the colorlegend.
3. Move mouse cursor into the map component, or change the `colorMapRange` prop again
4. Note that the colorlegend now updates.

Expected:
Legend should update on prop change

shadab-skhan:
Fixes  #1030


HansKallekleiv (approved):
Tested and verified fixing the colorscale issue seen from Dash.

PR #1058, opened 2022-06-16
fix wrong argument name

siner308:


shadab-skhan:
Discussed with @HansKallekleiv, it's fixed, hence closing it.

PR #1060, opened 2022-06-17
Intersection view unfolded path calculation

hkfb:


react/src/lib/components/DeckGLMap/layers/intersection/unfoldedGeoJsonLayer.ts
```diff
@@ -0,0 +1,53 @@
...
+import UnfoldedPathLayer from "./unfoldedPathLayer";
+import { Feature } from "geojson";
+import { zip } from "lodash";
+import { distance } from "mathjs";
+
+interface UnfoldedGeoJsonLayerProps<D> extends GeoJsonLayerProps<D> {
+    isIntersectionView?: boolean;
+}
+
+function getUnfoldedPath(object: any) {
```
hkfb, line 11: Added abscissa calculation here.

react/src/lib/components/DeckGLMap/layers/intersection/intersectionView.stories.jsx
```diff
@@ -0,0 +1,86 @@
+import React from "react";
```
hkfb, line 1: I think we should make this a tsx file. It will make it easier to adapt the examples in a typescript environment.

shadab-skhan (approved):

PR #1062, opened 2022-06-20
Fix for wrong color scale in legend

shruthirai:


shruthirai:
Not sure if this is the correct way. If we want color legend to adapt colormapfunction(value interpolation), then guess it will be new feature(custom scale). Its my assumption since now legend takes the color name and based on that just display the legend (linear gradient)

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -435,7 +436,10 @@ colorMapFunction.args = {
         {
             ...colormapLayer,
             image: "https://raw.githubusercontent.com/equinor/webviz-subsurface-components/master/react/src/demo/example-data/propertyMap.png",
-            colorMapFunction: (x) => [255 - x * 100, 255 - x * 100, 255 * x], // If defined this function will override the colormap.
+            //colorMapFunction: (x) => [255 - x * 100, 255 - x * 100, 255 * x], // kept for reference
```
hkfb, line 439: The color legend should work also with custom color mapping functions

hkfb (changes requested):
The color legend should work also with

hkfb (changes requested):
The color legend doesn't look correct. There is a sharp transition at around 3147 m in the legend, which is not present in the map or the function.

Also the legend does not update when changing the color mapping function, while the map does.

shruthirai:
> The color legend doesn't look correct. There is a sharp transition at around 3147 m in the legend, which is not present in the map or the function.
> 
> Also the legend does not update when changing the color mapping function, while the map does.

Resolved. I was using math.round before which was the reason for sharp transition.

hkfb (approved):

PR #1065, opened 2022-06-22
Use same layer to create unfolded view in Intersection view 

shadab-skhan:
using viewport type


hkfb:
Re. the build failure, seems to be a transitive dependency to setuptools_scm 7.0.1, which recently has dropped python 3.6 support, cf. https://github.com/pypa/setuptools_scm/blob/main/CHANGELOG.rst

PR #1084, opened 2022-06-27
Legend for 3d map layer

shruthirai:


react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -284,6 +284,7 @@ const meshMapLayer = {
     contours: [0, 100.0],
     isContoursDepth: true,
     colorMapName: "Physics",
+    colorMapFunction: (x) => [255 - x * 100, 255 - x * 100, 255 * x], // If defined this function will override the colormap.
```
hkfb, line 287: We already have another story that shows the usage of the color map function. You could add the color legend to that story instead.

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -404,6 +404,23 @@ export default class Map3DLayer extends CompositeLayer<
         );
         return [layer];
     }
+
+    getLegendData(): ContinuousLegendDataType {
+        const min = this.props.colorMapRange
```
hkfb, line 409: Should also check if `propertyTexture` is defined here.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
hkfb: > @hkfb but propertyTexture is an string . How will I check that for range ? You mean to show legend only when propertyTexture is defined ?

If `propertyTexture` is not defined or empty, then the `propertyValueRange` is not used for color sampling. The `meshValueRange` is used instead.

PR #1095, opened 2022-06-29
Changed bounds specification and defalt rotation point (origo = minx,…

nilscb:
…miny)


react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -46,18 +67,15 @@ function mapToRange(resolved_mesh: MeshType, meshValueRange: [number, number]) {
 function add_normals(
     resolved_mesh: MeshType,
     meshImageData: ImageData,
-    bounds: [number, number, number, number]
+    bounds: BoundsType
```
hkfb, line 70: I think we need a new name for this prop, so that we can keep but deprecate the old prop for a while.

Also, I think we could be more explicit in the `BoundsType`. For example:

```
bounds: {
  increment: [number, number],
  origin: [number, number],
  count: [number, number],
}

We can also make `count` optional in this case, because it's given by the data structure, but could serve as an assertion.

**Review by hkfb — CHANGES_REQUESTED** [review]

File: react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts (hunk 1 above)
Comment (nilscb) [reply react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts:70]: Maybe I can rename bonods to horizontalExtent. I think I need count otherwise it will not be possible to caclutale xmin, xmax ..

Comment (hkfb) [reply react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts:70]: > Maybe I can rename bonods to horizontalExtent. I think I need count otherwise it will not be possible to caclutale xmin, xmax ..

You have `ni` and `nj` given by the png image format. You can calculate the x and y extents from `xIncrement * ni` and `yIncrement * nj`

Comment (hkfb) [reply react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts:70]: > Maybe I can rename bonods to horizontalExtent.

How about "frame"?

File: react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts [hunk 2]
```diff
@@ -342,9 +347,14 @@ export default class Map3DLayer extends CompositeLayer<
 > {
     initializeState(): void {
         // Load mesh and texture and store in state.
+        const isBounds = typeof this.props.bounds !== "undefined";
+        const bounds = (
```
hkfb, line 351: This should be the other way around; if `frame` is defined, then it should be used instead of `bounds`.

Also, would be useful with a deprecation warning here if `bounds` are specified. Eg.

```
console.warn("\"bounds\" is deprecated. Use \"frame\" instead.);
```

hkfb (changes requested):

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -414,7 +419,7 @@ Axes.args = {
         viewports: [
             {
                 id: "view_1",
-                show3D: false,
+                show3D: true,
```
hkfb, line 422: Not sure why you changed this. I think it's good to have this example ortographic.

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -255,19 +273,22 @@ export interface Map3DLayerProps<D> extends ExtendedLayerProps<D> {
...
 
     // Mesh error in meters. The output mesh is in higher resolution (more vertices) if the error is smaller.
     meshMaxError: number;
 
     // Url to png image for map properties. (ex, poro or perm values as a texture)
     propertyTexture: string;
 
-    // Rotates map counterclockwise in degrees around 'rotPoint'.
+    // Rotates map counterclockwise in degrees around 'rotPoint' specified below.
     rotDeg: number;
```
hkfb, line 289: Not urgent, but I think we should move also `rotDeg` and `rotPoint` to the `frame` prop.

hkfb (changes requested):
We should update all the stories using the Map3DLayer to use the new `frame` prop.

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
nilscb: Can I do that in next pull request?

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
nilscb: Fixed

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -285,12 +296,6 @@ export interface Map3DLayerProps<D> extends ExtendedLayerProps<D> {
     // Url to png image for map properties. (ex, poro or perm values as a texture)
     propertyTexture: string;
 
-    // Rotates map counterclockwise in degrees around 'rotPoint' specified below.
-    rotDeg: number;
```
hkfb, line 268: This prop and possible `rotPoint` is used in some contexts, so we should keep it but deprecate it, similarly the bounds prop.

hkfb (changes requested):
We should update all the stories using the Map3DLayer to use the new props.

react/src/lib/components/DeckGLMap/layers/terrain/map3DLayer.ts
```diff
@@ -326,9 +358,19 @@ export default class Map3DLayer extends CompositeLayer<
 > {
     initializeState(): void {
         // Load mesh and texture and store in state.
+        const isBounds = typeof this.props.bounds !== "undefined";
+        if (isBounds) {
```
hkfb, line 362: Should also warn of deprecation if `rotDeg` or `rotPoint` is defined outside the `frame` prop.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
```diff
@@ -273,14 +273,18 @@ KhMapFlat.parameters = {
...
-    bounds: [432205, 6475078, 437720, 6481113],
-    meshMaxError: 100,
     mesh: "hugin_depth_25_m_normalized_margin.png",
     meshValueRange: [2782, 3513],
+    // Either "bounds" or "frame". "bounds" will be deprecated."
+    //bounds: [432205, 6475078, 437701, 6480898],  // [xmin, xmax, ymin, ymax]
+    frame: {
+        origin: [432205, 6475078],
+        count: [229, 291],
+        increment: [24, 20],
```
hkfb, line 283: Should be `[25, 25]`

hkfb (changes requested):
The map doesn't seem to update when changing the `frame.increment`

react/src/lib/components/DeckGLMap/storybook/DeckGLMap.stories.jsx
nilscb: Fixed

hkfb (approved):

Issue #1003, opened 2022-05-20
Render continuous legend upwards increasing


hkfb:

shruthirai:
Fixed

PR #1099, opened 2022-07-01
adding reverse range prop

shruthirai:


shruthirai:
Fix for https://github.com/equinor/webviz-subsurface-components/issues/1003

react/src/lib/components/DeckGLMap/DeckGLMap.tsx
```diff
@@ -33,6 +33,7 @@ export interface DeckGLMapProps {
         visible?: boolean | null;
         cssStyle?: Record<string, unknown> | null;
         horizontal?: boolean | null;
+        reverseRange?: boolean | null;
```
hkfb, line 36: Not sure we need this as a global prop. I think it's enough that we have it as an option on the color legend itself.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/components/ColorLegend.tsx
```diff
@@ -26,12 +26,14 @@ interface ColorLegendProps {
     horizontal?: boolean | null;
     layer: ExtendedLayer<unknown>;
     colorTables: colorTablesArray | string | undefined;
+    reverseRange?: boolean | null;
```
hkfb, line 29: Given the `?` this can be `bool` or `undefined`, so I don't think `null` is needed also.

hkfb (changes requested):

react/src/lib/components/DeckGLMap/components/ColorLegend.tsx
shruthirai: Agree. It was throwing lint error earlier so null was given but for latest change null is not required.

hkfb (approved):

PR #1100, opened 2022-07-05
Color legend for well logs is not updated when changing the well log

### LINKED ISSUE #984: Color legend for well logs is not updated when changing the current well log

shadab-skhan:
fix #984


hkfb (changes requested):
After changing log from eg. ZONELOG to POROSITY, then the legend disappears. The legend only appears after pointing the cursor somewhere in the canvas.

hkfb (changes requested):
Problem persists with the Discrete Well Logs story.

Issue #1068, opened 2022-06-22
Project a well trajectory in intersection view


hkfb:
Create intersection geometry from well trajectory and project trajectory on intersection

Intersected geometry should show both in intersection and 3d views

PR #1116, opened 2022-07-11
Added story to show well in intersection view

shadab-skhan:


react/src/lib/components/DeckGLMap/layers/intersection/unfoldedGeoJsonLayer.ts
```diff
@@ -15,11 +17,10 @@ function getUnfoldedPath(object: Feature) {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const a: any[] = [];
     delta.forEach((d) => {
-        const prev = a.at(-1) || 0;
+        const prev = a.at(-1) || first_x;
```
hkfb, line 20: World coordinates don't make sense in this projection. Instead X should mean distance from the beginning of the fence. We should change the axis origin in this viewport 0.

react/src/lib/components/DeckGLMap/layers/intersection/unfoldedGeoJsonLayer.ts
```diff
@@ -15,11 +17,10 @@ function getUnfoldedPath(object: Feature) {
...
         a.push(d + prev);
     });
-    const planeY = 2000;
-    const vAbscissa = zip(a, [...a].fill(planeY), z);
+    const vAbscissa = zip(a, y, z);
```
hkfb, line 23: What is the purpose changing the y here?

shruthirai (approved):

react/src/lib/components/DeckGLMap/layers/intersection/unfoldedGeoJsonLayer.ts
shadab-skhan: Changed back to 0

shadab-skhan:
fix #1068

PR #1124, opened 2022-07-15
Use sparse interval curve in Discrete well log story

hkfb:


shruthirai:
image

Unexpected result while changing the logname from ZONELOG to ZONE_MAIN

shadab-skhan:
> image
> 
> Unexpected result while changing the logname from ZONELOG to ZONE_MAIN

As there is only one log (ZONELOG) in the file the behavior is correct.

shadab-skhan (approved):

Issue #925, opened 2022-04-08
Set/get initial/current camera position


hkfb:
- [x] Should be optional. Default behaviour is to automatically zoom and focus on bounds of the data.

- [x] Also need to get the current camera position. (Callback?)

- [x] Create story with camera controls

VincentNevermore:
Closes #925

PR #1131, opened 2022-07-26
add a prop for customized camera postion in the deckgl component

VincentNevermore:
image
With user customer defined camera position

image
With default camera position

### DIFF
--- a/react/src/lib/components/DeckGLMap/components/Map.tsx
+++ b/react/src/lib/components/DeckGLMap/components/Map.tsx
@@ -241,9 +241,6 @@ function defaultTooltip(info: PickInfo<unknown>) {
     return feat?.properties?.["name"];
 }
 
-// function defaultCameraPosition() {
-//    return {}
-// }
 const Map: React.FC<MapProps> = ({
     id,
     resources,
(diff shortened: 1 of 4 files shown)


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -589,11 +600,24 @@ const Map: React.FC<MapProps> = ({
                     scaleUnit={coordinateUnit}
                 />
             ) : null}
-
             <StatusIndicator layers={deckGLLayers} isLoaded={isLoaded} />
-
             {coords?.visible ? <InfoCard pickInfos={hoverInfo} /> : null}
-
+            {Object.keys(getCameraPosition).length !== 0 ? (
```
shadab-skhan, line 605: We can have this div component in storybook, if it's here this div will be visible on the consuming component side too.

react/src/lib/components/DeckGLMap/storybook/components/Map.stories.jsx
```diff
@@ -27,10 +27,15 @@ const Template = (args) => {
             setEditedData={(updatedProps) => {
                 setEditedData(updatedProps.editedData);
             }}
+            getCameraPosition={{
+                target: [396638.95884797024, 6467831.83598219],
```
shadab-skhan, line 31: We can skip passing the prop here, as it's an optional prop

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -240,6 +241,9 @@ function defaultTooltip(info: PickInfo<unknown>) {
     return feat?.properties?.["name"];
 }
 
+// function defaultCameraPosition() {
```
shadab-skhan, line 244: We can remove this

shadab-skhan (changes requested):

shadab-skhan:
fix #925

PR #1132, opened 2022-07-26
All triangles

nilscb:


shadab-skhan (approved):
There are couple of messages on console related to bounds being deprecated, we can fix it too in the next patch.

PR #1133, opened 2022-07-27
Fixed two stories not working

nilscb:
+ fix use of normal vector in shader
+ stories didnt respond to changes on all properties


shadab-skhan:
image
In the Big Map With Hole story I tried to hover cursor on the annotated spot, the depth value of NaN got reported in the info card.

react/src/lib/components/DeckGLMap/layers/map/mapLayer.stories.tsx
```diff
@@ -384,11 +385,11 @@ export const MapDiscontinuous: ComponentStory<typeof DeckGLMap> = () => {
     );
 };
 
-MapDiscontinuous.parameters = {
+BigMapWithHole.parameters = {
```
shadab-skhan, line 388: At some region depth is read as NaN

react/src/lib/components/DeckGLMap/layers/map/mapLayer.stories.tsx
```diff
@@ -470,3 +474,60 @@ BreakpointColorMap.parameters = {
         },
     },
 };
+
+export const ColorMapRange: ComponentStory<typeof DeckGLMap> = (args) => {
```
shadab-skhan, line 478: Map3DDelatinMeshLayer/ColorMapRange example is interactive, it will be nice if we can make this story interactive too.

nilscb: Can I do that in next pull request? The story is copied from a simmilar story in terrain folder so I will have to figure out how to fix these.

nilscb:
The readout should be fixed now. No more "Nan"s

react/src/lib/components/DeckGLMap/layers/map/mapLayer.stories.tsx
nilscb: Is fixed now.

shadab-skhan (approved):
Also, we can make the storybook controls working in next PR

Issue #1107, opened 2022-07-06
Filter well heads by name or attribute


hkfb:

shadab-skhan:
Fixed with #1139

PR #1139, opened 2022-07-28
add feature: well head hidden

VincentNevermore:
image


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -268,6 +268,7 @@ const Map: React.FC<MapProps> = ({
         getViewState(bounds, zoom)
     );
 
+    console.log(layers);
```
shadab-skhan, line 271: We can remove the console log.

react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.stories.tsx
```diff
@@ -216,12 +216,14 @@ AllTrajectoryHidden.args = {
     layers: [
         {
             ...defaultProps.layers[0],
-            lineStyle: { color: [0, 0, 0, 0] },
+            lineStyle: { color: [100, 100, 100, 100] },
+            wellHeadSyle: { pointRadiusScale: 0 },
```
shadab-skhan, line 220: pointRadiusScale can be a callback function too, similar to lineStyle.
With that we will be able to hide the wellhead selectively.
Example, wellHeadStyle: {size?: callback | number, color?: callback | RGBAColor }

shadab-skhan (approved):

shadab-skhan:
#1107

Issue #1096, opened 2022-07-01
Show well picks in Correlation-view


hkfb:
- [X] Show lines for well picks for correlation-view, across several wells
- [X] Show lines in color according to pick name 
- [x] Show fill color between pick lines in the area between wells (like RMS)
- [x] Add Depth label in its depth tracks for MD and TVD.
- [x] Allow flattening on one (1) well pick 
- [x] Make the width between wells customizable , individual with specified between each pair of wells
- [x] Fill color and/or patterns in welllog view
- [x] Show fillpattern between pick lines in the area between wells (like RMS)
- [x] Allow flattening on 2 well picks

hkfb:
See also https://github.com/equinor/webviz-subsurface-components/issues/1051

hkfb:
@Midtveit @Vladimir-Kokin @HansKallekleiv @shadab-skhan @nilscb any thoughts about data format for well picks?

Do we have well picks in RoxarAPI, and what is the data format there?

HansKallekleiv:
We are using wellpicks in webviz  for the deckglmap.
Example on format we chose and script to extract with roxar api here: https://github.com/equinor/webviz-subsurface-testdata/tree/master/observed_data%2Fdrogon_well_picks

Vladimir-Kokin:
@hkfb It seems #1348 should close this issue (last check in issue description)
![изображение](https://github.com/equinor/webviz-subsurface-components/assets/86053741/fe763939-e999-43f2-bb2b-ffc0948f69a7)

PR #1140, opened 2022-08-01
Well log viewer well picks

Vladimir-Kokin:
Show well picks on WellLogView
#1096 

Use WellLog JSON format


react/src/lib/components/WellLogViewer/MapAndWellLogViewer.tsx
```diff
@@ -113,6 +111,15 @@ function addTemplateTrack(
     return templateNew;
 }
 
+const wellpick = {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    wellpick: require("../../../demo/example-data/wellpicks.json")[0],
```
shadab-skhan, line 116: Fixed data is used for this component, shouldn't it come from a prop?

shadab-skhan:
Hor_1 and Hor_6 are not visible in the track, is it expected?
image

shadab-skhan (approved):

Vladimir-Kokin:
@shadab-skhan Hor_1 and Hor_6 is out of range of MD for the well.
But possible we would have an option to exand MD range to show all well picks

PR #1145, opened 2022-08-03
Camera position

VincentNevermore:


shadab-skhan (commented):
Let's simplify it by introducing 2 new props 1. viewState and 2. onViewStateChange callback.
And use this viewState directly as a prop to DeckGL viewState. And pass DeckGL onViewStateCallback as it is to the client.

shadab-skhan (approved):

PR #1151, opened 2022-08-04
Multiple

### LINKED ISSUE #986: Multi selection of wells in DeckGl component map. hold 'ctrl' to select multi wells and fire callback after 'ctrl' is released
I am currently working with DeckGl Map components in webviz and i am trying to achieve a circular filtering of wells. I have a dropdown menu to filter the DeckGl well layer and i am using editedData parameter to track the user selection of wells inside the map. The challenge is that in editedData the user may select one well at a time, this will render the map and the dropdown to one well. This will leave no other wells to select on the map, the solution i am looking for is selection of multiple wells on the map before the callback is triggered. Is there any possibility to hold 'ctrl' key (or other options) and select all wells desired and run the callback after releasing the 'ctrl' key. This will resemble the range slider where the updatemode='mouseup' will only call the callback after the mouse button is released.

VincentNevermore:
add multi selection on wells with ctrl key pressed
fix #986

### DIFF
--- a/react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
+++ b/react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.ts
@@ -219,7 +219,6 @@ export default class WellsLayer extends CompositeLayer<
                 selectedWell: (info.object as Feature).properties?.["name"],
                 multiSelectedWells: this.state.multipleSelectionWells,
             });
-            console.log(this.state.multipleSelectionWells);
             return true;
         }
     }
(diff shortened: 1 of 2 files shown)


shadab-skhan (commented):
The change works fine, there should also be a mechanism to clear the multipleSelectionWells. It can be like selection in windows, when we press ctrl + click, multiple files/folders gets selected, and when you do a click without ctrl pressed, the selection is cleared.

shadab-skhan (approved):
works perfect!

PR #1153, opened 2022-08-07
Hide wellhead for intersection view example

shadab-skhan:


react/src/lib/components/DeckGLMap/layers/intersection/intersectionView.stories.tsx
```diff
@@ -86,8 +86,7 @@ WithWellsData.args = {
         number
     ],
     resources: {
-        wellsData:
-            "https://raw.githubusercontent.com/equinor/webviz-subsurface-components/master/react/src/demo/example-data/volve_wells.json",
+        wellsData: "./volve_wells.json",
     },
     layers: [
         {
```
VincentNevermore, line 92: For here, Shall we create a new story instead or the previous one is no longer needed ?

VincentNevermore (approved):

react/src/lib/components/DeckGLMap/layers/intersection/intersectionView.stories.tsx
shadab-skhan: It's not required, there was an existing story in the intersectionview/WithWellsData. There, I wanted to show only one well, so we hide other trajectories, but their well heads were still visible as there was no way to hide them. Now, as your solution to hide the wellhead is merged, I used that to hide unnecessary wellheads.

Issue #997, opened 2022-05-18
Annotations illegible on dark background


hkfb:
The following annotations are illegible when choosing dark background in storybook.

- [x] Axes, including ticks and labels
- [x] North arrow
- [x] scale ruler

![image](https://user-images.githubusercontent.com/19621822/169041955-dcd91200-2136-4d34-a611-d38c4a0911da.png)

PR #1158, opened 2022-08-15
add a boolean prop for user to switch between black and white theme a…

VincentNevermore:
image

image

### DIFF
--- a/react/src/lib/components/DeckGLMap/DeckGLMap.tsx
+++ b/react/src/lib/components/DeckGLMap/DeckGLMap.tsx
@@ -69,8 +69,15 @@ export interface DeckGLMapProps {
     cameraPosition?: ViewStateType | undefined;
 =======
 
+<<<<<<< HEAD
     distanceRuleColorScale?: Record<string, string>;
 >>>>>>> 1d3cef1 (update)
+=======
+    /**
+     *  Override distance ruler style with css input style
+     */
+    distanceRulerStyle?: Record<string, string>;
+>>>>>>> cfed744 (update)
 }
 
 const DeckGLMap: React.FC<DeckGLMapProps> = ({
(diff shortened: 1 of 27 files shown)


react/src/lib/components/DeckGLMap/DeckGLMap.tsx
```diff
@@ -60,6 +60,8 @@ export interface DeckGLMapProps {
      * Override default tooltip with a callback.
      */
     getTooltip?: TooltipCallback;
+
+    distanceRuleColorScale?: Record<string, string>;
```
shruthirai, line 64: It would be good if you add the comment here as shown for rest of the props

react/src/lib/components/DeckGLMap/layers/map/mapLayer.stories.tsx
```diff
@@ -196,6 +196,39 @@ MapLayer2d.parameters = {
...
+            },
+        ],
+    },
+};
+
+MapLayer2dDarkMode.parameters = {
+    docs: {
+        ...defaultParameters.docs,
+        description: {
+            story: "Example using png as mesh and properties data.",
```
shruthirai, line 226: It would be good if you give bit clear description in line 226 like in bracket dark mode or whatever you prefer, just to know that this is done for dark mode. Just an suggestion, you can also keep as it is, no problem :)

react/src/lib/components/DeckGLMap/DeckGLMap.tsx
```diff
@@ -81,6 +86,7 @@ const DeckGLMap: React.FC<DeckGLMapProps> = ({
     onMouseEvent,
     selection,
     getTooltip,
+    distanceRulerStyle,
```
shadab-skhan, line 89: I don't recommend adding a new prop to the Map component. There is a scale prop for distanceScale component, you can update that prop to pass styles.

shadab-skhan (changes requested):

shadab-skhan (approved):

shadab-skhan:
#997

PR #1159, opened 2022-08-16
VTK Grid data for testing

HansKallekleiv:
Testdata for a 3D grid in VTK polydata format.

There are two datasets:
1. The visible cells of a full 3D grid (vertices, polygons and scalar).
2. The intersected cells along a well trajectory (vertices and polygons).

`.np` are original numpy arrays.
`.json` same file converted to json for convienience.
`b64.json` base64 encoded arrays in json. - This is the currently preferred format.


hkfb:
@HansKallekleiv @nilscb is this still relevant?

PR #1164, opened 2022-08-17
fix WellsLayer.onClick to allow generating of onMouseEvent

Vladimir-Kokin:
Fix a bug described in https://github.com/equinor/webviz-subsurface-components/issues/1163


shadab-skhan (approved):
Thanks, it works fine.

One observation, I get this warning when I click the well.
image

Issue #1106, opened 2022-07-06
Show fence bounds in intersection view


hkfb:
- [ ] Show fence outline in map (2D and 3D) and intersection view
- [ ] Show axes with tick marks and labels
  - [ ]  Show abscissae
  - [ ]  Show X, Y, Z
  - [ ] Show MD

PR #1166, opened 2022-08-18
Show unfolded fence in intersection view

shadab-skhan:
Created storybook example with sample fence, surface and wells data
#1106


shadab-skhan:
#1106

react/src/lib/components/DeckGLMap/layers/intersection/unfoldedGeoJsonLayer.ts
```diff
@@ -23,64 +21,51 @@ function getUnfoldedPath(object: Feature): Position3D[] {
     });
     const vAbscissa = zip(a, [...a].fill(planeY), z);
 
-    return vAbscissa as Position3D[];
+    return vAbscissa as Position[];
```
VincentNevermore, line 24: For here, can we add a type check to make the code more safe rather than using as to enforce the type.

VincentNevermore (approved):

Issue #1017, opened 2022-05-24
Filling in SyncedLogViewer MDOA track changes with zoom level


Midtveit:
Zoomed out the filling looks correct.
![image](https://user-images.githubusercontent.com/42234769/170029824-7cbe3cd0-4708-4e1b-83ea-e6fa9a8fa3fc.png)

Zooming in makes filling get a stripe see picture below.
And filling changes with zoom level
![image](https://user-images.githubusercontent.com/42234769/170030092-c0bfc1d4-13dc-4a82-a7c0-7efccb45fe48.png)

Vladimir-Kokin:
It also can be reproduced in WellLogViewer Default story when TVD primary scale is used
![изображение](https://user-images.githubusercontent.com/86053741/170101865-9a3cbe5c-e10a-4eaf-88cb-d2f443c18b85.png)

shadab-skhan:
Waiting for response from Videx team

PR #1168, opened 2022-08-22
null depth data workaround for videx wellog library

Vladimir-Kokin:
A workaround for null values in welllog depth data.
See details https://github.com/equinor/videx-wellog/issues/165
and https://github.com/equinor/webviz-subsurface-components/issues/1017

### DIFF
--- a/react/src/lib/components/WellLogViewer/utils/tracks.ts
+++ b/react/src/lib/components/WellLogViewer/utils/tracks.ts
@@ -969,12 +969,12 @@ function createAreaData(
 }
 
 async function createStackData(
-    data: [number, number | string][],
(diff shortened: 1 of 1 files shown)


shadab-skhan (approved):

shadab-skhan:
Looks like there is another issue in MTOA track
image

Vladimir-Kokin:
> Looks like there is another issue in MTOA track 
It is due to non-monotonous behaviour on TVD data channel in some interval of MD [4973,4974]
изображение

So have backward-moving lines on the graphs 
изображение

I have mention the problem in https://github.com/equinor/videx-wellog/issues/165#issuecomment-1218492861

PR #1171, opened 2022-08-24
Changes for adding range for legend

shruthirai:


shadab-skhan:
image
Can we update the UI? The label is gone after entering the range.
Suggest to make Min: input box Max: Input box

shruthirai:
> image Can we update the UI? The label is gone after entering the range. Suggest to make Min: input box Max: Input box

Done

shadab-skhan:
The UI looks good, I observe a small issue that scale does not reflect min number when it's below 1000
image

shruthirai:
> The UI looks good, I observe a small issue that the scale does not reflect min and max when the value is out of data range image

Done. Currently you will see only 3 values because else it overlaps the values if there are more values, we can find some workaround for this in future if requirement is to show more values in the legend.

shadab-skhan (approved):

PR #1189, opened 2022-09-09
1104 separate viewport

### LINKED ISSUE #1104: Separate viewport control for intersection viewport

VincentNevermore:
fixes #1104


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -600,7 +620,7 @@ const Map: React.FC<MapProps> = ({
             {scale?.visible ? (
                 <DistanceScale
                     {...scale}
-                    zoom={viewState?.zoom}
+                    zoom={-5}
```
shadab-skhan, line 623: This breaks scalebar to adapt to the changing zoom. We can take the zoom from 1st view state to fix this.
If users will use multi-views, they can hide the default scalebar and use the distance scale component to add a separate scalebar for each view.

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -64,6 +64,12 @@ export interface ViewportType {
      * Layers to be displayed on viewport
      */
     layerIds?: string[];
+
+    target?: number[];
```
shadab-skhan, line 68: We can type target as [number, number] to enforce the array length.

shadab-skhan:
It works fine now, but there is an issue, when I change the default story to 3d view it's broken.

shadab-skhan (approved):

Issue #1190, opened 2022-09-13
Remove support for Python 3.6 and 3.7


jorgenherje:
Make Python version 3.8 minimum required version as `webviz-config` and `webviz-subsurface` now has Python 3.8 as minimum required version.

PR #1194, opened 2022-09-14
Drop Python `3.6` and `3.7` support

anders-kiaer:
Closes https://github.com/equinor/webviz-subsurface-components/issues/1190.


jorgenherje (approved):
Looks good to me! 👍

PR #1200, opened 2022-09-15
Allow editing of breakpoints

shruthirai:


shadab-skhan:
Selecting a new color scale from the color selector is not updating the color legend

PR #1201, opened 2022-09-18
Feature for editing breakpoints

shruthirai:


shadab-skhan:
I see a couple of issues here:
1. If I move the first break point, the map disappears.
image

2. If I remove the breakpoint, the color is still seen on the map.
image
image

3. After modifying the color scale by removing some breakpoint, any other color scale like Rainbow is not applied correctly
Actual:
image
Expected:
image

4. After applying Rainbow from step 3, apply Physics back, the color applied in the map is as per the edited breakpoint, also the breakpoint color is reset to Physics
image

shadab-skhan (approved):

PR #1206, opened 2022-09-22
reset cameraPosition if cameraPosition is null

### LINKED ISSUE #1191: [NGRM] - Missing option for resetting the camera position in DeckGL map component
We want to have full control of the DeckGL map camera position in PPA. Although this became partially possible with `cameraPosition` and `getCameraPosition` props, we still don't have at the moment a way to reset the map to its default position. 

The expected behaviour is that if we pass `null` (or {}) to cameraPosition we should revert back to a centered camera with the default zoom level and rotation (the target is not necessarily the initial one if users changed the window size at some point, so this is likely a computed one).

It would be also good to consider using normalized values for storing the camera position, otherwise different users will end up with different views depending the device size. For PPA, this means if I create a dashboard with maps and I want to share it with someone else, it's likely they will see a different view from mine if their monitor is larger or small than the one I use.

cc: @shadab-skhan

VincentNevermore:
fixes #1191 

Origin with CustomCameraPosition:
image

After set CustomCamraPosition as null:
image

Improvement: 
- [ ]I think we can  update the getCameraPosition after cameraPosition set to null, otherwise it will be a bit confusing

### DIFF
--- a/react/src/lib/components/DeckGLMap/components/Map.tsx
+++ b/react/src/lib/components/DeckGLMap/components/Map.tsx
@@ -302,18 +302,12 @@ const Map: React.FC<MapProps> = ({
     // state for views prop of DeckGL component
     const [viewsProps, setViewsProps] = useState<ViewportType[]>([]);
 
-    // React.useEffect(() => {
-    //     function handleResize() {
-    //         console.log(
-    //             "resized to: ",
-    //             window.innerWidth,
-    //             "x",
-    //             window.innerHeight
-    //         );
-    //     }
-
-    //     window.addEventListener("resize", handleResize);
-    // });
+    const initialViewState = getViewState(
+        boundsInitial,
+        views?.viewports[0].target,
+        views?.viewports[0].zoom,
+        deckRef.current?.deck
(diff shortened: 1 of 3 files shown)


shadab-skhan:
Reset camera works fine but it breaks the change when the bounds prop is optional.
You can verify it by removing the bounds prop from defaultProps in wellsLayer.stories.tsx and check "Wells Layer/Volve wells story".

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -383,9 +387,8 @@ const Map: React.FC<MapProps> = ({
         // If "bounds" or "cameraPosition" is not defined "viewState" will be
         // calculated based on the union of the reported bounding boxes from each layer.
         const isBoundsDefined =
-            typeof bounds !== "undefined" &&
+            typeof bounds !== "undefined" ||
```
shadab-skhan, line 390: Looks like this broke it.

shadab-skhan (approved):

PR #1207, opened 2022-09-26
put propertyname into properties array

### LINKED ISSUE #992: [NGRM] - put property name into the properties array
image
Here on the log, the properties array will show as an array but for each item, the name of the property will always be shown as "Property", Is there a way that we can put the lay id (i.e the "real" property name into this array object so that later when we experienced multiple values, it will be much more easy to track it.

VincentNevermore:
fixes #992 
image

### DIFF
--- a/react/src/lib/components/DeckGLMap/components/Map.tsx
+++ b/react/src/lib/components/DeckGLMap/components/Map.tsx
@@ -562,7 +562,7 @@ const Map: React.FC<MapProps> = ({
                                         .toString() +
                                     " " +
                                     unit;
-                                element.id = item.layer.id
+                                element.id = item.layer.id;
                             }
                         });
                     }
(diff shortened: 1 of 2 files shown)


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -539,6 +539,7 @@ const Map: React.FC<MapProps> = ({
             let unit = " _";
             pickInfos.forEach(
                 (item: {
+                    layer: any;
```
shadab-skhan, line 542: layer: Layer<unkown>

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -585,6 +587,7 @@ const Map: React.FC<MapProps> = ({
                 type: type,
                 infos: infos,
             };
+            console.log(ev);
```
shadab-skhan, line 590: no console

VincentNevermore:
Closed as the request no loner needed

PR #1214, opened 2022-09-28
map3Dlayer should handle "Frame" is not set but "bounds" is set.

nilscb:


shadab-skhan (approved):
It is taking a while to compute the bounds, from this layer. Because of which in the default story, the map loads first by computing the bounds from other layers and then it resets to new bound with this layer.
We may need to try a fix for it in future.

PR #1223, opened 2022-10-05
Will fix: "Automatically set initial rotation point to center of axis…

nilscb:
… bounds in 3D (vertical and lateral) #867"

### DIFF
--- a/react/src/lib/components/DeckGLMap/components/Map.tsx
+++ b/react/src/lib/components/DeckGLMap/components/Map.tsx
@@ -66,6 +66,21 @@ function addBoundingBoxes(b1: BoundingBox, b2: BoundingBox): BoundingBox {
     return [xmin, ymin, zmin, xmax, ymax, zmax];
 }
 
+function boundingBoxeCenter(box: BoundingBox): [number, number, number] {
+    const xmin = box[0];
+    const ymin = box[1];
+    const zmin = box[2];
+
+    const xmax = box[3];
+    const ymax = box[4];
+    const zmax = box[5];
+    return [
+        xmin + 0.5 * (xmax - xmin),
+        ymin + 0.5 * (ymax - ymin),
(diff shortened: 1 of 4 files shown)


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -66,6 +66,21 @@ function addBoundingBoxes(b1: BoundingBox, b2: BoundingBox): BoundingBox {
     return [xmin, ymin, zmin, xmax, ymax, zmax];
 }
 
+function boundingBoxeCenter(box: BoundingBox): [number, number, number] {
```
shadab-skhan, line 69: boundingBoxCenter

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -395,21 +416,38 @@ const Map: React.FC<MapProps> = ({
...
         setReportedBoundingBoxAcc(union_of_reported_bboxes);
         const is3D = views?.viewports?.[0]?.show3D ?? false;
+        const axesLayer = st_layers.find((e) => {
+            return e["@@type"] === "AxesLayer";
+        });
+        const isAxesLayer = typeof axesLayer !== "undefined";
+        // target: camera will look at either center of axes if it exists or center of data ("union_of_reported_bboxes")
+        const target = boundingBoxeCenter(
+            isAxesLayer
+                ? (axesLayer!["bounds"] as BoundingBox)
```
shadab-skhan, line 426: I think we can use axesLayer?.["bounds"] instead of (axesLayer!["bounds"] to get rid of non-null assertion

shadab-skhan (commented):
In the default storybook example the map appears and the disappears. It appears again when zoomed in.
Also, when I switched the map to 3d, the map didn't loaded.

shadab-skhan (approved):

PR #1247, opened 2022-10-14
Prevent recalculating of initial camera when adding a layer.

nilscb:


shadab-skhan:
The PR works fine for the added story, but it fails for the [MultiView story](https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/deckglmap--multi-view
)
I tried with adding "wells-layer" id in views.viewport[0].layerIds prop

VincentNevermore (approved):

Issue #1248, opened 2022-10-16
Map clipped by nearplane in Orthographic(2D)


nilscb:
![MicrosoftTeams-image](https://user-images.githubusercontent.com/136738/196056259-7a8ba491-8ef3-4b4d-b868-62784efcc9bb.png)

This happens because nearplane is set to 0 and this particular surface has both positive and negative depths.

PR #1249, opened 2022-10-16
Will fix issue: "Map clipped by nearplane in Orthographic(2D) #1248"

nilscb:


shadab-skhan (approved):
Just a small observation with North arrow, when zoom out too much, it disappears. It happens with the released version too.

PR #1250, opened 2022-10-17
Added feature to scale in z direction with up/down arrow key

nilscb:
…. Default value 1.

Also added a story totest it.


shadab-skhan:
Can we increase the size by pressing the arrow key like in RMS?
The advantage is, all the maps will have this feature by default similar to zoom.

shadab-skhan (approved):

PR #1254, opened 2022-10-18
Multiple layers

VincentNevermore:


shadab-skhan:
It selects multiple wells without pressing the ctrl key

shadab-skhan (changes requested):
It works fine now, the color-tables package is downgraded, that's not required I guess.

shadab-skhan (approved):

PR #1260, opened 2022-10-21
Wellpick flattening 

Vladimir-Kokin:
- Allow flattening on one or two well picks See #1096
- Support strict mode for Ajv scheme library
- Remove argType.defaultValue


shadab-skhan:
It works fine, but it throws some error
image

shadab-skhan (approved):
It works fine, but looks there is an update issue. In the storybook ,the changes done in wellpickFlatting prop is reflected when I scroll the track, it doesn't updates immediately.,

Issue #1004, opened 2022-05-20
Render discrete legend increasing downwards


hkfb:
Render legend with index 0 at top.

shruthirai:
![image](https://user-images.githubusercontent.com/44650767/198554191-500f6fb7-5f76-46bb-9719-312f9a490ad2.png)

Color scale for discrete data is different in log viewer and in map

shruthirai:
Fixed

PR #1263, opened 2022-10-26
Render discrete legend increasing downwards

shruthirai:


shruthirai:
Fixes https://github.com/equinor/webviz-subsurface-components/issues/1004

shruthirai:
# closing since the solution is different from what is expected

PR #1269, opened 2022-11-02
add a reset option to reset the multiple selected wells

VincentNevermore:


shruthirai (approved):

react/src/lib/components/DeckGLMap/DeckGLMap.tsx
```diff
@@ -72,6 +72,8 @@ export interface DeckGLMapProps {
      */
     getTooltip?: TooltipCallback;
     cameraPosition?: ViewStateType | undefined;
+
+    triggerResetOption?: Record<string, boolean>;
```
shadab-skhan, line 76: The prop type need not to be Record<string, boolean>
We can make it consistent with other triggers like triggerHome.

shadab-skhan (changes requested):

PR #1272, opened 2022-11-04
Opacity

VincentNevermore:
Provide a click function to highlight multiple selected wells in multiple layers. (Note now the highlight color I choose for the wells is Orange, it can be further defined by user and is open to discuss with Equinor people.)
Attached as a screenshot: 
image


VincentNevermore:
linked with issue: https://github.com/equinor/webviz-subsurface-components/issues/986

shruthirai (approved):

PR #1281, opened 2022-11-08
use log sample

shruthirai:


react/src/lib/components/DeckGLMap/layers/wells/wellsLayer.stories.tsx
```diff
@@ -482,16 +482,22 @@ const reverseRange = false;
 //eslint-disable-next-line
 const wellLayerTemplate = (args: any) => {
     const [getColorName, setColorName] = React.useState("Rainbow");
-
+    const [isLog, setLog] = React.useState(false);
```
VincentNevermore, line 485: normally, we just put set as the prefix of the variable name so I guess setIsLog will be more appropriate here to indicate that this is the state action of the vairable isLog

react/src/lib/components/DeckGLMap/layers/colormap/colormapLayer.ts
```diff
@@ -25,7 +25,8 @@ function getImageData(
     colorMapName: string,
     colorTables: colorTablesArray,
     colorMapFunction?: colorMapFunctionType,
-    breakpoint?: number[]
+    breakpoint?: number[],
```
VincentNevermore, line 28: breakpoints

VincentNevermore (approved):
just left two small comments :) and it will good to go then.

Issue #1290, opened 2022-11-15
Phong shading for map layer


hkfb:
Surface smoothing should be default enabled, but optionally disabled.

PR #1293, opened 2022-11-17
Fixes: Phong shading for map layer #1290

nilscb:
MapLayer now has a property "enableSmoothShading" which enabled will estimate normals for each vertex and henceenable phong (smooth) shading. Default true.


react/src/lib/components/DeckGLMap/layers/map/mapLayer.ts
```diff
@@ -32,6 +32,7 @@ export type Params = {
     propertiesData: Float32Array;
     isMesh: boolean;
     frame: Frame;
+    enableSmoothShading: boolean;
```
hkfb, line 35: The `enable` prefix is not consistent with other component prop naming. Suggest `smoothShading` instead.

hkfb (approved):

Issue #1295, opened 2022-11-18
`this.context.viewport.constructor.name` value is not reliable/robust


anders-kiaer:
`this.context.viewport.constructor.name` is _minified_ and is therefore not reliable. 

| `show3D`  | Not minified | Minified (in my case - I guess this could vary) |
| ------------- | ------------- | ----------- |
| `true`  | `OrbitViewport`  | `c` |
| `false`  | `OrthographicViewport`  | `u` |

Current usage of `OrthographicViewport`: https://github.com/equinor/webviz-subsurface-components/search?q=OrthographicViewport

Current usage of `OrbitViewPort`: https://github.com/equinor/webviz-subsurface-components/search?q=OrbitViewPort

anders-kiaer:
Checked a bit further. `this.context.viewport.constructor.name` is _minified_ and is therefore not reliable. 

| `show3D`  | Not minified | Minified (in my case) |
| ------------- | ------------- | ----------- |
| `true`  | `OrbitViewport  | `c` |
| `false`  | `OrthographicViewport`  | `u` |

PR #1299, opened 2022-11-18
Make pie chart work both on minified and non-minified code

anders-kiaer:
- [X] Verify that `debug=False` (minified JS code) goes through, but _not_ `debug=True` (non-minified JS code), in the Python deck.gl test.
- [x] Change pie chart layer such that both tests pass in CI.

Related to https://github.com/equinor/webviz-subsurface-components/issues/1295.

### DIFF
--- a/react/src/lib/components/DeckGLMap/layers/piechart/pieChartLayer.ts
+++ b/react/src/lib/components/DeckGLMap/layers/piechart/pieChartLayer.ts
@@ -70,12 +70,9 @@ export default class PieChartLayer extends CompositeLayer<
             return [];
         }
 
-        const is_orthographic =
-            this.context.viewport.constructor.name === "OrthographicViewport";
-
         const npixels = 100;
-        const p1 = is_orthographic ? [0, 0] : [0, 0, 0];
-        const p2 = is_orthographic ? [npixels, 0] : [npixels, 0, 0];
+        const p1 = [0, 0, 0];
+        const p2 = [npixels, 0, 0];
 
         const v1 = new Vector3(this.context.viewport.unproject(p1));
         const v2 = new Vector3(this.context.viewport.unproject(p2));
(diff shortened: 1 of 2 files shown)


react/src/lib/components/DeckGLMap/layers/piechart/pieChartLayer.ts
```diff
@@ -70,12 +70,9 @@ export default class PieChartLayer extends CompositeLayer<
...
-            this.context.viewport.constructor.name === "OrthographicViewport";
-
         const npixels = 100;
-        const p1 = is_orthographic ? [0, 0] : [0, 0, 0];
-        const p2 = is_orthographic ? [npixels, 0] : [npixels, 0, 0];
+        const p1 = [0, 0, 0];
+        const p2 = [npixels, 0, 0];
 
         const v1 = new Vector3(this.context.viewport.unproject(p1));
         const v2 = new Vector3(this.context.viewport.unproject(p2));
```
nilscb, line 78: This didnt quite work when I tried it. In 3D the distance "d" got calculated wrong and no pies appear. This however works: (must remember to import Vector2 isntead of Vector3)

    renderLayers(): SolidPolygonLayer<PolygonData>[] {
        const pieData = this.props.data as unknown as PiesData;
        if (!pieData?.pies) {
            // this.props.data is a sum type, and since TS doesn't have
            // pattern matching, we must check it this way.
            return [];
        }

        const npixels = 100;
        const p1 = [0, 0];
        const p2 = [npixels, 0];

        const p1_unproj = this.context.viewport.unproject(p1);
        const p2_unproj = this.context.viewport.unproject(p2);

        const v1 = new Vector2(p1_unproj[0], p1_unproj[1]);
        const v2 = new Vector2(p2_unproj[0], p2_unproj[1]);
        const d = v1.distance(v2);

        // Factor to convert a length in pixels to a length in world space.
        const pixels2world = d / npixels;

        const layer = new SolidPolygonLayer<PolygonData>(
            this.getSubLayerProps({
                data: makePies(pieData, pixels2world),
                getFillColor: (d: PolygonData) => d.properties.color,
            })
        );
        return [layer];
    }

anders-kiaer: Thanks for the review @nilscb. Suggestion added.

nilscb (approved):

PR #1300, opened 2022-11-18
boiler template for lasso layer implementation

VincentNevermore:


react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+            "https://raw.githubusercontent.com/equinor/webviz-subsurface-components/master/react/src/demo/example-data/propertyMap.png",
+        wellsData:
+            "https://raw.githubusercontent.com/equinor/webviz-subsurface-components/master/react/src/demo/example-data/volve_wells.json",
+        logData:
+            "https://raw.githubusercontent.com/equinor/webviz-subsurface-components/master/react/src/demo/example-data/volve_logs.json",
+    },
+    bounds: [432205, 6475078, 437720, 6481113],
+    layers: [
+        {
+            "@@type": "ColormapLayer",
```
hkfb, line 70: Should use the new `MapLayer` instead of `ColormapLayer` and `Map3DLayer`.

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.tsx
```diff
@@ -0,0 +1,115 @@
...
+            return [];
+        }
+
+        const LINE = "line";
+        const POINT = "point";
+        const isOrthographic =
+            this.context.viewport.constructor.name === "OrthographicViewport";
+        const positionFormat = isOrthographic ? "XY" : "XYZ";
+
+        const geoJsonLayer = new UnfoldedGeoJsonLayer({
```
hkfb, line 72: The `UnfoldedGeoJsonLayer` is for "unfolded" or "section" views as far as I remember. I think you can use the standard `GeoJsonLayer`.

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+        top: "0",
+        right: "0",
+    },
+});
+
+export const lassoLayerTemplate: ComponentStory<typeof DeckGLMap> = (args) => {
+    const [editedData, setEditedData] = React.useState(args.editedData);
+    const [enableLassoVisible, setEnableLassoVisible] =
+        React.useState<boolean>(false);
+    const handleChange1 = () => {
```
hkfb, line 29: Can you find a better name?

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+                <DeckGLMap
+                    {...args}
+                    editedData={editedData}
+                    setProps={(updatedProps) => {
+                        setEditedData(updatedProps);
+                    }}
+                    enableLassoVisible={enableLassoVisible}
+                />
+            </div>
+            <button onClick={handleChange1}> Enable Lasso Selection </button>
```
hkfb, line 52: I wonder if the `enableLassoVisible` prop for the `DeckGLMap` is necessary. Can we just add the layer when this button is clicked, and remove vice versa?

VincentNevermore: this might not be possible since it means we want to let the user to add the lasso layer by themselves. I think we should add the lassolayer into the resource data but just let the user to have the availability to control to turn it visible or not.

react/src/lib/components/DeckGLMap/DeckGLMap.tsx
```diff
@@ -96,6 +97,7 @@ const DeckGLMap: React.FC<DeckGLMapProps> = ({
     getCameraPosition,
     triggerHome,
     triggerResetMultipleWells,
+    isLassoSelectionAvailable,
```
hkfb, line 100: I think we can toggle lasso selection by changing the visibility state of the layer, then we don't need this prop.

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+                    {...args}
+                    editedData={editedData}
+                    setProps={(updatedProps) => {
+                        setEditedData(updatedProps);
+                    }}
+                    isLassoSelectionAvailable={isLassoSelectionAvailable}
+                />
+            </div>
+            <button onClick={handleChange1}> Enable Lasso Selection </button>
+            <button onClick={handleChange2}> Disable Lasso Selection </button>
```
hkfb, line 53: Since this is a toggle, we could represent it as checkbox or a toggle button instead of two buttons, with the benefit of showing the current state.

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
hkfb: How about `handleLassoSelectionToggle`?

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+            {
+                id: "view_1",
+                show3D: false,
+                layerIds: [],
+            },
+        ],
+    },
+};
+
+const disableLassoArgs = {
```
hkfb, line 134: Please consolidate `enable/disableLassoArgs`. They are almost identical.

Also, please remove props that are not relevant to this example.

Also, please use local resource urls (see other examples).

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+    },
+});
+
+export const lassoLayerTemplate: ComponentStory<typeof DeckGLMap> = (args) => {
+    const [editedData, setEditedData] = React.useState(args.editedData);
+    const [argsState, setArgsState] =
+        React.useState<Record<string, unknown>>(disableLassoArgs);
+    const [alignment, setAlignment] = React.useState("");
+
+    const handleChange1 = React.useCallback(
```
hkfb, line 33: Why is there a "1" suffix?

VincentNevermore: let me remove it...

hkfb (changes requested):
Some suggestions:

- [x] Remove annotations, including the color legend. (This will lead to fewer false positives if we cover this story with automated tests)
- [ ] The lasso layer should be enabled when first opening the story, since it is the purpose of this story to demonstrate the lasso functionality
- [ ] Suggest replacing the icon in the toggle button with a text, eg. "Lasso selection", or using a switch with a label (https://mui.com/material-ui/react-switch/)

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+                    {...argsState}
+                    editedData={editedData}
+                    setProps={(updatedProps) => {
+                        setEditedData(updatedProps);
+                    }}
+                    legend={{ visible: false }}
+                />
+            </div>
+            <div style={{ textAlign: "center" }}>
+                <div style={{ display: "inline" }}>Display lasso layer</div>
```
hkfb, line 65: Suggest using a Label instead, in order to increase the clicking surface: https://mui.com/material-ui/react-switch/#label

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+    },
+    legend: {
+        width: 100,
+        position: "absolute",
+        top: "0",
+        right: "0",
+    },
+});
+
+export const lassoLayerTemplate: ComponentStory<typeof DeckGLMap> = (args) => {
```
hkfb, line 26: Please change the name of the story into something more descriptive, eg. "LassoSelection". This is not really a template.

hkfb (changes requested):

hkfb (changes requested):

react/package.json
```diff
@@ -57,6 +57,7 @@
         "@equinor/eds-icons": "^0.9.1",
         "@equinor/videx-wellog": "^0.7.0",
         "@material-ui/core": "^4.12.3",
+        "@material-ui/icons": "^4.11.3",
```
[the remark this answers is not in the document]
VincentNevermore: no it is not needed. I will delete it then.

react/src/lib/components/DeckGLMap/layers/LassoLayer/lassoLayer.stories.tsx
```diff
@@ -0,0 +1,126 @@
+import { FormControlLabel, makeStyles } from "@material-ui/core";
+import Switch from "@material-ui/core/Switch";
```
hkfb, line 2: You wrote over my change. You should avoid force-pushing.

The second import statement can be combined with the first.

hkfb (changes requested):

hkfb (approved):

Issue #1291, opened 2022-11-15
Map layer does not work without meshUrl


hkfb:
Should show a flat map

PR #1301, opened 2022-11-21
Fixes: Map layer does not work without meshUrl #1291

nilscb:


react/src/lib/components/DeckGLMap/layers/map/mapLayer.ts
```diff
@@ -348,7 +348,7 @@ export default class MapLayer extends CompositeLayer<MapLayerProps<unknown>> {
                 meshLines: this.state["mesh_lines"],
                 pickable: this.props.pickable,
                 modelMatrix: rotatingModelMatrix,
-                contours: this.props.contours,
+                contours: !isMesh ? [-1, -1] : this.props.contours,
```
hkfb, line 351: Maybe we should consider the case where you want contours on a property, eg. thickness, even if there is no mesh. 

@Midtveit do you think it can be useful to have contours on properties?

hkfb (approved):

react/src/lib/components/DeckGLMap/layers/map/mapLayer.ts
nilscb: Yes that should be possible. I merged it already but I'll make a new issue on it

Issue #1297, opened 2022-11-18
Editing or changing legend breakpoint do not effect the map


shruthirai:
![image](https://user-images.githubusercontent.com/44650767/202688954-82930e68-afda-409c-917f-8bb051066df6.png)

shruthirai:
Resolved in PR https://github.com/equinor/webviz-subsurface-components/pull/1368

PR #1302, opened 2022-11-21
Adding nearest sampling

shruthirai:


react/src/lib/components/DeckGLMap/layers/colormap/colormapLayer.ts
```diff
@@ -103,6 +111,9 @@ export interface ColormapLayerProps extends BitmapLayerProps {
 
     // check for logarithmic values
     isLog?: boolean;
+
+    // // check for nearest interpolation
+    isNearest?: boolean;
```
hkfb, line 116: I don't think we should need these settings. The `colorMapFunction` should cover interpolation as well as breakpoints.

hkfb (changes requested):

shruthirai:
Breakpoints will be handled with this task https://github.com/equinor/webviz-subsurface-components/issues/1297

shruthirai:
@hkfb This is ready for review

hkfb (approved):
We should stop using the `ColorMapLayer` and use the `MapLayer` instead for all relevant stories. Please move this story to `react/src/lib/components/DeckGLMap/layers/map/mapLayer.stories.tsx` (in a separate PR if you like)

PR #1311, opened 2022-11-23
Update workflow to remove `node 12` deprecation warnings in CI

anders-kiaer:
In order to remove these:
image


avanti-em:
Hi @anders-kiaer , can you please update "actions/cache@v1" to v3 as well. That action is also deprecating for node12.
And also codecov v3 will have to be approved through the settings.

avanti-em (approved):

Issue #1303, opened 2022-11-22
MapLayer without mesh url should be able to show contours on properties


nilscb:
A flat layer results when mesh url is not set. It should still be possible to show contours on properties.

PR #1316, opened 2022-11-23
Fixes: MapLayer without mesh url should be able to show contours on p…

nilscb:
…roperties #1303


anders-kiaer:
I've done a small change to allowed GitHub actions (that is why the checks above are pending and yellow now). Could you make a dummy commit @nilscb in order to restart the GitHub CI workflow?

PR #1323, opened 2022-11-25
Fix python tests

HansKallekleiv:


anders-kiaer (approved):
LGTM - one small suggestion only, which might be a cheap way of extending test coverage to `debug=True` in Dash apps.

hkfb:
If this fixes the current breakage in the pipeline, then I suggest merging ASAP.

PR #1329, opened 2022-11-29
Fixes issue 1138: [NGRM] - Map View Axes - in 2D, the axes should be …

nilscb:
…static items on the side of the viewport

### DIFF
--- a/react/src/lib/components/DeckGLMap/layers/axes2d/axes2DLayer.stories.tsx
+++ b/react/src/lib/components/DeckGLMap/layers/axes2d/axes2DLayer.stories.tsx
@@ -8,8 +8,8 @@ export default {
 } as ComponentMeta<typeof DeckGLMap>;
 
 const layerProps = {
-    marginH: 8, // Horizontal margin (%)
-    marginV: 5, // Vertical margin (%)
+    marginH: 80, // Horizontal margin (in pixels)
+    marginV: 40, // Vertical margin (in pixels)
 };
 
 const meshMapLayerPng = {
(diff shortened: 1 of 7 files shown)


react/src/lib/components/DeckGLMap/layers/axes2d/axes2DLayer.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+
+export default {
+    component: DeckGLMap,
+    title: "DeckGLMap / Axes2D",
+} as ComponentMeta<typeof DeckGLMap>;
+
+const layerProps = {
+    marginH: 8, // Horizontal margin (%)
+    marginV: 5, // Vertical margin (%)
```
hkfb, line 12: I needed a bigger margin on my setup. Maybe increase `marginV` to 8.

react/src/lib/components/DeckGLMap/layers/axes2d/axes2DLayer.ts
```diff
@@ -0,0 +1,448 @@
...
+        }
+
+        const is_xaxis_label = d.from[1] !== d.to[1];
+        return is_xaxis_label ? "top" : "center";
+    }
+
+    renderLayers(): LayersList {
+        const vpBounds = this.context.viewport.getBounds();
+
+        const mh = this.props.marginH / 100; // to percentage.
```
hkfb, line 99: Margin specified as percentage doesn't seem to work well with resizing. Would it be better to specify in pixels?

hkfb (approved):

PR #1335, opened 2022-12-02
Working version.

nilscb:


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -540,11 +540,35 @@ const Map: React.FC<MapProps> = ({
     };
 
     useEffect(() => {
-        setScaleZ(scaleZ * 1.05);
+        const newScaleZ = scaleZ * 1.05;
```
hkfb, line 543: Suggest extracting common scaling functionality into a common function, eg. taking negative or positive as an argument.

hkfb (changes requested):

hkfb (approved):

PR #1336, opened 2022-12-05
Cypress Map Layer tests

avanti-em:


avanti-em:
Hi @anders-kiaer , can you please approve this third party action : cypress-io/github-action@d79d2d530a66e641eb4a5f227e13bc985c60b964

(I guess you had already approved its older version a long time ago,  but that is not relevant anymore since it is not compatible with cypress 10) So you can remove the older SHA and replace it with this newer one.

PR #1338, opened 2022-12-05
Updated multi view story to use MapLayer instead of ColorMap layer.

hkfb:


shruthirai (approved):
Note: I see warning in console, not sure though if its related to this PR or not. 
**Warning**: Failed prop type: The prop `id` is marked as required in `DeckGLMap`, but its value is `undefined`.

PR #1339, opened 2022-12-06
Multi view annotation

hkfb:


shruthirai (approved):
Note : Just thought of mentioning that, even you might have noticed that the legend and the scale overlaps at the top.

Issue #1327, opened 2022-11-29
Optionally render 3D objects flat in ortographic view


hkfb:
Should be an option to render eg. parts of well trajectories traversing below a 3D topology map when showing in an ortographic viewport.

- [ ] Ideally the same layer instance should render conditionally on the type of viewport, ie. render flat in ortographic and 3D in 3D view.
- [ ] When rendering flat, the layer ordering should be honored.

See also:
- https://deck.gl/docs/api-reference/core/layer#getpolygonoffset
- https://deck.gl/docs/developer-guide/tips-and-tricks

@nilscb @t0oF-azpn @vardit123y

PR #1343, opened 2022-12-06
Optionally render 3D objects flat in ortographic view #1327

nilscb:


hkfb (changes requested):
Would be nice with a story comparing side-by-side the effects on a wells layer, depth tested and not, respectively, on top of a mesh map.

react/src/lib/components/DeckGLMap/layers/map/mapLayer.stories.tsx
```diff
@@ -251,6 +251,66 @@ MapLayer3dPng.parameters = {
     },
 };
 
+const wellsLayerNoDepthTest = {
```
hkfb, line 254: Should we put this story in `DeckGLMap.stories.tsx` since it's a higher level concept?

hkfb (approved):

PR #1369, opened 2022-12-29
fix: fix reservoir scalars color maping

almersawi:
- **Current behavior:** first scalar color (with index: 0) is never used and all colors is shifted by 1 cell.
- There is always undefined color for the last cell because it's index is scalars.length 
- **Reason:** we increase no. of polygons by 1 before we use its color
- **Solution:** use the cell color and increase no. of polygons at the end of the loop


nilscb (approved):
I can not quite see how this changes the behavior as on line 27 pn is post incremented so it should be 0 the first time (pn++ as opposed to ++pn)

PR #1372, opened 2023-01-03
Export ViewFooter and ViewAnnotation components

### LINKED ISSUE #1371: View annotations are not available in Dash
The view annotations (color legend) are not available in Dash due to missing exports.
E:g. this example: https://github.com/equinor/webviz-subsurface-components/blob/master/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx#L219

cannot be used in Dash as the `View` and `ViewFooter` components are not exported from the module.

==> `View` and `ViewFooter` (possible more?)  must be exported from the module so that dash components are generated.

hkfb:
Fixes #1371


hkfb:
Opening a new PR

PR #1373, opened 2023-01-03
Expose View and ViewFooter components

### LINKED ISSUE #1371: View annotations are not available in Dash

hkfb:


VincentNevermore (approved):

hkfb:
Will try to solve the dash component generation in a separate PR @HansKallekleiv

PR #1377, opened 2023-01-04
bug fixing for empty view object

### LINKED ISSUE #1231: DeckGL component fails and stops responding to callbacks if given empty "views"
As part of a dynamically set up page, we are initially not presenting any views. We previously solved this by initializing the DeckGL component `views` property with an empty dict. Upgrading from `0.4.14a5` to `0.4.14rc2` this no longer works.

Setting `views` to empty dict now crashes in the browser with:
> TypeError: Cannot read properties of undefined (reading '0')
>      at Map (Map.tsx?b604:307:9)

The fix for us is to just not set the `views` property at initialization. But this means that any bug in our code that accidentally sets the `views` to an empty dict will break the DeckGL component for the user, until a page refresh.

I am not sure if this should be considered a bug.

Example to reproduce on `0.4.14rc2`:
```python
import dash
import webviz_subsurface_components

layers = [
    {
        "@@type": "GeoJsonLayer",
        "id": "layer1",
        "data": {
            "type": "FeatureCollection",
            "features": [{
                "type": "Feature",
                "geometry": {
                    "type": "GeometryCollection",
                    "geometries": [{"type": "Point", "coordinates": (0, 0)}],
                },
            }],
        },
        "getPointRadius": 1,
    }
]
views = {
    "layout": [1, 1],
    "viewports": [
        {
            "id": "layer1",
            "name": "asdf",
        },
    ],
}

app = dash.Dash(__name__)
app.layout = webviz_subsurface_components.DeckGLMap(
    id="map",
    layers=[],
    bounds=[-1.0, -1.0, 1.0, 1.0],
    views={},  # Commenting out this line fixes the issue
)

@app.callback(
    dash.Output("map", "layers"),
    dash.Output("map", "views"),
    dash.Input("map", "layers"),
)
def test(_):
    return layers, views

app.run_server()
```

VincentNevermore:
Fix the problem caused by empty view object as mentioned here:
image

Fixes #1231

### DIFF
--- a/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
+++ b/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
@@ -286,6 +286,7 @@ ViewObjectInitializedAsEmpty.args = {
     ],
     views: {} as ViewsType,
 };
+
 const wellsLayerNoDepthTest = {
     ...defaultWellsLayer,
     id: "wells-layer-no-depth-test",
--- a/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
+++ b/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
@@ -14,7 +14,7 @@ import {
     ViewFooter,
     View,
 } from "../..";
-import { ViewStateType } from "./components/Map";
+import { ViewStateType, ViewsType } from "./components/Map";
 
 export default {
     component: DeckGLMap,
--- a/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
+++ b/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
@@ -21,6 +21,8 @@ export default {
     title: "DeckGLMap",
 } as ComponentMeta<typeof DeckGLMap>;
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const exampleData = require("../../../demo/example-data/deckgl-map.json");
 const defaultWellsLayer = {
     "@@type": "WellsLayer",
     data: "@@#resources.wellsData",
@@ -267,6 +269,18 @@ MultiViewAnnotation.args = {
 export const ViewObjectInitializedAsEmpty = MultiViewAnnotationTemplate.bind(
     {}
 );
+
+// multiple view with empty View Object
+export const MultiView2 = MultiViewAnnotationTemplate.bind({});
+MultiView2.args = {
+    ...exampleData[0],
+    legend: {
+        visible: false,
+    },
+    id: "test",
+    layers: [...exampleData[0].layers],
+    views: {},
+};
 
 ViewObjectInitializedAsEmpty.args = {
     id: "view_initialized_as_empty",
--- a/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
+++ b/react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
@@ -255,6 +255,10 @@ MultiViewAnnotation.args = {
             {
                 id: "view_1",
                 layerIds: ["hugin"],
+                target: [437500, 6475000],
+                zoom: -5.0,
+                rotationX: 90,
+                rotationOrbit: 0,
             },
             {
                 id: "view_2",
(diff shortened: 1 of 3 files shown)


hkfb:
Please add an issue link

VincentNevermore:
Fix for Bug #1231

react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
```diff
@@ -267,6 +267,26 @@ MultiViewAnnotation.args = {
     },
 };
 
+export const ViewObjectInitializedAsEmpty = MultiViewAnnotationTemplate.bind(
+    {}
+);
+
+ViewObjectInitializedAsEmpty.args = {
+    id: "view_initialized_as_empty",
+    legend: {
```
hkfb, line 276: The `legend` prop is deprecated. Please remove

hkfb (changes requested):

hkfb (changes requested):

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -323,19 +323,27 @@ const Map: React.FC<MapProps> = ({
...
         boundsInitial,
-        views?.viewports[0].target,
-        views?.viewports[0].zoom,
+        isViewsDefined ? views?.viewports[0].target : undefined,
+        isViewsDefined ? views?.viewports[0].zoom : undefined,
         deckRef.current?.deck
     );
 
     // Local help function.
+    function checkIfViewsIsDefined() {
```
hkfb, line 339: Please avoid local functions in render loop, because it can affect render performance. This function can be module scoped if parameterized.

hkfb (changes requested):

hkfb:
@VincentNevermore please click "Resolve conversation" once you believe a comment has been resolved.

react/src/lib/components/DeckGLMap/components/Map.tsx
VincentNevermore: I have move this function outside the React.FC. in the meantime I am thinking if we should also remove the calcDefaultViewStates function since it is also a local helper function ? Correct me if I am understand it in the wrong way....

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -298,6 +298,15 @@ function adjustCameraTarget(
     return vs;
 }
 
+function checkIfViewsIsDefined(views: ViewsType | undefined) {
```
hkfb, line 301: I think this does the same as `lodash.isEmpty()` function. Could we replace it?

hkfb (changes requested):

react/src/lib/components/DeckGLMap/components/Map.tsx
VincentNevermore: tested and it seems work for both empty or undefined cases. Have replaced it in the new push

hkfb (changes requested):
If I set `views` to something valid in the story, then the view does not update. Maybe there is some missing update.

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -323,15 +323,16 @@ const Map: React.FC<MapProps> = ({
...
     const [viewsProps, setViewsProps] = useState<ViewportType[]>([]);
     const [alteredLayers, setAlteredLayers] = useState([{}]);
 
+    const isViewsDefined = !isEmpty(views);
+
     const initialViewState = getViewState(
         boundsInitial,
-        views?.viewports[0].target,
-        views?.viewports[0].zoom,
+        isViewsDefined ? views?.viewports[0].target : undefined,
```
hkfb, line 334: Could we achieve the same result with the '?' operator, eg.

`views?.viewports?.[0].target,`

hkfb (changes requested):

react/src/lib/components/DeckGLMap/components/Map.tsx
VincentNevermore: > If I set `views` to something valid in the story, then the view does not update. Maybe there is some missing update.

just had a test on the master branch, seems like it is a long existing issue that was not caused by this change. Also find out that the zoom is being updated while the problem is about the target parameter that I put in does not update. I think we should resolve this Bug first and create another to track this view issue.
image

VincentNevermore:
> If I set `views` to something valid in the story, then the view does not update. Maybe there is some missing update.

I will create a new Bug Task to track this issue. Following what I investigated, the issue has been existed in the master branch as well and strangly, the example in the jsx file works but not for the examples in tsx one.

hkfb (approved):

PR #1378, opened 2023-01-05
Dash wrapper for SubsurfaceViewer

hkfb:
Allows mapping annotations to views in Dash.

Example:

```
annotation: {
    view_1: <ViewFooter>Hugin</ViewFooter>,
    view_2: <ViewFooter>kH Netmap</ViewFooter>,
}
```

### DIFF
--- a/react/src/lib/components/DeckGLMap/SubsurfaceViewerDashWrapper.stories.tsx
+++ b/react/src/lib/components/DeckGLMap/SubsurfaceViewerDashWrapper.stories.tsx
@@ -1,0 +1,60 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import SubsurfaceViewerDashWrapper from "./SubsurfaceViewerDashWrapper";
+import { ViewFooter } from "../..";
+
+export default {
+    component: SubsurfaceViewerDashWrapper,
+    title: "SubsurfaceViewerDashWrapper",
+} as ComponentMeta<typeof SubsurfaceViewerDashWrapper>;
+
+const mapLayer = {
+    "@@type": "MapLayer",
+    id: "hugin",
+    meshUrl: "hugin_depth_25_m.float32",
+    frame: {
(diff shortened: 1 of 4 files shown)


react/src/lib/components/DeckGLMap/SubsurfaceViewerDashWrapper.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+        count: [291, 229],
+        increment: [25, 25],
+        rotDeg: 0,
+    },
+    propertiesUrl: "kh_netmap_25_m.float32",
+    contours: [0, 100],
+    material: false,
+};
+
+const Template: ComponentStory<typeof SubsurfaceViewerDashWrapper> = (args) => (
```
VincentNevermore, line 26: Maybe we an call it DashWrapperTemplate to make it more precise than Template.

VincentNevermore:
Maybe add a link for this PR to indicate which issue this PR will be solved.

react/src/lib/components/DeckGLMap/SubsurfaceViewerDashWrapper.tsx
```diff
@@ -0,0 +1,278 @@
...
+        getTooltip: getTooltip,
+        cameraPosition: cameraPosition,
+        getCameraPosition: getCameraPosition,
+        triggerHome: triggerHome,
+        triggerResetMultipleWells: triggerResetMultipleWells,
+        children: children,
+    };
+
+    return (
+        <>
```
VincentNevermore, line 133: Think we do not need the <></> tag here as we do not need to wrap any JSX element.

hkfb:
Btw, @HansKallekleiv I will create a new PR for generating the `ViewFooter` Dash component. Not sure what is the best way to do that yet. But this should work with other kinds of annotations, eg. `<div>Name</div>` or `<ColorLegend/>`

VincentNevermore (approved):

Issue #1341, opened 2022-12-06
Axes2D flickering


hkfb:
Axes are flickering when zooming in or out in this story:

https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/deckglmap-axes2d--base

@t0oF-azpn @nilscb @vardit123y

PR #1379, opened 2023-01-06
Fixes: "Axes2D flickering #1341"

nilscb:


react/package.json
```diff
@@ -77,6 +77,7 @@
         "deck.gl": "^8.8.11",
         "deep-equal": "^2.0.5",
         "fast-json-patch": "^3.0.0-1",
+        "gl-matrix": "^3.4.3",
```
hkfb, line 80: Should also push `package-lock.json` so that we get this package when we run `npm ci`

hkfb (changes requested):

react/src/lib/components/DeckGLMap/layers/axes2d/axes2DLayer.ts
```diff
@@ -1,134 +1,256 @@
...
 
-        const screen_from = this.context.viewport.project(d.from);
-        const screen_to = this.context.viewport.project(d.to);
-        const is_labels = d.label !== "X" && d.label !== "Y" && d.label !== "Z"; // labels on axis or XYZ annotations
-        if (is_labels) {
-            if (screen_from[0] < screen_to[0]) {
-                return "start";
-            }
-        }
+        const promise = load("font-atlas.png", ImageLoader, {
```
hkfb, line 124: I don't think this will work in an npm package. I think you should put the image in a lib directory and use an import statement.

hkfb (changes requested):

hkfb (changes requested):
Can we import the font atlas png using webpack instead? Eg. https://webpack.js.org/guides/asset-management/#loading-images

hkfb:
I think we should test this deployed before merging (in order to test the image file packaging). Maybe @t0oF-azpn could help?

react/src/lib/components/DeckGLMap/layers/axes2d/axes2DLayer.stories.tsx
```diff
@@ -8,7 +8,7 @@ export default {
 } as ComponentMeta<typeof DeckGLMap>;
 
 const layerProps = {
-    marginH: 80, // Horizontal margin (in pixels)
+    marginH: 100, // Horizontal margin (in pixels)
```
VincentNevermore, line 11: I think we can revert this change here or is there a specific reason that we want to use 100 instead of 80 ?

VincentNevermore, line 11: I think we should revert it here or is there a specific reason we use 100 ?

VincentNevermore (changes requested):

react/src/lib/components/DeckGLMap/layers/axes2d/axes2DLayer.stories.tsx
nilscb: Horisontal margin needs to be a bit less than the vertical one so I think these are sensible default values,

nilscb:
This is now tested by  @t0oF-azpn and found to be working.

VincentNevermore (approved):

react/package.json
```diff
@@ -70,13 +70,15 @@
         "@vivaxy/png": "^1.3.0",
         "@webviz/core-components": "^0.6.0",
         "ajv": "^7.2.1",
+        "buffer": "^6.0.3",
```
hkfb, line 73: What is this used for? It seems to work fine without it...

hkfb (changes requested):

hkfb (approved):

PR #1385, opened 2023-01-10
Bug fix for bounding undefined

VincentNevermore:
The reason for this bug issue has following two probleams:
1) the bounds is not provided and once the user update the views, the bounds will not be updated and still remained as undefined, In the initiali stage, the default cal function can be called and provide a bound for it but in this case the func will not be called and thus cause the error.
2) we need to update the viewProps immediately rather than wait till the setviewProps  get the viewProps prop updated. I have put a fix for this in this PR and hopefully it address all the issues

### DIFF
--- a/react/src/lib/components/DeckGLMap/components/Map.tsx
+++ b/react/src/lib/components/DeckGLMap/components/Map.tsx
@@ -362,45 +362,25 @@ const Map: React.FC<MapProps> = ({
 
         let tempViewStates: Record<string, ViewStateType> = {};
         const isBoundsDefined = typeof bounds !== "undefined";
-        if (input) {
-            tempViewStates = Object.fromEntries(
-                input.map((item, index) => [
-                    item.id,
-                    isBoundsDefined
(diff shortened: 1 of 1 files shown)


hkfb (changes requested):
This change adds some repeated code, which reduces maintainability and increases change resistance. Please refactor in order to reduce repeated code.

shruthirai (approved):

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -335,10 +335,9 @@ const Map: React.FC<MapProps> = ({
     );
 
     // Local help function.
-    function calcDefaultViewStates() {
+    function calcDefaultViewStates(input?: ViewportType[]) {
```
hkfb, line 338: Does this function need to be local? Can we move it out of the render loop?

hkfb (changes requested):

hkfb (approved):

PR #1387, opened 2023-01-13
Had to update several test on existence of cameraProperty.

nilscb:
Earlier I belive it was always at least an empty object but now it mau be undefined.

### DIFF
--- a/react/src/lib/components/SubsurfaceViewer/components/Map.tsx
+++ b/react/src/lib/components/SubsurfaceViewer/components/Map.tsx
@@ -436,7 +436,7 @@ const Map: React.FC<MapProps> = ({
         if (isCameraPositionDefined) {
             tempViewStates = Object.fromEntries(
                 viewsProps.map((item) => [item.id, cameraPosition])
-            );
+            ) as unknown as Record<string, ViewStateType>;
         } else {
             tempViewStates = Object.fromEntries(
                 viewsProps.map((item, index) => [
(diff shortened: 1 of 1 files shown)


react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -396,10 +396,13 @@ const Map: React.FC<MapProps> = ({
 
     useEffect(() => {
         let tempViewStates: Record<string, ViewStateType> = {};
-        if (cameraPosition && Object.keys(cameraPosition).length !== 0) {
+        const isCameraPositionDefined =
```
VincentNevermore, line 399: Not sure why do we need to introduce this const variable here ? I think the if statement before should be doing the same check ?

react/src/lib/components/DeckGLMap/components/Map.tsx
```diff
@@ -427,10 +430,13 @@ const Map: React.FC<MapProps> = ({
     // calculate view state on deckgl context load (based on viewport size)
     const onLoad = useCallback(() => {
         let tempViewStates: Record<string, ViewStateType> = {};
-        if (cameraPosition && Object.keys(cameraPosition).length !== 0) {
+        const isCameraPositionDefined =
```
VincentNevermore, line 433: feels like here has a lot of duplication code have been introduced(looks exactly as from line 399). Please consider to resolve this.

VincentNevermore (changes requested):

react/src/lib/components/DeckGLMap/components/Map.tsx
nilscb: I placed the variable in the top of the file so that it is calculated only once.

hkfb (approved):

VincentNevermore (approved):

PR #1388, opened 2023-01-16
fix issue for color mapping loading issue when map has been successfully loaded

VincentNevermore:


hkfb (changes requested):
The spinner doesn't seem to work with the default example: https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/deckglmap--default

w1nklr:
Closed due to no activity since January 2023.

PR #1391, opened 2023-01-17
Rename "DeckGLMap" to "SubsurfaceViewer"

### LINKED ISSUE #825: Rename DeckGLMap
Suggest renaming to something less technical, for example "SubsurfaceView".

- [ ] Should keep alias to old DeckGLMap name for backwards compatibility

hkfb:
Fixes #825.


shruthirai:
Note : Inside intersectionView.stories.tsx and DeckGLMapSchema.stories.jsx ---> we have const name "DeckGLMapTemplate", incase if you forgot to rename. There is one more python file named DeckGLMap.py (not sure if this should be renamed as well or not)

hkfb:
> There is one more python file named DeckGLMap.py (not sure if this should be renamed as well or not)

I believe this is generated, so it should automatically be generated with the new name if you delete it.

shruthirai:
Also maybe we can update the warning in these 2 test files(DrawModeSelector_performance.test.tsx and LayerProperty_performance.test.tsx) where they have mentioned the path referring deckgl(/components/DeckGLMap/components/settings/)

shruthirai (approved):

Issue #1389, opened 2023-01-17
Cannot position colorlegend in Dash


HansKallekleiv:
To reproduce:
- Replace `children` in the [python annotation example](https://github.com/equinor/webviz-subsurface-components/blob/master/examples/deckgl_annotation.py) with:

```python
    children=[
        wsc.ViewAnnotation(
            id="view_1",
            children=[
                wsc.ViewFooter(children="Hugin"),
                wsc.WebVizContinuousLegend(
                    "test",
                    min=0,
                    max=10,
                    colorName="seismic",
                    horizontal=True,
                    position=[16, 10], # <- Has no effect
                ),
            ],
        ),
        wsc.ViewAnnotation(id="view_2", children=[wsc.ViewFooter(children="sdfsfd")]),
    ]
```

The legend is locked to the top left. Using the `position` prop does not have any effect.

shruthirai:
This is because currently we do not have position prop in the latest version of color component but instead it can be positioned using prop cssLegendStyles (implemented by Amro).

shruthirai:
No, this prop is not yet used in Webviz.

shruthirai:
> No, this prop is not yet used in Webviz.



> > This is because currently we do not have position prop in the latest version of color component but instead it can be positioned using prop cssLegendStyles (implemented by Amro).
> 
> Is the `cssLegendStyles` exposed from the WebVizContinuousLegend component?

No, this prop is not yet used in Webviz.

PR #1392, opened 2023-01-17
fix: adding missing prop for legend

shruthirai:


shruthirai:
Fixes : https://github.com/equinor/webviz-subsurface-components/issues/1389

react/src/lib/components/ColorLegends/WebVizContinuousLegend.tsx
```diff
@@ -4,10 +4,15 @@ import { ContinuousLegend } from "@emerson-eps/color-tables";
 import { colorTablesArray } from "@emerson-eps/color-tables/";
 
 interface LegendProps {
-    title: string;
+    title?: string;
     min: number;
     max: number;
-    position?: number[] | null;
+    cssLegendStyles?: {
```
hkfb, line 10: I would think a more proper type for CSS would be `Record<string, unknown>` or `React.CSSProperties` (if this works with Dash). There are several more style options than the ones listed.

hkfb (changes requested):

hkfb (changes requested):
Got this error in the example_webviz_continuouslegend.py example:

```
ContinuousLegend.js?07bb:297 TypeError: colorTables.find is not a function
    at eval (ContinuousLegend.js?07bb:121:1)
    at step (ContinuousLegend.js?07bb:43:1)
    at Object.eval [as next] (ContinuousLegend.js?07bb:24:46)
    at fulfilled (ContinuousLegend.js?07bb:15:43)
```

shruthirai:
> Got this error in the example_webviz_continuouslegend.py example:
> 
> ```
> ContinuousLegend.js?07bb:297 TypeError: colorTables.find is not a function
>     at eval (ContinuousLegend.js?07bb:121:1)
>     at step (ContinuousLegend.js?07bb:43:1)
>     at Object.eval [as next] (ContinuousLegend.js?07bb:24:46)
>     at fulfilled (ContinuousLegend.js?07bb:15:43)
> ```

I believe, this might be because either colortable is not available or we get colortable as string from dash (which i had observed before), will fix it.

shruthirai:
@hkfb I pushed the changes but could not test. Hopefully it should fix the colortable error.

hkfb (approved):

PR #1396, opened 2023-01-18
Multi color legend

VincentNevermore:
image


react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
```diff
@@ -340,3 +346,222 @@ DepthTest.parameters = {
...
+            colorMapName: colorName,
+            colorMapRange:
+                colorRange && isAuto == false ? colorRange : [min, max],
+            colorMapFunction: colorMapFunc(),
+        },
+    ];
+    return (
+        <>
+            <div className={useStyles().colorSelector}>
+                <div style={{ marginTop: 50, height: 70 }}>
```
hkfb, line 499: Could we try using the `cssLegendStyles` prop on the ColorLegend instead of nesting in a div?

hkfb (changes requested):

react/src/lib/components/DeckGLMap/DeckGLMap.stories.tsx
VincentNevermore: Had a try on using the View component but unluckily it still does not work. I guess it is because View component can only take one component and if it becomes two, there is some "outside" div has been created by react as I can still see this     
position: relative;
    height: 92vh;
    width: 97vw;

even if I am using the View component as suggested.

hkfb (changes requested):
The log curves are obscured by them map. This example would be more useful with `depthTest` disabled for the log curves. Not sure if this works yet, have you tried?

Also, the view does not cover the available area, any idea why?

image

PR #1397, opened 2023-01-18
Replace mechanism for default values - mapLayer

nilscb:


react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
```diff
@@ -217,6 +215,26 @@ export interface MapLayerProps<D> extends ExtendedLayerProps<D> {
     depthTest: boolean;
 }
 
+const defaultProps = {
+    "@@type": "MapLayer",
+    name: "Map",
+    id: "map3d-layer-float32",
+    pickable: true,
+    visible: true,
+    // Url for the height field.
```
hkfb, line 224: Should consider tsdoc format

hkfb (changes requested):

hkfb (approved):

Issue #1364, opened 2022-12-20
[NGRM] Additional support for in-memory array data as input in Grid3DLayer


LeonidPolukhin:
It would be nice if Grid3DLayer accepts points, polygons and properties as ether Base64 encoded stings or typed JavaScript arrays.

nilscb:
In addition to giving the input data as url's (for points polygons and properties) you may now also specify these  properties as native javascipt arrays.

PR #1401, opened 2023-01-25
Grid3 d layer array input

nilscb:
This pull request is for fixing issue: "[NGRM] Additional support for in-memory array data as input in Grid3DLayer #1364"

It does not change the current behavior only adds to it (by allowing an additional method of setting input data)


t0oF-azpn:
Can we have a similar API for fetching value arrays through whole project ?
For example the meshUrl or propertiesUrl of the Map layer accept both a Url or a base64 encoded string holding data array.

Could that be a common API everywhere expecting an array of value:
- string should handle either plain url or base60 encoded data (as for Map layer)
- array of values, like introduced here

anders-kiaer (approved):
LGTM - given that we get the CI tests to go through 🟢

Issue #1402, opened 2023-01-27
[NGRM] Additional support for in-memory array data as input in MapLayer


nilscb:
This is similar to issue #1364 but for MapLayer.

Support input as native javascript arrays in addition to already existing URL input.

PR #1403, opened 2023-01-27
[NGRM] Additional support for in-memory array data as input in MapLay…

nilscb:
…er #1402

### DIFF
--- a/react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
+++ b/react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
@@ -59,10 +59,7 @@ async function load_mesh_and_properties(
 
     //-- PROPERTIES. --
     let properties: Float32Array;
-    if (propertiesData.constructor === Float32Array) {
-        // Input data is Float32Array.
-        properties = propertiesData;
-    } else if (Array.isArray(propertiesData)) {
+    if (Array.isArray(propertiesData)) {
         // Input data is native javascript array.
         properties = new Float32Array(propertiesData);
     } else {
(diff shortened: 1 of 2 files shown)


react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
```diff
@@ -40,7 +40,7 @@ export type Params = {
 
 async function load_mesh_and_properties(
     meshUrl: string,
-    propertiesUrl: string,
+    propertiesUrl: string | number[],
```
hkfb, line 43: If properties now can be not only URL, is the name "propertiesUrl" still appropriate?

Also, is it possible to use the Float32Array directly in order to enforce 32 bit float from the client directly, which could avoid confusion caused by truncation, say if double precision was used?

hkfb (changes requested):

react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
```diff
@@ -172,7 +184,7 @@ export interface MapLayerProps<D> extends ExtendedLayerProps<D> {
      * each direction then the property values will be pr cell and the cell will be constant
      * colored.
      */
-    propertiesUrl: string;
+    propertiesUrl: string | number[];
```
hkfb, line 187: If properties now can be not only URL, is the name "propertiesUrl" still appropriate?

Also, is it possible to use the Float32Array directly in order to enforce 32 bit float from the client directly, which could avoid confusion caused by truncation, say if double precision was used?

hkfb (approved):

nilscb:
Next step may be to also accept Typed arrays but that must be done in a separate pull request as it turned out not to be straight forward.

PR #1413, opened 2023-02-02
Axes2D - Add non transparent background to the axes and add ability to set color on axes.

nilscb:


react/src/lib/components/SubsurfaceViewer/layers/axes2d/axes2DLayer.ts
```diff
@@ -467,6 +506,36 @@ Axes2DLayer.defaultProps = defaultProps;
...
+            viewMatrix[7],
+            viewMatrix[8],
+            viewMatrix[9],
+            viewMatrix[10],
+            viewMatrix[11],
+            viewMatrix[12],
+            viewMatrix[13],
+            viewMatrix[14],
+            viewMatrix[15]
+        )
```
anders-kiaer, line 533: I'm not 100% sure, but maybe this can be shortened with something like
```suggestion
        mat4.fromValues(...viewMatrix.slice(0, 16))
```

react/src/lib/components/SubsurfaceViewer/layers/axes2d/axes2DLayer.ts
```diff
@@ -467,6 +506,36 @@ Axes2DLayer.defaultProps = defaultProps;
...
+    );
+
+    return [pos_v[0], pos_v[1], pos_v[2]];
```
anders-kiaer, line 536:
```suggestion
    return pos_v.slice(0, 3);
```

nilscb: Unfortunately those suggestions gives problems with typescript. mat4.fromValues for example requires exactly 16 argument.

react/src/lib/components/SubsurfaceViewer/layers/axes2d/axes2DLayer.ts
nilscb: Typescript does not like this unfortunately. mat4.fromValues expects exactly 16 arguments

anders-kiaer (approved):
Before merge @nilscb - will you update the PR title to a more descriptive one? The PR titles are included automatically in suggested squash commit message, and included in the automatic changelog on releases.

anders-kiaer:
@nilscb Regarding TypeScript and the "manual slice" I got an internal suggestion that something like this might work:
```typescript
mat4.fromValues(
    ...(viewMatrix.slice(
        0,
        mat4.fromValues.arguments.length
    ) as Parameters<typeof mat4.fromValues>))
```

hkfb (commented):
Suggest improving the story by exposing controls for color, etc, like this: https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/subsurfaceviewer-axes--custom-label

nilscb:
Typescript liked the suggestion from Anders. But I can not make it run. It seems like the "argument" property is only accessible inside a fiunction and not from outside. I can investigate this further in another pull request maybe.

anders-kiaer:
@nilscb When you try it out: `mat4.fromValues.length` might work instead of `mat4.fromValues.arguments.length`

Issue #1424, opened 2023-02-17
Calculated bounding box from map layer does not honour rotation


HansKallekleiv:
The camera is misaligned due to  incorrect bounds provided from the map layer.

See example below for a surface with a rotation of 250.

The center of the viewport seems to be positioned as if the map is not rotated.

<img width="812" alt="image" src="https://user-images.githubusercontent.com/16436291/219696533-56e8d78a-1b7a-42fc-a455-edd00d1f203d.png">

```python
import io

import numpy as np
import xtgeo
import dash
import webviz_subsurface_components as wsc


ncol, nrow = 1200, 700
xori, yori = 400000, 6800000
xinc, yinc = 25, 25
rotation = 250
values = np.zeros((nrow, ncol))
values = np.transpose(values)

depth_surface = xtgeo.RegularSurface(
    ncol=ncol,
    nrow=nrow,
    xori=xori,
    yori=yori,
    xinc=xinc,
    yinc=yinc,
    rotation=rotation,
    values=values,
)
depth_surface.quickplot("depth.png")
# depth_surface.unrotate()
app = dash.Dash(__name__)

app.layout = dash.html.Div(
    style={
        "position": "relative",
        "border": "solid",
        "height": "800px",
        "width": "60%",
    },
    children=wsc.DashSubsurfaceViewer(
        id="deckgl-map",
        layers=[
            {
                "@@type": "AxesLayer",
                "id": "axes-layer",
                "bounds": [
                    depth_surface.xmin,
                    depth_surface.ymin,
                    -np.nanmax(depth_surface.values),
                    depth_surface.xmax,
                    depth_surface.ymax,
                    np.nanmin(depth_surface.values),
                ],
            },
            {
                "@@type": "MapLayer",
                "id": "mesh-layer",
                "meshData": "/map/mesh",
                "frame": {
                    "origin": [depth_surface.xori, depth_surface.yori],
                    "count": [depth_surface.ncol, depth_surface.nrow],
                    "increment": [depth_surface.xinc, depth_surface.yinc],
                    "rotDeg": depth_surface.rotation,
                },
                "gridLines": False,
                "material": True,
                "colorMapName": "Physics",
                "name": "mesh",
            },
        ],
        views={
            "layout": [1, 1],
            "showLabel": True,
            "viewports": [
                {
                    "id": "view_1",
                    "show3D": False,
                    "name": "Depth surface",
                    "layerIds": ["mesh-layer", "axes-layer"],
[...]

PR #1431, opened 2023-02-22
Fix of: "Calculated bounding box from map layer does not honour rotat…

nilscb:
…ion #1424"

### DIFF
--- a/react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
+++ b/react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
@@ -7,6 +7,19 @@ import * as png from "@vivaxy/png";
 import { makeFullMesh } from "./webworker";
 import { Matrix4 } from "math.gl";
 
+// Rotate x,y around x0, y0 rad radians
+function rotate(
+    x: number,
+    y: number,
+    x0: number,
+    y0: number,
+    rad: number
+): [number, number] {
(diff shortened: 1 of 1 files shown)


react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.ts
```diff
@@ -342,14 +355,32 @@ export default class MapLayer extends CompositeLayer<MapLayerProps<unknown>> {
...
+                    const rotRad = (rotDeg * (2.0 * Math.PI)) / 360.0;
+
+                    // Rotate x,y around "center" "rad" radians
+                    const [x0, y0] = rotate(xMin, yMin, center[0], center[1], rotRad); // eslint-disable-line
+                    const [x1, y1] = rotate(xMax, yMin, center[0], center[1], rotRad); // eslint-disable-line
+                    const [x2, y2] = rotate(xMax, yMax, center[0], center[1], rotRad); // eslint-disable-line
+                    const [x3, y3] = rotate(xMin, yMax, center[0], center[1], rotRad); // eslint-disable-line
+
+                    // Rotated bounds in x/y plane.
+                    const x_min = Math.min(x0, x1, x2, x3);
```
hkfb, line 373: Not sure why you're switching between camel and kebab case :D I think the convention is camel

hkfb (approved):

Issue #1436, opened 2023-02-24
[NGRM] - Axes2D. Add missing properties to set background and text color


nilscb:
It should be possible to control both of these individually.

PR #1437, opened 2023-02-24
Fix of: "[NGRM] - Axes2D. Add missing properties to set background an…

nilscb:
…d text color #1436"


react/src/lib/components/SubsurfaceViewer/layers/axes2d/axes2DLayer.ts
```diff
@@ -58,6 +58,7 @@ export interface Axes2DLayerProps<D> extends ExtendedLayerProps<D> {
     labelFontSize?: number;
     fontFamily?: string;
     axisColor?: Color;
+    backGroundColor?: Color;
```
hkfb, line 61: Suggest `backgroundColor`

hkfb (changes requested):

react/src/lib/components/SubsurfaceViewer/layers/axes2d/axes2DLayer.ts
nilscb: Fixed. And also made the text sharper and more readable.

hkfb:
The text looks more blurry to me in the diff?

nilscb:
The png look more blurry because it is smoothed woth a gaussian blur. But the text on the axes should be sharper.

hkfb (approved):

Issue #1423, opened 2023-02-13
[NGRM] Axes 2D - When Axes 2D is active "bounds" should be adjusted to match area inside axes.


nilscb:
As it is currently the 2D Axes will hide/crop part of the the scene. 

The camera should be adjusted so that the area specified by "bounds" will appear totally inside the axes area.

PR #1467, opened 2023-04-05
Fixes:"[NGRM] Axes 2D - When Axes 2D is active "bounds" should be adj…

nilscb:
…usted to match area inside axes. #1423"

### DIFF
--- a/react/src/lib/components/SubsurfaceViewer/components/Map.tsx
+++ b/react/src/lib/components/SubsurfaceViewer/components/Map.tsx
@@ -29,7 +29,12 @@ import {
 import { LayerPickInfo } from "../layers/utils/layerTools";
 import { getLayersByType } from "../layers/utils/layerTools";
 import { getWellLayerByTypeAndSelectedWells } from "../layers/utils/layerTools";
-import { WellsLayer, AxesLayer, Axes2DLayer, NorthArrow3DLayer } from "../layers";
+import {
+    WellsLayer,
+    AxesLayer,
+    Axes2DLayer,
+    NorthArrow3DLayer,
+} from "../layers";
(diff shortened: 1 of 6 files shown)


react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.stories.tsx
```diff
@@ -530,12 +530,20 @@ export const MapLayer2d: ComponentStory<typeof SubsurfaceViewer> = (args) => {
     return <SubsurfaceViewer {...args} />;
 };
 
+const axesLayer2D = {
+    "@@type": "Axes2DLayer",
```
hkfb, line 534: Would encourage to use `new Axes2DLayer` directly to improve type safety

react/src/lib/components/SubsurfaceViewer/components/Map.tsx
```diff
@@ -328,15 +333,25 @@ const Map: React.FC<MapProps> = ({
     const [viewsProps, setViewsProps] = useState<ViewportType[]>([]);
     const [alteredLayers, setAlteredLayers] = useState<LayersList>([]);
 
+    // Viewport margins. Will differ from 0 if axes2Dlayer is used.
+    const [marginLeft, setMarginLeft] = useState<number>(0);
+    const [marginBottom, setMarginBottom] = useState<number>(0);
+
     const initialViewState = getViewState(
```
hkfb, line 340: Unrelated, but I think this could benefit from being memoized, ie. `const initialViewState = useMemo(() => {...}, [...])`

hkfb (approved):

Issue #1479, opened 2023-04-25
Initial rotation center should be in the middle of data


nilscb:
The initial rotation center is now set to the middle. 
A bounding box is calculated for each "item" in the view. For instance a bounding box for a map and a bounding box for a well. The middle is then set to the mean value of all the bounding box's middles. This also include the 3D axes box.

PR #1480, opened 2023-04-25
Rotate initially around center of data

nilscb:
This fixes "Initial rotation center should be in the middle of data #1479"


react/src/lib/components/SubsurfaceViewer/components/Map.tsx
```diff
@@ -352,6 +351,7 @@ const Map: React.FC<MapProps> = ({
     const initialViewState = getViewState(
```
hkfb, line 351: Can we memoize this call?

hkfb (approved):

PR #1484, opened 2023-04-26
Migrate SubsurfaceViewer to @mui5

hkfb:


react/package.json
```diff
@@ -179,6 +181,10 @@
         "webpack-conditional-loader": "^1.0.12",
         "webpack-dev-server": "^4.11.0"
     },
+    "peerDependencies": {
+        "react": "^17 || ^18",
```
t0oF-azpn, line 185: Support of react 18 will only be valid once there is no dependency on @material-ui any more :/

react/src/lib/components/SubsurfaceViewer/components/ViewFooter.tsx
```diff
@@ -1,23 +1,19 @@
...
+import { styled } from "@mui/system";
 
-const useStyles = makeStyles((theme: Theme) =>
-    createStyles({
-        viewFooter: {
-            position: "absolute",
-            bottom: theme.spacing(0),
-            right: theme.spacing(2),
-            zIndex: 999999,
+const StyledViewFooter = styled("div")(({ theme }) => ({
```
t0oF-azpn, line 5: Can't you simple call that ViewFooter, whitout the need of the wrapper on line 22 ?

react/package.json
t0oF-azpn: I'm not good enough to answer :/
But I tried that out:
- I could update your branch to react 18 without problem (I was surprised)
- `npm ls react` fails because of missing dependencies to react 17
- running the storybook fails to load all the subsurface and well log stories :(

I double checked, and @material-ui does not allow react 18 in its peer dependencies.

react/src/lib/components/SubsurfaceViewer/components/ViewFooter.tsx
hkfb: Not sure how to do that while still capturing the `theme`.

hkfb:
Opened a new PR: https://github.com/equinor/webviz-subsurface-components/pull/1491

PR #1490, opened 2023-05-03
Hook dependencies

nilscb:


react/src/lib/components/SubsurfaceViewer/layers/BoxSelectionLayer/boxSelectionLayer.stories.tsx
```diff
@@ -122,7 +125,9 @@ const enableLassoArgs = {
 export const boxSelectionWithCallback: ComponentStory<
     typeof SubsurfaceViewer
 > = () => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
hkfb, line 128: Already disabled at file level. We don't need both?

hkfb (approved):

PR #1499, opened 2023-05-09
If input data are given as arrays they should be copied.

nilscb:
This is because the data may be changed by the layer (multiply --1 on z values) and changes should not be reflected on the outside of the layer.


anders-kiaer:
We are testing currently on some large, real grids, and we see that we might be stretching frontend/client resources with current `main`. 

I'm a bit afraid that the `deepclone` would increase rendering time + memory consumption in the browser (~doubling the memory footprint of the grid data probably?). 

Should we discuss alternatives to the `deepclone` approach? I guess one alternative would be that we just settle on a sign the backend should send, but there might be other approaches which also doesn't degrade frontend experience? @nilscb @hkfb

hkfb (approved):

PR #1500, opened 2023-05-10
feat: lithology track with pattern

karbor:


karbor:
> 

As far as I could see from the storybook and understand from the code, patterns are only possible for wellpicks, i.e. they will cover the entire area across all tracks for a well. What we need is a track with that possibility - i.e. pattern(and later image), color and labels in a single track. If I'm wrong and this is possible in the existing code, could you provide an example?

hkfb:
> > 
> 
> As far as I could see from the storybook and understand from the code, patterns are only possible for wellpicks, i.e. they will cover the entire area across all tracks for a well. What we need is a track with that possibility - i.e. pattern(and later image), color and labels in a single track. If I'm wrong and this is possible in the existing code, could you provide an example?

I see your point, and I agree it looks like your use case is not supported.

Is the only thing missing a mapping from discrete code / value to a pattern or image?

If so would it make sense to add to individual plots or tracks the pattern mapping shown in this example? https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/welllogviewer-demo-synclogviewer--default

@Midtveit @Vladimir-Kokin

karbor:
> 

I considered reusing the patternsTable/patterns from the existing interface also. Ended up with this as a proposal/start of maybe a more intuitive interface? I just started (with both webviz and front-end programming) so I don't know it this is a good proposal for the interface. But we have seen working from the python-side with the component for a while, that there could be beneficial to look into if the interface to the component could be more intuitive and flexible. Maybe we could have more general discussion around that?

hkfb:
> > 
> 
> I considered reusing the patternsTable/patterns from the existing interface also. Ended up with this as a proposal/start of maybe a more intuitive interface(and just keep it on the side of the existing to start with)? I am new to both webviz and front-end programming so I don't know it this is a good proposal for the interface. But we have seen working from the python-side with the component for a while, that there could be beneficial to look into if the interface to the component could be more intuitive and flexible. Maybe we could have more general discussion around that (in a meeting)?

Would be good to have a discussion on the API design. I think there is already planned a discussion on the WellLogViewer around the end of the month @anders-kiaer although I'm on leave the next couple of weeks, so I probably won't attend.

PR #1502, opened 2023-05-11
[VITE] Moved react/react-dom to peerDependencies

rubenthoms:


react/package.json
```diff
@@ -189,5 +189,9 @@
     "engines": {
         "node": ">=14.17",
         "npm": ">=6.14"
+    },
```
hkfb, line 192: This is already done. I think we should aim for `"react": ^17 || ^18` but we should wait until we have completed migration to MUI 5.

hkfb (changes requested):

react/package.json
rubenthoms: @hkfb Can we close this PR then?

hkfb: Yes.

t0oF-azpn: Yes.

hkfb:
Closing this since it's already fixed.

PR #1504, opened 2023-05-11
[VITE] Moved all shader code to ts files

rubenthoms:


anders-kiaer (approved):
The PR LGTM (after the tests go through). Not sure why the CI tests fail with messages like `ERROR: Was not able to compile the 35633 shader` in browser console. Some subtle difference I don't see between how the shader code snippets are imported in this PR and before it?

t0oF-azpn (approved):

rubenthoms:
> The PR LGTM (after the tests go through). Not sure why the CI tests fail with messages like `ERROR: Was not able to compile the 35633 shader` in browser console. Some subtle side-effect I don't see between how the shader code snippets are imported in this PR and before it?

It turned out that stringified GLSL code needs to have a line break at the end. I did not find anything official on the matter, pattern-recognition, suddenly passing tests, and a SO comment that kind of confirmed the requirement of my changes.

https://stackoverflow.com/a/20816244

PR #1505, opened 2023-05-11
[VITE] Replaced requires with imports

rubenthoms:


react/src/lib/components/SubsurfaceViewer/components/settings/LayersButton.test.tsx
```diff
@@ -8,8 +8,8 @@ import userEvent from "@testing-library/user-event";
 import { EmptyWrapper } from "../../test/TestWrapper";
 import LayersButton from "./LayersButton";
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const exampleData = require("../../../../../demo/example-data/deckgl-map.json");
+// @rmt: Changed require to import
```
t0oF-azpn, line 11: remove comment

react/src/lib/components/WellCompletions/components/WellCompletionComponent.tsx
```diff
@@ -6,8 +6,8 @@ import ErrorPlaceholder from "./Common/ErrorPlaceholder";
 import DataProvider from "./DataLoader";
 import WellCompletionsViewer from "./WellCompletionsViewer";
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const inputSchema = require("../../../inputSchema/wellCompletions.json");
+// @rmt: Changed require to import
```
t0oF-azpn, line 9: Remove comment

react/src/lib/inputSchema/validator.tsx
```diff
@@ -3,14 +3,14 @@ import { ErrorObject, ValidateFunction } from "ajv/dist/types/index";
 
 // schema definations
 /* eslint-disable @typescript-eslint/no-var-requires */
```
t0oF-azpn, line 5: Still needed ?

react/src/lib/components/SubsurfaceViewer/SubsurfaceViewer.tsx
```diff
@@ -284,7 +285,7 @@ SubsurfaceViewer.propTypes = {
      * Parameters for the Distance Scale component
      * Unit for the scale ruler
      */
-    coordinateUnit: PropTypes.string,
+    coordinateUnit: PropTypes.any,
```
t0oF-azpn, line 288: What is the difference with `PropTypes.objectOf(PropTypes.any)` ?

rubenthoms: We have to change this back to `string` and then pretend to have gotten a valid value. The valid values are only defined in TS and can, hence, not be converted to `propTypes` (other way around would work). Writing our own list of valid values would not be sustainable. `dash-generate-components` does, however, support parsing TS files now, so it should be able to extract the types and transfer them to Python. However, I am not sure if it prioritizes `propTypes` or `TS types` when both are available in a file. We can keep this for later.

rubenthoms:
```typescript
convert().possibilities()
```
It is actually possible to get a list of all valid units. We can use `PropTypes.oneOf(...)` to allow only these units at runtime. However, `dash-generate-components` cannot execute functions when parsing component code. The type is automatically casted to `unknown`. I think we can live with this. `PropTypes` should make sure that we always get valid units. In addition, I added a check to `SubsurfaceViewer` - if the unit is invalid, it will log an error and set the unit to undefined.

react/src/lib/components/SubsurfaceViewer/components/DistanceScale.tsx
```diff
@@ -28,7 +27,7 @@ const DistanceScale: React.FC<ScaleProps> = ({
     style,
     scaleUnit,
 }: ScaleProps) => {
-    if (!zoom || !widthPerUnit || !incrementValue) return null;
+    if (!zoom || !widthPerUnit || !incrementValue || !scaleUnit) return null;
```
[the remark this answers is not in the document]
rubenthoms: @t0oF I changed the whole component. The current design with disabling important React errors is not really good. I just removed the state and the useEffect and added a check for the unit.

react/src/lib/components/SubsurfaceViewer/SubsurfaceViewer.tsx
rubenthoms: I removed the custom warning as `PropTypes` is already logging a warning to the console.

t0oF-azpn (approved):

anders-kiaer (approved):

PR #1506, opened 2023-05-11
[VITE] Removed conditional loaders

rubenthoms:


t0oF-azpn (commented):
Why is package-lock.json impacted ? Or is package.json missing ?

t0oF-azpn (approved):

anders-kiaer (approved):

hkfb:
I'm puzzled by this one. I also tried to remove the dependency, with the same result. Does the wsc pip package contain an package.json file which is used by accident?

PR #1507, opened 2023-05-11
[VITE] Removed inline-loaders and adjusted webpack config

rubenthoms:


rubenthoms:
This PR depends on #1504 (since there is no loader for glsl files anymore).

t0oF-azpn (approved):
Seems good for me.
Ideally there could be a section in the README to list all the needed loaders (or a reference to the corresponding config file).

anders-kiaer (approved):

PR #1516, opened 2023-05-16
Update older components to use `d3` version 7

anders-kiaer:
We have updated `d3` version in this repository without moving some of the older components to the new syntax. This fixes that.

Closes https://github.com/equinor/webviz-subsurface/issues/1219.


HansKallekleiv (approved):
👍

PR #1521, opened 2023-05-22
Will fix: "https://github.com/equinor/webviz-subsurface-components/is…

nilscb:
…sues/1513"

### DIFF
--- a/react/src/lib/components/SubsurfaceViewer/components/Map.tsx
+++ b/react/src/lib/components/SubsurfaceViewer/components/Map.tsx
@@ -429,6 +429,8 @@ const Map: React.FC<MapProps> = ({
                 viewsProps.map((item) => [item.id, cameraPosition])
             ) as unknown as Record<string, ViewStateType>;
         } else {
+            const isBoundsDefined = typeof bounds !== "undefined";
+
             tempViewStates = Object.fromEntries(
                 viewsProps.map((item, index) => {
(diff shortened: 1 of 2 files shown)


react/src/lib/components/SubsurfaceViewer/components/Map.tsx
```diff
@@ -429,20 +429,29 @@ const Map: React.FC<MapProps> = ({
...
+
             tempViewStates = Object.fromEntries(
                 viewsProps.map((item, index) => {
                     const viewState = viewStates[item.id];
                     return [
                         item.id,
                         typeof viewState !== "undefined"
                             ? viewState
-                            : getViewState(
+                            : isBoundsDefined
```
hkfb, line 441: Nested ternary operators are hard to read.

https://eslint.org/docs/latest/rules/no-nested-ternary

Suggest adding the `no-nested-ternary` condition to the eslint config.

hkfb (changes requested):

react/src/lib/components/SubsurfaceViewer/components/Map.tsx
nilscb: Suggest adding lint rule "no-nested-ternary" is separate PR as it requires multiple unrelated files to be rewritten.

hkfb (approved):

PR #1523, opened 2023-05-22
Fix axes2 d

nilscb:


react/src/lib/components/SubsurfaceViewer/layers/axes2d/axes2DLayer.ts
```diff
@@ -171,7 +171,8 @@ export default class Axes2DLayer extends Layer<Axes2DLayerProps<unknown>> {
         min: number,
         max: number,
         viewSide: ViewSide,
-        pixel2world: number
+        pixel2world: number,
+        zoom_scale: number
```
hkfb, line 175: Please check variable case consistency. I believe camel case is the most common style in this project. Perhaps there is an eslint rule that we should enable?

hkfb (changes requested):

hkfb (approved):

PR #1531, opened 2023-06-01
Add callback which will be called after all layers are loaded.

nilscb:


nilscb:
This pull request introduces a callBack property which will be triggered when data in all layers are loaded.

hkfb:
Suggest rebasing on master.

typescript/packages/subsurface-viewer/src/storybook/SubsurfaceViewer.stories.jsx
```diff
@@ -644,3 +681,55 @@ ViewMatrixMargin.args = {
...
+                <SubsurfaceViewer {...props} />
+            </div>
+            <button onClick={handleChange}> Add layer </button>
+        </Root>
+    );
+};
+
+IsLoadedCallback.args = {
+    id: "DeckGL-Map",
+    isLoadedCallback: (isLoaded) => console.log("isLoadedCallback", isLoaded),
```
hkfb, line 713: Should be more visible, like eg. a label in the map

typescript/packages/subsurface-viewer/src/storybook/SubsurfaceViewer.stories.jsx
```diff
@@ -644,3 +681,55 @@ ViewMatrixMargin.args = {
         ],
     },
 };
+
+export const IsLoadedCallback = (args) => {
```
hkfb, line 685: Please move to `SubsurfaceViewer.stories.tsx`. Let's try to deprecate the JSX stories.

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/storybook/SubsurfaceViewer.stories.jsx
```diff
@@ -326,6 +324,34 @@ const meshMapLayerPng = {
     colorMapName: "Physics",
 };
 
+const meshMapLayerBig = {
```
hkfb, line 327: Where is this used?

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx
```diff
@@ -16,12 +16,32 @@ import { ExtendedLayerProps, LayerPickInfo } from "./layers/utils/layerTools";
...
 import { FeatureCollection } from "geojson";
 import { ViewFooter } from "./components/ViewFooter";
+import { styled } from "@mui/material/styles";
 
 export default {
     component: SubsurfaceViewer,
     title: "SubsurfaceViewer",
 } as ComponentMeta<typeof SubsurfaceViewer>;
 
+const PREFIX = "Default";
```
hkfb, line 26: This looks like boilerplate `codemod` autogenerated code copied from elsewhere. I think there's room for improvement. At least I don't think the `PREFIX` is needed.

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/storybook/SubsurfaceViewer.stories.jsx
nilscb: It is used int the story?! The point is to have a big map which takes some tine to load

hkfb: It's not used in any of those stories. You moved it to the .tsx file.

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx
nilscb: you're right my mistake :-)

typescript/packages/subsurface-viewer/src/storybook/SubsurfaceViewer.stories.jsx
nilscb: Your're right my mistake :-)

hkfb (changes requested):
I get some typescript errors:
```
packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx:485:34 - error TS7006: Parameter 'args' implicitly has an 'any' type.

485 export const IsLoadedCallback = (args) => {                 
                                     ~~~~                                                 
                                                                                          
packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx:491:24 - error TS2740: Type 'MapLayer' is missing the following properties from type 'WellsLayer': setSelection, setMultiSelection, getLegendData, setLegend, and 4 more.
                                                                                          
491             setLayers([meshMapLayerBig, defaultWellsLayer]);                                                                                                                     
                           ~~~~~~~~~~~~~~~

packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx:499:28 - error TS7006: Parameter 'isLoaded' implicitly has an 'any' type.
                                                                                          
499         isLoadedCallback: (isLoaded) => {                                                                                                                                        
                               ~~~~~~~~                                                   
```

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx
```diff
@@ -445,3 +481,61 @@ ViewStateSynchronization.argTypes = {
...
+        },
+        layers,
+    };
+
+    return (
+        <Root>
+            <div className={classes.main}>
+                <SubsurfaceViewer {...props} />
+            </div>
+            <button onClick={handleChange}> Add layer </button>
```
hkfb, line 512: A bit confusing that this button both adds and removes a layer when it always says "Add layer".

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx
```diff
@@ -445,3 +481,64 @@ ViewStateSynchronization.argTypes = {
...
+        },
+        layers,
+    };
+
+    return (
+        <Root>
+            <div className={classes.main}>
+                <SubsurfaceViewer {...props} />
+            </div>
+            <button onClick={handleChange}> Add/remove layer </button>
```
hkfb, line 515: Maybe you could use a switch or toggle instead, saying "Show map layer"?

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx
nilscb: I am not quite sure how to make a checkbox but I changed the text dynamically to say either "Add layer" or "Remove layer" depending.

hkfb (approved):

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.stories.tsx
nilscb: Used mui switch instead

Issue #1532, opened 2023-06-04
Add possibility to add space around views in matrix mode.


nilscb:
When showing multiple views in a matrix there shpuld be an option to add some space around each.

Added new property on SubSurfaceViewer "marginPixels" which will add this space specified in pixels.

In AHA this corresponds to issue "50330 Webviz: Add option to have a spacing around the entire component"

PR #1533, opened 2023-06-04
Add possibility to add space around views in matrix mode. #1532

nilscb:


react/src/lib/components/SubsurfaceViewer/components/Map.tsx
```diff
@@ -257,6 +257,12 @@ export interface MapProps {
 
     getTooltip?: TooltipCallback;
     cameraPosition?: ViewStateType;
+
+    /**
+     * Number of pixels used for the margin in matrix mode.
+     * Defaults to 0.
+     */
+    marginPixels?: number;
```
hkfb, line 265: Suggest nesting this prop in the `views` prop.

react/src/lib/components/SubsurfaceViewer/components/Map.tsx
```diff
@@ -362,8 +369,9 @@ const Map: React.FC<MapProps> = ({
         viewPortMargins,
         boundsInitial,
         boundingBoxCenter(reportedBoundingBoxAcc),
-        views?.viewports?.[0].target,
-        views?.viewports?.[0].zoom,
+        views?.viewports?.[0],
+        views?.layout,
+        marginPixels ?? 0,
```
hkfb, line 374: You can set the default value in the `Map` or `SubsurfaceViewer` initialization.

react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.stories.tsx
```diff
@@ -1263,6 +1263,7 @@ const ContourLinesStory = (props: {
     contourOffset: number;
     zContourInterval: number;
     propertyContourInterval: number;
+    marginPixels: number;
```
hkfb, line 1266: Suggest creating a new story to demonstrate the margins. If there are regressions specific to margins in the future it will be isolated to that story.

Would also be good to look into improving cypress test coverage. We have some tests already that you could look into. Or start using this storybook addon: https://storybook.js.org/docs/react/writing-tests/visual-testing

hkfb (changes requested):
I think more ideal would be to have the views position calculated with a callback. Ie. allowing `views` as a callback with the default value as the current and then implementing new functions for more advanced layouts. I'm sure there will be more layout use cases in the future, and this will allow custom layouts by consumers. Would also be easier to maintain I think.

react/src/lib/components/SubsurfaceViewer/components/Map.tsx
nilscb: I nested the "marginPixels" inside views as suggested.
However the "views as a callback" is probably better done in a separate PR as it is more of an architectural change.

react/src/lib/components/SubsurfaceViewer/layers/map/mapLayer.stories.tsx
nilscb: Added a story at the top level ("Matrix")
Also the cyprus test hopefully can go in a separate PR as I need some time to figure out how cypress tests are made and run.

react/src/lib/components/SubsurfaceViewer/storybook/SubsurfaceViewer.stories.jsx
```diff
@@ -306,6 +305,37 @@ const meshMapLayer = {
     colorMapName: "Physics",
 };
 
+const meshMapLayerPng = {
+    "@@type": "MapLayer",
```
hkfb, line 309: I think we should only use this method when we specifically want to demonstrate de-serialization from Json. Suggest using `new` instantiation instead.

react/src/lib/components/SubsurfaceViewer/storybook/SubsurfaceViewer.stories.jsx
```diff
@@ -574,3 +604,40 @@ export const MapInContainer = (args) => {
 MapInContainer.args = {
     ...exampleData[0],
 };
+
+export const Matrix = EditDataTemplate.bind({});
+Matrix.args = {
```
hkfb, line 609: Suggest a more descriptive name, eg. "ViewMatrix".

react/src/lib/components/SubsurfaceViewer/components/Map.tsx
hkfb: APIs have a tendency to cement quickly, so better to do it early, but I leave it for you to decide.

react/src/lib/components/SubsurfaceViewer/storybook/SubsurfaceViewer.stories.jsx
nilscb: Used "new" instead of serializtaion and chenaged the storys name to "ViewMatrix"
I think it is a very good idea to rewrite how the view are set up using a callback (or extern function) as you mentioned. For now I would prefer to get this PR in first and then possibly start a rewrite (which may be some work i suspect).

hkfb: I think it's confusing that we have another story named "Multi View", which btw. doesn't work properly. Perhaps we should remove it or give this story a more specific name, like "ViewMatrixMargin"?

react/src/lib/components/SubsurfaceViewer/layers/wells/wellsLayer.stories.tsx
```diff
@@ -414,6 +414,15 @@ AllWellHeadsHidden.parameters = {
 export const Wells3d = Template.bind({});
 Wells3d.args = {
     ...defaultProps,
+    layers: [
+        {
+            ...defaultProps.layers[0],
+            //lineStyle: { dash: true },
```
hkfb, line 420: Suggest to remove this comment.

hkfb (approved):

PR #1540, opened 2023-06-07
Closed polylines

LeonidPolukhin:
PolylinesLayer handles closed polylines and the correspnding API added.


react/src/lib/components/SubsurfaceViewer/layers/polylines/polylinesLayer.ts
```diff
@@ -8,6 +8,9 @@ import {
     defineBoundingBox,
 } from "../utils/layerTools";
 
+type TIsPolylineClosedFunc = (index: number) => boolean;
```
hkfb, line 11: Not a huge fan of the "T" prefix. I don't think it helps readability and you can usually see that it's a type in the editor.

react/src/lib/components/SubsurfaceViewer/layers/polylines/polylinesLayer.ts
```diff
@@ -8,6 +8,9 @@ import {
...
+type TPathType = "open" | "loop" | null;
```
hkfb, line 12: Do we need the `null` option?

hkfb (changes requested):

react/src/lib/components/SubsurfaceViewer/layers/polylines/polylinesLayer.ts
LeonidPolukhin: This is Deck.GL PathLayer prop. If all the polylines are open or closed it can be set in PathLayer and it  will skip some checks for better performance.
If this prop is not set or null PathLayer checks closeness by comparing coordinates and decides how to draw the polylines.

react/src/lib/components/SubsurfaceViewer/layers/polylines/polylinesLayer.ts
```diff
@@ -68,6 +75,7 @@ interface IDataAttributes {
             size: number;
         };
     };
+    pathType: TPathType;
```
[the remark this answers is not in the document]
LeonidPolukhin: Added some explanations, see above

react/src/lib/components/SubsurfaceViewer/layers/polylines/polylinesLayer.ts
LeonidPolukhin: T prefixes deleted

hkfb (approved):

PR #1542, opened 2023-06-07
Side-independent lighting of surfaces in triangle layer

LeonidPolukhin:
Lighting algorytm in triangles layer  fragment shader modified to light the both sides of surfaces.


hkfb (approved):
LGTM :+1:

Issue #1526, opened 2023-05-23
Prevent nested ternaries in the code


nilscb:
Rewrite all nested ternaries. 

Identify them by adding the rule ""no-nested-ternary" in eslint config file.

PR #1550, opened 2023-06-12
Fix of "Prevent nested ternaries in the code #1526"

nilscb:


react/.eslintrc
```diff
@@ -37,6 +37,7 @@
   ],
   "rules": {
     "react-hooks/rules-of-hooks": "error", // For checking rules of hooks
-    "react-hooks/exhaustive-deps": "error" // For checking hook dependencies 
+    "react-hooks/exhaustive-deps": "error", // For checking hook dependencies
+    "no-nested-ternary": 2
```
hkfb, line 41: What does "2" mean?

hkfb: Can you use "error" like above? It's clearer.

hkfb (changes requested):

react/src/lib/components/SubsurfaceViewer/components/Map.tsx
```diff
@@ -1281,11 +1281,16 @@ function createViews(
                 const currentViewport: ViewportType =
                     views.viewports[deckgl_views.length];
 
-                const ViewType = currentViewport.show3D
-                    ? OrbitView
-                    : currentViewport.id === "intersection_view"
-                    ? IntersectionView
-                    : OrthographicView;
+                let ViewType:
```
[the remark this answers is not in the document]
nilscb: It already has a default as it set to OrbitView initially.

hkfb (approved):

hkfb (approved):

Issue #827, opened 2022-02-16
Introduce conventional commits and semantic release


hkfb:
- [x] Enforce conventional commits in CI
- [x] Commitlint commit hook
- [x] Automatic changelog
- [x] Semantic release

hkfb:
Suggest using [nx-semantic-release](https://github.com/TheUnderScorer/nx-semantic-release) for automatic semantic releases

PR #1581, opened 2023-06-30
Enforce conventional commits.

hkfb:
Partial fix for #827


anders-kiaer:
I think GitHub action with conventional commits (like this one) and/or enforce changelog entries in relevant package on all PRs (https://keepachangelog.com/en/1.0.0/) is the way to go.

I'm a bit unsure on how best to use conventional commits in a mono-repo though. Do you see how it will be used in that context @hkfb?

hkfb:
> I think GitHub action with conventional commits (like this one) and/or enforce changelog entries in relevant package on all PRs (https://keepachangelog.com/en/1.0.0/) is the way to go.
> 
> I'm a bit unsure on how best to use conventional commits in a mono-repo though. Do you see how it will be used in that context @hkfb?

The main purpose besides generating changelogs is to increment the version number semantically, eg. `x.x.N` (patch), `x.N.x` (minor), etc.

It will be up to the semantic release mechanism to determine which package version should be incremented, tagged and released, eg. `subsurface-viewer@x.x.N` or `well-log-viewer@x.N.x`. It can determine that from which files have been modified in corresponding commits.

t0oF-azpn:
While I'm really in favor to setup conventional release (which will manage the version in the package.json ;) I'm not convinced that webiny/action-conventional-commits is the right tool (not very active, and I didn't see that it supports mono-repo :/ )
@hkfb can you confirm that this tool is indeed handling mono repos ?

BTW, when set-up, we will need to pay much more attention to our PR (how to write the final commit message)

hkfb:
> @hkfb can you confirm that this tool is indeed handling mono repos ?

The only purpose of this tool is to enforce that all commit messages follow the conventional commits specification, ie. `<type>[optional scope]: <description>`. Ie. PRs not following this format will be denied.

The semantic release itself, ie. parsing the commits and generating a semantic version and changelog, will come in a separate PR.

> BTW, when set-up, we will need to pay much more attention to our PR (how to write the final commit message)

Good point, I can add a note in the README.md.

t0oF-azpn (approved):

hkfb:
@anders-kiaer you will need to allow the `webiny/action-conventional-commits@03a9c76bdb2deeaf23f50e8ea9c6e99417bad467` action.

PR #1588, opened 2023-07-04
package cleaning

t0oF-azpn:
Package cleaning:
- remove ununsed package-lock.json
- fix build order (subsurface-viewer is built before well-log-viewer
- clean build rule to generate clean packages
- fix package generation
- add perr dependencies on react and react-dom
- move @mui/ to peer dependencies


t0oF-azpn:
This PR does fix #1589

t0oF-azpn:
How come that the check fails on subsurface-view/package.json tab size 2, while all the other packages file have a tab size 2 ?

Here I simply updated the tab size to be consistent with the other packages :/

typescript/package.json
```diff
@@ -26,7 +26,7 @@
     "build:py": "npm run transpile:dash && copyfiles -a ../webviz_subsurface_components/. ./webviz_subsurface_components/ && copyfiles ./package.json ./webviz_subsurface_components/ && copyfiles -f ./src/lib/assets/VectorDefinitions.json ./webviz_subsurface_components/ && dash-generate-components ./dist/lib/components webviz_subsurface_components -p package.json --ignore '(^index.js|.stories.js)$' && copyfiles -a -f ./webviz_subsurface_components/* ../webviz_subsurface_components/",
     "build": "npm run transpile && npm run build:js && npm run build:js-dev && npm run tidy-up && npm run build:py && npm run tidy-up",
     "typecheck": "tsc --noEmit",
-    "format": "eslint --fix *.js *.json *.ts packages src",
+    "format": "eslint --verbose *.js *.json *.ts packages src",
```
hkfb, line 29: Is this intentional?

t0oF-azpn: nope

typescript/packages/subsurface-viewer/package.json
```diff
@@ -1,37 +1,50 @@
...
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "transpile": "tsc --project ./tsconfig.json",
+    "build": "git clean -xdff dist && npm run transpile"
```
hkfb, line 12: Not sure, but I think you need the `copy-files` to copy some asset files.

t0oF-azpn:
Has anyone an idea why the PR does not see my last commit ?

hkfb:
> How come that the check fails on subsurface-view/package.json tab size 2, while all the other packages file have a tab size 2 ?

I seem to get different results when linting locally vs in the pipeline. Can't explain it yet.

typescript/packages/subsurface-viewer/package.json
t0oF-azpn: It has been added to subsurface-viewer and wll-log-viewer, which are the 2 only projects that have png files.

Issue #1590, opened 2023-07-04
[NGRM] - 2D rulers/axes should not overlap


nilscb:
Given for example a left and bottom ruler none of the rulers should extend all the way out in the corner. So that in this case the bottom left corner will not contain any ticks.

nilscb:
![image](https://github.com/equinor/webviz-subsurface-components/assets/136738/bbac60d8-d9ac-40c5-9c11-cbc917c72913)

PR #1592, opened 2023-07-04
Fix of "NGRM] - 2D rulers/axes should not overlap #1590"

nilscb:


typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -476,12 +476,20 @@ export default class Axes2DLayer extends Layer<Axes2DLayerProps> {
...
+        const isB = this.props.isBottomRuler;
+        const isT = this.props.isTopRuler;
+        const isL = this.props.isLeftRuler;
+        const isR = this.props.isRightRuler;
+
         //- BOTTOM RULER ----------------------------------------
-        if (this.props.isBottomRuler) {
-            const axes = [xMin, yMin + mv, zDepth, xMax, yMin + mv, zDepth];
+        if (isB) {
+            const xmin = xMin + (isL ? mh : 0);
```
hkfb, line 486: `xmin`, `xmax`, etc. calculation is duplicated, and it's confusing to have variable pairs `xmin`, `xMin`, etc. named so similarly.

hkfb (changes requested):

hkfb (approved):

PR #1609, opened 2023-07-12
Fix packages

t0oF-azpn:
- Centralize @mui/ dev dependencies.
  Move @mui/* dev dependencies to top level package.
- Remove build ordering hack in the pipeline
  Remove hack to ensure subsurface-viewer gets compile before well-log-viewer in the pipeline.
  This dependency is now handled by nx
- API changes
  - Make colorTables optional in WellLogView and WellLogViewer (fix error in lob for tests)
  - rename extensions of of some files from jsx to tsx
- Refactoring to enable tests
  - cleaning typescript/package.json to remove irrelevant scripts
  - update packages' package.json to add testing scripts
  - fix jest and babel configuration to support running the tests
  - re-enable "Run correctness tests" in github workflow
  - unfortunately, skipping tests in GroupTree.test.tsx, GroupTreeViewer.test.tsx and GroupTreeComponent.test.tsx
  - unfortunately, modifying WellLogViewr.test.tsx to make it pass
  - update the folder hierarchy to set test-related files in <package>/src/test/
    This includes duplicating performanceMetrics.ts (prev. onRenderFunction.ts) files
- Fix storybook
  - rename extensions of all the stories from jsx to tsx
  - stories: fix access to example-data
- Bump some dependencies (babel and jest related...)


rubenthoms (approved):
Everything looks good. Did mostly look at changes to the dev/build setup, not so much at other changes. However, since you already removed all scripts for generating Dash components from the main package.json file, we should maybe take this as a starting point for a discussion about how we'd like the Dash components, and, hence, the Python code to be generated in the future (maybe something to look at when you are back, @anders-kiaer, @HansKallekleiv? - I do not know if there are recent changes to the components we'd urgently like to have in Dash Webviz).

Issue #1594, opened 2023-07-04
[NGRM] - Provide a props for WellLogViewer to show\hide for discrete logs in StackedTrace view


mirisb:
Discrete well logs are shown in WellLogViewer with black lines separating the different values:
![image](https://github.com/equinor/webviz-subsurface-components/assets/18436935/291d0fa5-5a8f-4268-b148-6a36e4786fea)
vedex-wellog class StackedTrack has such option 'showLines' but currently WellLogView set it always to true.

We request a property added to template track prop of the WLV to control the appearance of the lines.

PR #1620, opened 2023-07-26
Add showLabels and showLines

Vladimir-Kokin:
Add showLabels and showLines options to template plot options for stacked track.
See https://github.com/equinor/webviz-subsurface-components/issues/1594


Vladimir-Kokin:
Create a replacement #1640

PR #1624, opened 2023-08-02
feat(grid3d): prop to show-hide grid lines added

LeonidPolukhin:
The prop "gridLines" added to Grid3d layer.

### DIFF
--- a/typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
@@ -103,6 +103,11 @@ export interface Grid3DLayerProps extends ExtendedLayerProps {
      */
     colorMapFunction?: colorMapFunctionType | false;
 
+    /** Enable lines around triangles.
+     *  default: true.
+     */
+    gridLines: boolean;
+
     /** Surface material properties.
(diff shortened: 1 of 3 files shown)


hkfb (approved):

typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
```diff
@@ -103,6 +103,11 @@ export interface Grid3DLayerProps extends ExtendedLayerProps {
      */
     colorMapFunction?: colorMapFunctionType | false;
 
+    /** Enable lines around cell faces.
+     *  default: true.
+     */
+    gridLines: boolean;
```
hkfb, line 109: Suggest making `gridLines` optional with the `?` operator

LeonidPolukhin: The same prop isn't optional in triangle layer. For consistency, should all the props of the kind be optional?

hkfb: I think so, yes

LeonidPolukhin: As it touches several layers suggest to make it in a separate task, if you don't mind.

PR #1628, opened 2023-08-04
refactor: remove dependency between WellLogViewer and SubsurfaceViewer

hkfb:
Introduced new shared dependency, `@webviz/wsc-common`.


LeonidPolukhin (approved):

typescript/nx.json
```diff
@@ -8,5 +8,10 @@
                 "cacheableOperations": ["build", "lint", "test", "e2e"]
             }
         }
+    },
+    "targetDefaults": {
```
t0oF-azpn, line 12: Does this enforce build dependency (and replace the nx.targets.build... rule in well-log-viewer/package.json ?

typescript/packages/wsc-common/package.json
```diff
@@ -0,0 +1,28 @@
+{
```
t0oF-azpn, line 1: You need to re-run a top-level npm install. It does a slight modification in the top level package-lock.json.

t0oF-azpn: OK, it seems so from my tests ;)

typescript/packages/well-log-viewer/src/WellLogViewer.stories.tsx
```diff
@@ -3,7 +3,49 @@
 import React from "react";
 import WellLogViewer from "./WellLogViewer";
 import { argTypesWellLogViewerProp } from "./WellLogViewer";
-import { colorTables } from "@emerson-eps/color-tables";
+
+import exampleData from "../../../../example-data/deckgl-map.json";
+
+import { WeakValidationMap } from "react";
+import SubsurfaceViewer from "@webviz/subsurface-viewer";
```
t0oF-azpn, line 10: Wow, I'm really not in favor of finding a dependency on subsurface-viewer here :/
Any chance to consider moving all the storybook stuff in its own package ? Storybook is anyhow built on top level-directory ?

hkfb: This doesn't add any dependency, so I don't understand the issue.

t0oF-azpn:
@hkfb Can you fix The top level package.json "lint" script ?
"eslint *.js *.json \"packages/subsurface-viewer/src/**/*.{js,jsx,ts,tsx,mdx,json}\"" => "eslint *.js *.json \"packages/*/src/**/*.{js,jsx,ts,tsx,mdx,json}\"",

typescript/packages/well-log-viewer/src/WellLogViewer.stories.tsx
t0oF-azpn: This is only because we took care to exclude stories from build (was not always the case ;) )
I think it would express clearer code organization, and avoid people looking at the code to have to think twice about it before understanding that it does not reflect a dependency...
I can do it in a separate PR.

t0oF-azpn: Let's do it in a separate PR...

PR #1629, opened 2023-08-07
Split Well Completion into control and view

### LINKED ISSUE #1567: [ENHANCEMENT] Split up `WellCompletion` in control and view components without Redux
In the new Webviz application, we'd like to use only the view part of the component and to create our own settings component. Hence, we'd like to split up this component by removing Redux and make it possible to use the view only.

jorgenherje:
Split previous `package/well-completions` package into a view-component (`well-completions-plot`), independent of `redux` and `mui`. The old `WellCompletions`-component is now placed in `packages/dash-components` for usage in Dash.

Thereby the new component named `well-completions-plot` can be utilized, making it possible to use the view only.

New packages are:
- `packages/dash-components/WellCompletions`
- `packages/well-completions-plot`

-----
Closes #1567
Closes: #1631


hkfb (commented):
Would be good with at least a storybook example for `WellCompletionsPlot`.

typescript/packages/well-completions-plot/demo/example-data/well-completions-csv2json.py
```diff
@@ -0,0 +1,464 @@
+import pandas
```
t0oF-azpn, line 1: This file already exists in /example-data. Can we use the top-level file instead of duplicatiog it ?

typescript/packages/well-completions-plot/demo/example-data/well-completions.json
```diff
@@ -0,0 +1,2390 @@
+{
```
t0oF-azpn, line 1: This file already exists in /example-data. Can we use the top-level file instead of duplicating it ?

jorgenherje:
> Would be good with at least a storybook example for `WellCompletionsPlot`.

The `WellCompletionsPlot` is used in the package `dash-components/WellCompletions`, which has a storybook. Would you prefer a separate storybook for the `WellCompletionsPlot` as well? 

As it is a static plot view, I can fore instance retrieve 3 "states" of data from the `WellCompletions` module and provide a storybook which can show these 3 different views for the `WellCompletionsPlot`?

typescript/packages/well-completions-plot/demo/example-data/well-completions.json
jorgenherje: This file is not used for the `WellCompletionsPlot`, so I'll delete it. However the `dash-components/WellCompletions`-package is a rename and move of the original/refactored `WellCompletions`-package, which has the same files in its folder `WellCompletions/example-data`. Is it preferred to have a top-level folder for example-data, or keep the example-data close  to the owner?

typescript/packages/well-completions-plot/src/WellCompletionsPlot.tsx
```diff
@@ -1,47 +1,45 @@
...
-const Root = styled("div")(({ theme }) => ({
-    display: "flex",
-    flex: 1,
-    height: "80%",
-    padding: theme.spacing(1),
-}));
+import "./WellCompletionsPlot.css";
 
-interface Props {
+interface WellCompletionsPlotProps {
```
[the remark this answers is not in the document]
jorgenherje: Good point

typescript/packages/well-completions-plot/demo/example-data/well-completions-csv2json.py
jorgenherje: Not used, so will be deleted

hkfb:
> > Would be good with at least a storybook example for `WellCompletionsPlot`.
> 
> The `WellCompletionsPlot` is used in the package `dash-components/WellCompletions`, which has a storybook. Would you prefer a separate storybook for the `WellCompletionsPlot` as well?

Yes, ideally we should have a story for each component. It can be a very basic example. The main purpose is to be able to catch and isolate regressions and also to demonstrate the API and usage.

jorgenherje:
> > > Would be good with at least a storybook example for `WellCompletionsPlot`.
> > 
> > 
> > The `WellCompletionsPlot` is used in the package `dash-components/WellCompletions`, which has a storybook. Would you prefer a separate storybook for the `WellCompletionsPlot` as well?
> 
> Yes, ideally we should have a story for each component. It can be a very basic example. The main purpose is to be able to catch and isolate regressions and also to demonstrate the API and usage.

Okay. I've added a storybook for the component.

@hkfb and @t0oF-azpn - Last question from my side: Should the `well-completions.json` be placed at the top-level `example-data`-folder, or should it be placed within the package `dash-components/src/WellCompletions`?

hkfb:
> @hkfb and @t0oF-azpn - Last question from my side: Should the `well-completions.json` be placed at the top-level `example-data`-folder, or should it be placed within the package `dash-components/src/WellCompletions`?

I guess if it's only relevant for that specific package, then it makes sense to keep it local to that package.

t0oF-azpn:
> > @hkfb and @t0oF-azpn - Last question from my side: Should the `well-completions.json` be placed at the top-level `example-data`-folder, or should it be placed within the package `dash-components/src/WellCompletions`?
> 
> I guess if it's only relevant for that specific package, then it makes sense to keep it local to that package.
> 
> Could be, however, that you might face difficulties if you want to use that file as an external shared file in storybook. In that case you might have to put in the global `example-data`, which is the directory that contains static files shared through storybook.

It's used for stories and for tests.
As long as the file is not duplicated, it's fine for me.

typescript/packages/well-completions-plot/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "version": "0.0.0",
+    "description": "",
+    "main": "dist/index.js",
+    "types": "dist/index.d.ts",
+    "files": [
+        "/dist"
+    ],
+    "scripts": {
+        "transpile": "tsc --project ./tsconfig.json",
+        "copy-files": "copyfiles package.json dist && copyfiles README.md dist && copyfiles --exclude **/*.{tsx,jsx,js,ts,snap} --up 1 \"src/**/*\" dist/",
```
t0oF-azpn, line 12: please use the same template than other packages.
package.json and README.md are automatically copied (no need here), and former copyfile lead to duplicated code at the wrong place in the generated package :/

hkfb (approved):
LGTM :+1:

t0oF-azpn (approved):

typescript/packages/well-completions-plot/src/WellCompletionsPlot.tsx
```diff
@@ -1,80 +1,80 @@
...
+export const WellCompletionsPlot: React.FC<WellCompletionsPlotProps> =
+    React.memo((props: WellCompletionsPlotProps) => {
         const { width, height, ref } = useResizeDetector({
             refreshMode: "debounce",
             refreshRate: 50,
             refreshOptions: { trailing: true },
         });
-        const layout = useMemo(
+
+        const layout = React.useMemo(
```
rubenthoms, line 31: Consider if `useMemo` is necessary here. The computation in `createLayout` does not seem very heavy.

typescript/packages/well-completions-plot/src/WellCompletionsPlot.tsx
```diff
@@ -1,80 +1,80 @@
...
+                    {layout && props.plotData && (
                         <svg
                             id={"svg-context"}
                             width={width}
                             height={height}
                             style={{ position: "relative" }}
                         >
                             <StratigraphyPlot
-                                data={plotData.stratigraphy}
+                                data={props.plotData?.stratigraphy ?? []}
```
rubenthoms, line 56: In line 48, you do already assert that `props.plotData` is defined.

typescript/packages/dash-components/src/WellCompletions/components/DataLoader.tsx
```diff
@@ -1,9 +1,11 @@
 import React, { PropsWithChildren, ReactNode, useMemo } from "react";
 import { Provider as ReduxProvider } from "react-redux";
 import { createReduxStore } from "../redux/store";
-import { Data, UISettings, Zone } from "../redux/types";
+import { Data, UISettings } from "../redux/types";
 import { findSubzones, preprocessData } from "../utils/dataUtil";
 
+import { Zone } from "../../../../well-completions-plot/src/types/dataTypes";
```
rubenthoms, line 7: Import directly from `well-completions-plot` without going any deeper?

typescript/packages/well-completions-plot/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { WellCompletionsPlot } from "./WellCompletionsPlot";
+export { CompletionPlotData, PlotData, WellPlotData } from "./types/dataTypes";
```
rubenthoms, line 2: You could include e.g. `Zone` here as well since it's consumed by the `dash/WellCompletions` component.

typescript/packages/dash-components/src/WellCompletions/components/Settings/SortTable.tsx
```diff
@@ -6,13 +6,10 @@ import React, { useCallback, useEffect, useMemo, useState } from "react";
...
-import { SortDirection } from "../../redux/types";
 import {
-    SORT_BY_COMPLETION_DATE,
-    SORT_BY_NAME,
-    // eslint-disable-next-line prettier/prettier
-    SORT_BY_STRATIGRAPHY_DEPTH
-} from "../../utils/sort";
+    SortBy,
+    SortDirection,
+} from "../../../../../well-completions-plot/src/types/dataTypes";
```
rubenthoms, line 12: Expose these type in main `index.ts`?

typescript/packages/dash-components/src/WellCompletions/components/Settings/ZoneSelector.tsx
```diff
@@ -7,9 +7,9 @@ import DropdownTreeSelect, {
 import "react-dropdown-tree-select/dist/styles.css";
 import { useDispatch } from "react-redux";
 import { updateFilteredZones } from "../../redux/actions";
-import { Zone } from "../../redux/types";
 import { findSubzones } from "../../utils/dataUtil";
 import { DataContext } from "../DataLoader";
+import { Zone } from "../../../../../well-completions-plot/src/types/dataTypes";
```
rubenthoms, line 12: See above.

typescript/packages/dash-components/src/WellCompletions/components/WellCompletionsViewer.tsx
```diff
@@ -7,7 +7,7 @@ import ReactResizeDetector from "react-resize-detector";
 import { usePlotData } from "../hooks/usePlotData";
 import { WellCompletionsState } from "../redux/store";
 import { DataContext } from "./DataLoader";
-import WellCompletionsPlot from "./Plot/WellCompletionsPlot";
+import { WellCompletionsPlot } from "../../../../well-completions-plot/src/WellCompletionsPlot";
```
rubenthoms, line 10: Import from `well-completions-plot` (main index.ts file)?

typescript/packages/dash-components/src/WellCompletions/components/WellCompletionsViewer.tsx
```diff
@@ -154,6 +155,7 @@ const WellCompletionsViewer: React.FC = () => {
                                     }}
                                 >
                                     <WellCompletionsPlot
+                                        id={"well-completions-plot"}
```
rubenthoms, line 158: What if two `WellCompletions` are used on same page, they'll end up with same id. Consider https://react.dev/reference/react/useId.

typescript/packages/dash-components/src/WellCompletions/redux/reducer.ts
```diff
@@ -1,11 +1,11 @@
 import { combineReducers, createSlice, PayloadAction } from "@reduxjs/toolkit";
 import {
     Attributes,
-    SortDirection,
     TimeAggregation,
     // eslint-disable-next-line prettier/prettier
-    UISettings
+    UISettings,
 } from "./types";
+import { SortDirection } from "../../../../well-completions-plot/src/types/dataTypes";
```
rubenthoms, line 8: Expose `SortDirection` at top level `index.ts`? (see above)

typescript/packages/dash-components/src/WellCompletions/test/testReduxState.ts
```diff
@@ -1,10 +1,6 @@
 import { UISettings } from "../redux/types";
-import {
-    SORT_BY_COMPLETION_DATE,
-    SORT_BY_NAME,
-    // eslint-disable-next-line prettier/prettier
-    SORT_BY_STRATIGRAPHY_DEPTH
-} from "../utils/sort";
+import { SortBy } from "../../../../well-completions-plot/src/types/dataTypes";
```
rubenthoms, line 2: See above.

typescript/packages/dash-components/src/WellCompletions/utils/dataUtil.ts
```diff
@@ -1,13 +1,14 @@
...
+    CompletionPlotData,
+    PlotData,
+    Units,
     WellInfo,
-    // eslint-disable-next-line prettier/prettier
-    Zone
-} from "../redux/types";
+    WellPlotData,
+    Zone,
+} from "../../../../well-completions-plot/src/types/dataTypes";
```
rubenthoms, line 11: See above.

typescript/packages/dash-components/src/WellCompletions/utils/sort.ts
```diff
@@ -1,22 +1,24 @@
...
-//Deault sort methods
-export const SORT_BY_NAME = "well name";
-export const SORT_BY_STRATIGRAPHY_DEPTH = "stratigraphy depth";
-export const SORT_BY_COMPLETION_DATE = "earliest comp date";
+import {
+    AttributeType,
+    SortBy,
+    SortDirection,
+    WellPlotData,
+} from "../../../../well-completions-plot/src/types/dataTypes";
```
rubenthoms, line 6: See above.

typescript/packages/well-completions-plot/src/WellCompletionsPlot.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+const Template = (data) => {
+    const [plotData, setPlotData] = React.useState<PlotData>(data.plotData);
+    const [timeSteps, setTimeSteps] = React.useState<string[]>(data.timeSteps);
+
+    React.useEffect(() => {
+        setPlotData(data.plotData);
+    }, [data.plotData]);
+    React.useEffect(() => {
+        setTimeSteps(data.timeSteps);
+    }, [data.timeSteps]);
```
rubenthoms, line 35: According to https://react.dev/learn/you-might-not-need-an-effect, consider removing the effects and store the previous states.

rubenthoms (changes requested):
Well done :+1: A couple of suggestions and some comments that hopefully help to improve the code a bit.

typescript/packages/well-completions-plot/src/types/layoutTypes.ts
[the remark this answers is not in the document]
jorgenherje: Placed function into `utils`

typescript/packages/well-completions-plot/src/types/dataTypes.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export interface PlotData {
+    stratigraphy: Zone[];
+    wells: WellPlotData[];
+    units: Units;
+}
+
+// ---------------------------  PropTypes ---------------------------------------
+
```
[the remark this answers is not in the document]
jorgenherje: No link, but added comment describing issue

typescript/packages/dash-components/src/WellCompletions/components/Settings/SortTable.tsx
```diff
@@ -6,13 +6,11 @@ import React, { useCallback, useEffect, useMemo, useState } from "react";
...
 import { WellCompletionsState } from "../../redux/store";
-import { SortDirection } from "../../redux/types";
 import {
-    SORT_BY_COMPLETION_DATE,
-    SORT_BY_NAME,
-    // eslint-disable-next-line prettier/prettier
-    SORT_BY_STRATIGRAPHY_DEPTH
-} from "../../utils/sort";
+    SortBy,
+    SortByString,
```
rubenthoms, line 11: `SortByString` sounds like something to use for sorting by a given string. Maybe `SortByEnumToStringMapping`?

typescript/packages/well-completions-plot/src/WellCompletionsPlot.tsx
```diff
@@ -0,0 +1,77 @@
...
+                <div
+                    ref={ref as React.LegacyRef<HTMLDivElement>}
+                    id={props.id}
+                    className={"WellCompletionsPlot"}
+                    data-tip
+                    data-for="plot-tooltip"
+                >
+                    {layout && props.plotData && (
+                        <svg
+                            id={"svg-context"}
```
rubenthoms, line 47: Consider using a combination of `props.id` and `svg-context` here.

typescript/packages/well-completions-plot/src/components/WellsPlot.tsx
```diff
@@ -0,0 +1,133 @@
...
+    props: WellTooltipContentProps
+) => {
+    return (
+        <table style={{ color: "#fff" }}>
+            <tbody>
+                {/* earliest completion date */}
+                <tr key={`well-tooltip-${props.name}-earliest-comp`}>
+                    <td>
+                        <b>
+                            {capitalizeFirstLetter(
```
rubenthoms, line 31: Still necessary with the enum -> string mapping?

rubenthoms (changes requested):
A few comments. Otherwise, good work 😃

typescript/packages/well-completions-plot/src/components/WellsPlot.tsx
jorgenherje: The `SortByEnumToStringMapping` is used other places where lower case is preferred, so I'll keep lower case in mapping and use `capitalizeFirstLetter` here.

typescript/packages/well-completions-plot/src/types/dataTypes.ts
```diff
@@ -0,0 +1,93 @@
...
+export enum SortBy {
+    Name = "well name",
+    StratigraphyDepth = "stratigraphy depth",
+    CompletionDate = "earliest comp date",
+}
+
+export const SortByEnumToStringMapping = {
+    [SortBy.Name]: "well name",
+    [SortBy.StratigraphyDepth]: "stratigraphy depth",
+    [SortBy.CompletionDate]: "earliest comp date",
```
rubenthoms: Use capital letters? These should only be used when exposing enum to GUI, hence, this is what the user is going to see when interacting with the enum.

rubenthoms (approved):
LGTM

PR #1634, opened 2023-08-16
build(deps): replace eds-core-react with mui in subsurface-viewer

hkfb:
Also removes unused settings components in `subsurface-viewer`


t0oF-azpn:
I don't know if no-one is using the settings.

But if you remove the settings, you can also remove the whole redux directory and utils/specExtractor.ts...

Issue #1646, opened 2023-08-31
[NGRM] - move shader code to local shader module


nilscb:
triangle layer uses a modifcation of luma's shader code to calculate two sided phong lighting. This code should be moved into a local shader module so that it may also be easily used by other layers.

PR #1647, opened 2023-08-31
refactor: Fix of [NGRM] - move shader code to local shader module #1646

nilscb:


typescript/packages/subsurface-viewer/src/layers/shader_modules/localPhongLighting.ts
```diff
@@ -0,0 +1,8 @@
+import { fs } from "./lighting.fs.glsl";
+import { phongLighting } from "@luma.gl/shadertools";
+
+export default {
```
hkfb, line 4: Possible to avoid `default` export here also?

hkfb (approved):

PR #1651, opened 2023-09-08
build: use personal access token for semantic release

hkfb:
Need to bypass branch protection in order to create release tags.


t0oF-azpn (approved):
Seems strange to use your personal account and email...

You can resolve

hkfb:
> Seems strange to use your personal account and email...
> 
> You can resolve

It has to be an admin user to bypass the branch protection. Had to use my personal access token. Not ideal, but this is the only solution I could find besides disabling branch protection.

PR #1656, opened 2023-09-15
fix: Improve "isLoadedCallback" to not return true before first dra…

nilscb:
…w of layer not only just when data are loaded.


hkfb:
Apparently the "patch" prefix is not supported. This looks like a "fix".

hkfb (approved):

PR #1667, opened 2023-09-25
fix: Remove duplicate vertices from well path. May cause crash

nilscb:


hkfb (changes requested):
Can we do this in `initializeState` instead, so that duplicates are also removed when not interpolating?

Would it be useful to get a warning if duplicates occur?

Also, a unit test would be good for this.

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.test.ts
hkfb, line 1: Would be good to capture also corner cases, like eg. an empty set of coordinates

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.ts
```diff
@@ -7,6 +7,72 @@ import { cloneDeep, range } from "lodash";
...
+        }
+
+        let coords = lineString.coordinates as Position3D[];
+
+        const nOrig = coords.length;
+        [coords, mds[0]] = removeConsecutiveDuplicates(coords, mds[0]);
+
+        const n = coords.length;
+        if (n != nOrig) {
+            console.warn("Well no ", well_no, " contains duplicates.");
```
hkfb, line 57: "Well no" is confusing. Better to spell it out.

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.ts
```diff
@@ -246,14 +315,29 @@ export function coarsenWells(data_in: FeatureCollection): FeatureCollection {
...
-
-        const coordsSimplified = simplify(lineString, options);
-        lineString.coordinates = coordsSimplified.coordinates as Position3D[];
+        const n = lineString.coordinates.length;
+        const isHorizontalWell =
+            lineString.coordinates[0][0] == lineString.coordinates[n - 1][0] &&
+            lineString.coordinates[0][1] == lineString.coordinates[n - 1][1];
+
+        if (isHorizontalWell) {
+            // The simplify algorithm below did not work on horizontal wells hence in this case we only use first and last point.
```
hkfb, line 324: Do you mean "vertical"?

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.ts
```diff
@@ -246,14 +315,29 @@ export function coarsenWells(data_in: FeatureCollection): FeatureCollection {
...
-            highQuality: false,
-            mutate: false,
-        };
-
-        const coordsSimplified = simplify(lineString, options);
-        lineString.coordinates = coordsSimplified.coordinates as Position3D[];
+        const n = lineString.coordinates.length;
+        const isHorizontalWell =
+            lineString.coordinates[0][0] == lineString.coordinates[n - 1][0] &&
+            lineString.coordinates[0][1] == lineString.coordinates[n - 1][1];
```
hkfb, line 321: This doesn't look like a proper check for a horizontal or vertical well

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.stories.tsx
```diff
@@ -437,6 +437,40 @@ Wells3d.parameters = {
     },
 };
 
+export const OneWell = Template.bind({});
+OneWell.args = {
+    id: "well_176",
+    bounds: [756125, 166200, 757327, 167938],
+    layers: [
+        {
+            "@@type": "WellsLayer",
```
hkfb, line 446: Better to use the class directly in order to get proper type checking.

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.stories.tsx
```diff
@@ -437,6 +437,40 @@ Wells3d.parameters = {
...
+            data: "./well_176.json",
```
hkfb, line 447: Do we need more well data examples?

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.stories.tsx
```diff
@@ -437,6 +437,40 @@ Wells3d.parameters = {
     },
 };
 
+export const OneWell = Template.bind({});
```
hkfb, line 440: This story name is not very descriptive.

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.ts
nilscb: I think this is sufficient in 99.9999% of the cases. But I could of course di a more expensive test and use all points in the well path.

example-data/one_well_duplicate_vertexes.json
hkfb, line 1: Maybe nitpicking now, but normally it's called "vertices"

nilscb: Good point! Spelling is important! (I learned something new :-)

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.stories.tsx
```diff
@@ -437,6 +437,70 @@ Wells3d.parameters = {
...
+WellWithDuplicates.args = {
+    id: "well_176",
+    bounds: [756125, 166200, 757327, 167938],
+    layers: [
+        new WellsLayer({
+            data: "./one_well_duplicate_vertices.json",
+            refine: true,
+            simplifiedRendering: true,
+        }),
+    ],
```
hkfb, line 481: I think it would be better to create some simple synthetic well trajectories in the stories themselves instead of adding more well files.

hkfb:
The well is invisible until I zoom out: 
image
image

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -248,6 +249,8 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
                 data = invertPath(data);
             }
 
+            removeDuplicates(data);
```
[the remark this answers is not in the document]
nilscb: In the unit test the function "removeConsecutiveDuplicates" is tested. But I changed "removeDuplicates" so that it may only be used in one way.

PR #1673, opened 2023-09-28
Created new setup for generating Dash components in Python

rubenthoms:
- Moved Python-only components to Python directory and adjusted code and setup
- Created wrapper components for `typescript/packages`
- Removed stories and tests in Python directory
- Adjusted package.json entry points in `group-tree` package
- Added `DashSubsurfaceViewer` to `subsurface-viewer` package


rubenthoms:
Commit messages not according to semantic release requirements.

Issue #1690, opened 2023-10-10
View resets to default when adding/removing a layer.


nilscb:
This is best illustrated with the story "IsRenderedCallback".
 - Set "show3D" to true.
 - Rotate a bit the view
 - Add a layer by using the switch at the top
 - Observe that the view resets to default.

PR #1691, opened 2023-10-10
fix: View resets to default when adding/removing a layer. #1690

nilscb:
- for some reason this has gone missing from the code at some point.

### DIFF
--- a/typescript/packages/subsurface-viewer/src/components/Map.tsx
+++ b/typescript/packages/subsurface-viewer/src/components/Map.tsx
@@ -589,6 +589,10 @@ const Map: React.FC<MapProps> = ({
 
     const bboxInitial: BoundingBox3D = [0, 0, 0, 1, 1, 1];
 
+    const isCameraPositionDefined =
+        typeof cameraPosition !== "undefined" &&
+        Object.keys(cameraPosition).length !== 0;
+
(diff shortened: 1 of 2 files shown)


typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -647,6 +654,9 @@ const Map: React.FC<MapProps> = ({
     }, [triggerHome]);
 
     useEffect(() => {
+        if (didUserChangeCamera && !isCameraPositionDefined) {
+            return;
+        }
         const [Views, viewStates] = createViewsAndViewStates(
```
hkfb, line 660: Is what you want to achieve that `createViewsAndViewStates` is only called on first render and if camera is defined?

nilscb: The aim is to not call "createViewsAndViewStates  (and hence reset the camera) if the user has manipulated the camera by interacting with it.

hkfb: I think `onViewStateChanged` could be called for other reasons, like eg. syncing cameras across components, so maybe `didUserChangeCamera` should be named something like `viewStateChanged`?

hkfb (approved):

PR #1696, opened 2023-10-13
build: Bump Deck.gl to latest version.

nilscb:
-This should contain a fix for the scrolling issues we have experienced.


t0oF-azpn (approved):

hkfb (changes requested):
Need to add your `package-lock.json` also.

t0oF-azpn:
> Need to add your `package-lock.json` also.

Good catch !

hkfb (commented):
Is it just me or do the wells look slightly blurred now?

image

t0oF-azpn:
> Is it just me or do the wells look slightly blurred now?
> 
> image

They are blurred, but it is already the case in the published storybook (ie. if there is a regression, it does not come from this PR):

Image from the storybbok:
image

hkfb:
> > Is it just me or do the wells look slightly blurred now?
> > image
> 
> They are blurred, but it is already the case in the published storybook (ie. if there is a regression, it does not come from this PR):
> 
> Image from the storybbok: image

The upper image looks more blurry to me.

nilscb:
Its a bit hard to say. I agree that the upper appears slightly  more blurred

hkfb (approved):
I'll change to approved and let you decide. I think we should drop a note to deck.gl about the blurred polygons

t0oF-azpn:
I confirm that there is no visible difference in full-screen on my machine.

PR #1726, opened 2023-10-20
fix: prevent situation where width is defined and 0

nilscb:


t0oF-azpn (changes requested):
We discussed an other fix to handle all the places where the issue can occur...

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -480,14 +480,14 @@ function calculateZoomFromBBox3D(
         return camera;
     }
 
-    if (typeof deck === "undefined") {
+    if (!deck) {
```
t0oF-azpn, line 483: if( !deck?.width || !deck?.height) {

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -480,14 +480,14 @@ function calculateZoomFromBBox3D(
...
         camera_.zoom = 0;
         camera_.target = [0, 0, 0];
         return camera_;
     }
 
-    const width = deck.width;
-    const height = deck.height;
+    const width = (deck.width ?? 1) || 1;
```
t0oF-azpn, line 489: revert

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1299,9 +1299,9 @@ function getViewState3D(
 
     let width = xMax - xMin;
     let height = yMax - yMin;
-    if (deck) {
-        width = deck.width;
-        height = deck.height;
+    if (deck?.width && deck?.height) {
```
t0oF-azpn, line 1302: if (deck?.width > 0 && deck?.height > 0) {

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1299,9 +1299,9 @@ function getViewState3D(
...
+        width = deck.width || 1;
```
t0oF-azpn, line 1303: revert

t0oF-azpn (changes requested):

typescript/packages/subsurface-viewer/src/components/Map.tsx
nilscb: Need to do like this since deck may be undefined
    if (deck && deck?.width > 0 && deck?.height > 0) {
        width = deck.width;
        height = deck.height;
    }

If deck is defined then decl.width must be defined (I think)

t0oF-azpn (approved):

PR #1733, opened 2023-10-24
fix: Grid3DLayer and MapLayer;  Depth readout not reported for value 0

nilscb:
- also ZIncreasingDownward flag not correct in Grid3DLayer story.

### DIFF
--- a/typescript/packages/subsurface-viewer/src/layers/grid3d/privateLayer.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/grid3d/privateLayer.ts
@@ -258,7 +258,7 @@ export default class privateLayer extends Layer<privateLayerProps> {
 
         const vertexIndex = 256 * 256 * r + 256 * g + b;
 
-        if (info.coordinate?.[2]) {
+        if (typeof info.coordinate?.[2] !== "undefined") {
             const depth = this.props.ZIncreasingDownwards
                 ? -info.coordinate[2]
                 : info.coordinate[2];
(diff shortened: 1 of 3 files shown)


typescript/packages/subsurface-viewer/src/layers/grid3d/privateLayer.ts
```diff
@@ -258,7 +258,7 @@ export default class privateLayer extends Layer<privateLayerProps> {
 
         const vertexIndex = 256 * 256 * r + 256 * g + b;
 
-        if (info.coordinate?.[2]) {
+        if (typeof info.coordinate?.[2] !== "undefined") {
```
t0oF-azpn, line 261: This was correct: no need to invert 0.

The fix is to move line 265 (ie. layer_properties.push(...)) outside the if block (ie. after line 266)

typescript/packages/subsurface-viewer/src/layers/map/privateMapLayer.ts
```diff
@@ -262,7 +262,7 @@ export default class privateMapLayer extends Layer<privateMapLayerProps> {
 
         const vertexIndex = 256 * 256 * r + 256 * g + b;
 
-        if (info.coordinate?.[2]) {
+        if (typeof info.coordinate?.[2] !== "undefined") {
```
t0oF-azpn, line 265: Revert this line.
Move line 269 after line 270.

t0oF-azpn (changes requested):

t0oF-azpn (approved):

PR #1750, opened 2023-11-03
ci: Add script to workflow for testing storybooks.

nilscb:


typescript/packages/subsurface-viewer/src/storybook/components/settings/LayerSettingsButton.stories.tsx
```diff
@@ -6,29 +6,31 @@ export default {
...
-DrawingSettings.args = {
-    layer: {
-        name: "Drawing",
-        "@@type": "DrawingLayer",
-        id: "drawing-layer",
-        visible: true,
-        mode: "drawLineString",
-    },
-};
+// const Template = (args) => {
```
hkfb, line 11: Suggest removing instead of commenting out. The code is still in the git history.

.github/workflows/smoke_test.yml
```diff
@@ -0,0 +1,41 @@
...
+    name: Smoke Tests
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
```
hkfb, line 23: You need version >= 18. I've experienced some instabilities with version 18 in the past, but let's try

.github/workflows/smoke_test.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          node-version: '>= 18'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install Test Dependencies
+        run: npm install concurrently http-server wait-on
+
+      - name: Build Storybook
```
hkfb, line 33: I guess we need to build the wsc packages first using `nx`. But we could also consider adding the tests to the `typescript.yml` workflow instead, where everything is build anyway, including storybook and the wsc packages

hkfb (approved):

PR #1755, opened 2023-11-06
test: fix storybook test runner

hkfb:
The tests were not run due to misuse of `concurrently`

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -115,6 +115,7 @@ jobs:
           HOST: 127.0.0.1
         run: |
           npm install concurrently http-server wait-on
+          npx playwright install
           npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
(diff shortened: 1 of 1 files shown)


.github/workflows/typescript.yml
```diff
@@ -111,10 +111,10 @@ jobs:
 
       - name: Serve Storybook and run tests
         env:
-            PORT: 8080
-            HOST: 127.0.0.1
+          PORT: 8080
+          HOST: 127.0.0.1
         run: |
-            npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
-              "npx playwright install"  \
```
nilscb, line 118: Is this not needed anymore?

nilscb (approved):

t0oF-azpn (approved):

.github/workflows/typescript.yml
hkfb: It was needed, so re-added it

PR #1762, opened 2023-11-07
fix: If camera or bounds property is set, the view should not change when adding a layer.

nilscb:
I had to split an effect to accomplish this.


t0oF-azpn (approved):
Tested locally with viewers. works fine.

PR #1787, opened 2023-11-24
feat(PointsLayer):  ZIncreasingDownwards is applied in vertex shader, depthTest flag added

LeonidPolukhin:
1) For performance boost the layer applies ZIncreasingDownwards in the vertex shader.
2) depthTest flag is added in the layer API.
3) Fixed crash in the story book when a property of the layer is changed.

### DIFF
--- a/typescript/packages/subsurface-viewer/src/layers/utils/layerTools.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/utils/layerTools.ts
@@ -184,7 +184,7 @@ export function defineBoundingBox(
         maxZ = z > maxZ ? z : maxZ;
     }
     if (zIncreasingDownwards) {
-        [maxZ, minZ] = [minZ, maxZ];
+        [maxZ, minZ] = [-minZ, -maxZ];
     }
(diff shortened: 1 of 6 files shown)


typescript/packages/subsurface-viewer/src/layers/utils/layerTools.ts
```diff
@@ -182,6 +183,9 @@ export function defineBoundingBox(
         maxY = y > maxY ? y : maxY;
         maxZ = z > maxZ ? z : maxZ;
     }
+    if (zIncreasingDownwards) {
+        [maxZ, minZ] = [minZ, maxZ];
```
nilscb, line 187: Do you need to multiply by -1 here?

nilscb (approved):

typescript/packages/subsurface-viewer/src/layers/utils/layerTools.ts
LeonidPolukhin: Good catch, thank you. Fixed.

PR #1818, opened 2023-12-05
feat: Well Markers Layer

LeonidPolukhin:
The layer can display markers of three flat shapes: circle, square and triangle.
For every marker the following attributes can be defined:
- center position in 3D.
- azimuth,  rotation angle relative to north direction.
- inclination, rotation angle relative to vertical direction.
- size.
- fill color with transparency.
- outline color.


typescript/packages/subsurface-viewer/src/layers/well_markers/wellMarkersLayer.stories.tsx
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable prettier/prettier */
```
hkfb, line 1: Should try to honor prettier rules and only override in specific cases

typescript/packages/subsurface-viewer/src/layers/well_markers/wellMarkersLayer.ts
```diff
@@ -0,0 +1,386 @@
+/* eslint-disable prettier/prettier */
```
hkfb, line 1: Should try to honor prettier rules and only override in specific cases

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/well_markers/wellMarkersLayer.ts
LeonidPolukhin: oops, good catch, thanks!

LeonidPolukhin: prettier is enabled, style fixed.

typescript/packages/subsurface-viewer/src/layers/well_markers/wellMarkersLayer.stories.tsx
LeonidPolukhin: prettier is enabled, style fixed.

hkfb (approved):

PR #1819, opened 2023-12-07
fix: camera state when hidden page is rendered

t0oF-azpn:
Fix for #1820 

Add formal size state and effect to retrieve size from deck ref for code clarity
Move global bounding box handling to React reducer instead of useEffect.
Move camera computation to useMemo instead of useEffect.
Memoize viewPortMargins and deckGLLayers

Add a new story displaying the subsurface viewer in tabs to reproduce the issue

Note that this is a first step to reduce the number of useEffect in the Map component

### DIFF
--- a/typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
+++ b/typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
@@ -58,7 +58,11 @@ export interface SubsurfaceViewerProps {
     id: string;
     resources?: Record<string, unknown>;
     layers?: Record<string, unknown>[] | LayersList;
+
     bounds?: [number, number, number, number] | BoundsAccessor;
+    cameraPosition?: ViewStateType | undefined;
+    triggerHome?: number;
+
     views?: ViewsType;
     coords?: {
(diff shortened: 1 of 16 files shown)


hkfb:
> Note that this is a first step to reduce the number of useEffect in the Map component

This is just refactoring and creating an example to demonstrate the bug?

nilscb (approved):

t0oF-azpn:
> > Note that this is a first step to reduce the number of useEffect in the Map component
> 
> This is just refactoring and creating an example to demonstrate the bug?

Nope, it fixes also the bug (ie. the example was showing the bug before I fixed it in the same PR)

hkfb (commented):
I get this error in the console: 
image
Not sure if it's related to your change though.

hkfb (commented):
The second tab seems wrongly initialized the first time it's opened:
image
The second time I open it I can see the map.

t0oF-azpn:
> The second tab seems wrongly initialized the first time it's opened: image The second time I open it I can see the map.

:/
How are you doing ?
What is your browser ??

hkfb (approved):

PR #1868, opened 2024-01-12
fix: zoom keys (arrow up and down) only apply to last SubSurfaceViewer

t0oF-azpn:
Now the Z scaling applies to the SubSurfaceViewer that has the focus (ie. the one selected).

Update SyncedCameraSettingsComponent to control camera sync (on/off).


hkfb (approved):
LGTM :+1:

PR #1882, opened 2024-01-19
fix: do not display "Loading assets" feedback when isRenderedCallback exists

t0oF-azpn:
The client code can handle the assets loading feedback on its own using isRenderedCallback. This way it can control the look, feel and layout of this feedback, and will prevent possible inference with application own components.

BTW, this default feedback gets centered in the PR.

Viewers team has an example where the feedback component does move the application components...

### DIFF
--- a/typescript/packages/subsurface-viewer/src/components/Map.tsx
+++ b/typescript/packages/subsurface-viewer/src/components/Map.tsx
@@ -417,7 +417,7 @@ export interface MapProps {
      * Will be called while layers have rendered data.
      * progress is a number between 0 and 100.
      */
-    onRenderedProgress?: (progress: number) => void;
+    onRenderingProgress?: (progress: number) => void;
 
     onDragStart?: (info: PickingInfo, event: MjolnirGestureEvent) => void;
     onDragEnd?: (info: PickingInfo, event: MjolnirGestureEvent) => void;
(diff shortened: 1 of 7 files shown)


typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -896,8 +896,25 @@ const Map: React.FC<MapProps> = ({
...
+                        height: "90%",
+                        width: "90%",
+                        zIndex: 200,
+                    }}
+                >
+                    <StatusIndicator
+                        layers={deckGLLayers}
+                        isLoaded={isLoaded}
+                    />
+                </div>
```
hkfb, line 915: I think centering the progress indicator is too obtrusive. Maybe it could be repositioned to the lower corner.

Moreover, I think we can remove the whole `ProgressIndicator` and instead provide a progress callback similar to (or replacing) the `isRenderedCallback`, ie. something like `loadProgress(progress: number)`, where `progress` is in [0..1]. That way the consumer of this component could add use the `ProgressIndicator` as a `ViewAnnotation` or use any progress display any way they want.

t0oF-azpn: OK to move it to lower corner.

Unfortunately, there is no way to have a `progress` number. Thus the `isRenderedCallback` is indeed the way to provide the only available information (`on` or `off`) that allows the client code to display this information...

t0oF-azpn: change callback to `onRenderedProgress` which takes a number between 0 and 100.
moved default feedback to bottom right:
image

Example with use of callback:
image

LeonidPolukhin (approved):

PR #1890, opened 2024-01-26
feat(SubsurfaceViewer): Record-based definition of layers that supports TS Objects implemented.

LeonidPolukhin:
1) typedArraySupport optional flag is added to SubsurfaceViewer API. If set to true records with layer definitions are used for layer creation bypassing JSONConverter.
2) PointsLayer, TriangleLayer API are updated to accept typed arrays as input data skipping extra copying. The corresponding stories are added.
3) MapLayer and Grid3DLayer already support typed array inputs. The stories updated.

PolylinesLayer has an internal algo for making polylines closed when requested so its modification is postponed for a while.

Known issue: in the story book "typed array stories" don't work if any property of layer is changed in Controls Tab. It happens because the story book deserializes input arrays from JSON represented in Controls loosing typeinfos.


typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
```diff
@@ -123,6 +123,11 @@ export interface SubsurfaceViewerProps {
     lights?: LightsType;
 
     children?: React.ReactNode;
+
+    /**
+     * Forces the component to use the records provided in 'layers' prop as is.
+     */
+    skipLayerPropsConversion?: boolean;
```
LeonidPolukhin, line 130: Not happy with this name, any better proposals?

typescript/packages/subsurface-viewer/src/layers/map/mapLayer.stories.tsx
```diff
@@ -574,7 +574,9 @@ const TypedArrayInputComponent: React.FC<{
         id: "map",
         layers: [
             // Can not use Record<string, unknown> because makeData() is not supported
```
t0oF-azpn, line 576: Update the comment here

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
t0oF-azpn: jsObjectLayerPropsSupport ?

LeonidPolukhin: or useTypedArrays? To underline the main purpose of the flag.

typescript/packages/subsurface-viewer/src/layers/map/mapLayer.stories.tsx
LeonidPolukhin: Comment deleted.

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
LeonidPolukhin: Renamed.

nilscb (approved):

Issue #1895, opened 2024-01-30
Stories with layers exposed as controls will crash if layers are instantiated with new (as opposed to json)


nilscb:
Prevent storybook from copying layers created with "new" because the layer will loose its prototype (superclass) in this case. It will then not be possible to call "clone" on it which is implemented in the superclass (Layer class).

t0oF-azpn:
Fixed by #1896

PR #1896, opened 2024-01-30
test: Stories with layers exposed as controls will crash if layers are instantiated with new (as opposed to json) #1895

nilscb:
- Prevent storybook from copying layers created with "new" because the layer will loose its prototype (superclass) in this case. It will then not be possible to call "clone" on it which is implemented in the superclass (Layer class).

- Also fixed a missing "ZIncreasingDownwards" flag.


t0oF-azpn:
> Doesn't strictly deserve the "fix" prefix in the PR title, since it only affects testing or documentation.

I completely removed the Map3DLayer: even if it was not  working any more (nothing gets displayed), client code can still depend on it...

typescript/packages/subsurface-viewer/src/storybook/sharedSettings.tsx
```diff
@@ -98,6 +98,7 @@ export const volveWellsWithLogsLayer = {
     logrunName: "BLOCKING",
     logName: "ZONELOG",
     logColor: "Stratigraphy",
+    ZIncreasingDownwards: false,
```
t0oF-azpn, line 101: If this is missing here, it should also miss above (ie. for all the layers using ./volve_wells.json, right ?

t0oF-azpn (approved):

t0oF-azpn:
Hi, can you find in all the stories where this pattern happens (ie. layers created using new in the story args) to fix these ?

nilscb:
I have checked that none of them crashes,

t0oF-azpn (approved):

hkfb (approved):

PR #1899, opened 2024-01-30
fix: Crash in MapLayer - "Contour Lines" story

nilscb:


typescript/packages/subsurface-viewer/src/layers/map/mapLayer.stories.tsx
```diff
@@ -830,7 +830,7 @@ const ContourLinesComponent: React.FC<ContourLinesComponentProps> = (
 export const ContourLines: StoryObj<typeof ContourLinesComponent> = {
     args: {
         syncViewports: true,
-        show3d: true,
+        show3d: false,
```
t0oF-azpn, line 833: Why changing this ?

nilscb: I believe it originally was 2D and I don't see a point in having it 3D but I can change it if you will.

t0oF-azpn: No, it's fine.

t0oF-azpn (approved):

PR #1900, opened 2024-01-30
fix: mapLayer web worker crashes if no property data provded.

LeonidPolukhin:
The crash if fixed.
Some mapLayer code refactored.


nilscb (approved):

typescript/packages/subsurface-viewer/src/layers/map/mapLayer.ts
```diff
@@ -43,14 +43,70 @@ type Frame = {
...
-    propertiesData: Float32Array,
+    meshData: Float32Array | null,
+    propertiesData: Float32Array | null,
     isMesh: boolean,
     frame: Frame,
     smoothShading: boolean,
     gridLines: boolean,
 ];
 
+async function loadURLData(url: string): Promise<Float32Array | null> {
```
t0oF-azpn, line 54: Shouldn't these function be shared and used in every place where data arrays are loaded ?

LeonidPolukhin: Yes, they should be but it's more preferable to refactor in a dedicated PR rather than in this bug fixing one.

PR #1904, opened 2024-02-02
feat: Added property "formatLabelFunc" to set a callback funtion for formating label numbers

nilscb:
- also in the case that axis labels are smaller if displayed in exponential format this will be done.


typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -460,15 +489,35 @@ export default class Axes2DLayer extends Layer<Axes2DLayerProps> {
...
-        const mh = this.props.marginH * pixel2world;
-        const mv = this.props.marginV * pixel2world;
-
         const viewport_bounds_w = this.context.viewport.getBounds(); //bounds in world coordinates.
         const xBoundsMin = viewport_bounds_w[0];
         const xBoundsMax = viewport_bounds_w[2];
         const yBoundsMin = viewport_bounds_w[1];
         const yBoundsMax = viewport_bounds_w[3];
 
+        const [useExponential, useFracDigits, maxLength] = UseExponential(
```
hkfb, line 498: Are you deciding whether to use scientific format or not depending on the available space?

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -172,17 +202,18 @@ export default class Axes2DLayer extends Layer<Axes2DLayerProps> {
         min: number,
         max: number,
         viewSide: ViewSide,
-        pixel2world: number
+        pixel2world: number,
+        mv: number,
+        mh: number,
+        useExponential: boolean,
```
hkfb, line 208: I think it would be better to be able to override the number formatting, eg. with a callback. Say a callback that takes a number and returns a formatted number as a string. This way we could ensure consistent formatting in the map view, readout, and other representations of the data. A good default could be to use [d3-format](https://github.com/d3/d3-format)

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
nilscb: No. Its decided by the size of the numbers. Huge numbers will be in scientific. But they have to be quite big as sceitific also takes up quite a few characetrs (ex: -1.23e12 = 8 chearacters.)

hkfb:
I think it would be good to override the number formatting to allow consistency, eg. with other data views and the readout. But I'll reset my vote and let you decide.

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -77,8 +107,8 @@ const defaultProps = {
     id: "axes2d-layer",
     visible: true,
     coordinateSystem: COORDINATE_SYSTEM.CARTESIAN,
-    marginH: 80, // Horizontal margin (in pixles)
-    marginV: 30, // Vertical margin (in pixles)
+    marginH: undefined, // Horizontal margin (in pixles)
```
t0oF-azpn, line 110: Does not match the definition line 91/92.

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -53,9 +53,39 @@ enum ViewSide {
     Top,
 }
 
+function MakeExpLabel(a: number): string {
+    return a.toExponential(2).replace("e", "E");
+}
+
+function UseExponential(a: number, b: number): [boolean, boolean, number] {
```
t0oF-azpn, line 60: I'm not a fan to rewrite the wheel.
Can't we use d3 format, or some standard lib ?

t0oF-azpn (changes requested):

t0oF-azpn:
Playing with your stories, if I miltiply the bounds Y values by 10, I get the following which is not helpfull for the user:
image

typescript/package.json
```diff
@@ -129,5 +129,8 @@
     },
     "workspaces": [
         "packages/*"
-    ]
+    ],
```
hkfb, line 132: Please install this dependency in the `subsurface-viewer` package instead.

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -53,13 +54,57 @@ enum ViewSide {
...
 }
 
+function MakeExpLabel(a: number): string {
+    let label = d3.format(".2e")(a);
+    label = label.replace("e", "E");
+    label = label.replace("\u2212", "-");
+    return label;
+}
+
+function UseExponential(
```
hkfb, line 64: I don't think it's useful to condition notation on the space it takes, but rather a fixed rule, eg. scientific notation if number of digits > 6.

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -53,13 +54,57 @@ enum ViewSide {
...
 const zDepthAxesTicksAndLabels = zDepthAxesAxis + 1;
+const tickLineLength = 10;
+const pixelScale = 8;
 
 export interface Axes2DLayerProps extends ExtendedLayerProps {
-    marginH: number;
-    marginV: number;
+    marginH: number | undefined;
+    marginV: number | undefined;
+    formatLabelFunc: (x: number) => string | undefined;
```
hkfb, line 107: Good that we have the ability to override. As long as we have that, maybe we don't need to change the default?

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
nilscb: I changed it to a fixed rule but so that utm coordinates shows up in non-exponential (digits > 7). Otherwise there may be several ticks with the same exponential number.

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
nilscb: I think the option to not define the margins is nice so that the margins get auto calculated? 
That happen if one uses the formatLabelfunc or not.

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
hkfb: Ok, makes sense. The `UseExponential` still seems a bit magic to me. Maybe there should be some unit tests to cover it?

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
hkfb: I agree that the automatic margins are useful. I just wonder if we need to implement a new default label formatting behaviour if we can instead stick with the current default behaviour and override it instead? You can resolve.

hkfb:
@nilscb we added screenshot comparison for stories, so you might need to update the screenshots. See https://github.com/equinor/webviz-subsurface-components/tree/master/typescript#smoke-tests

hkfb:
@nilscb see the last two commits. The x axis ticks are appearing and disappearing randomly. Any idea why?

nilscb:
This pull request enables possibility to specify callback function to format labels on the axes. (See FormatLabelFunction story)
Also fixes missing tick marks (hopefully in all situations)

typescript/packages/subsurface-viewer/src/storybook/layers/Axes2DLayer.stories.tsx
```diff
@@ -49,6 +49,33 @@ export const Base: StoryObj<typeof SubsurfaceViewer> = {
     },
 };
 
+function MakeLabel(a: number): string {
```
hkfb, line 52: Stick to conventional (camel?) case. Can we enforce this with an eslint rule?

typescript/packages/subsurface-viewer/src/storybook/layers/Axes2DLayer.stories.tsx
```diff
@@ -49,6 +49,33 @@ export const Base: StoryObj<typeof SubsurfaceViewer> = {
...
+    // Choos exponential format with 3 digits after point.
+    const label = d3.format(".3e")(a);
+    //const label = d3.format(".1f")(a) // fixed decimal  KEEP
```
hkfb, line 55: Can we remove the comment?

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -83,10 +87,12 @@ const defaultProps = {
     coordinateSystem: COORDINATE_SYSTEM.CARTESIAN,
     marginH: 80, // Horizontal margin (in pixles)
     marginV: 30, // Vertical margin (in pixles)
+    //formatLabelFunc: undefined,
     isLeftRuler: true,
     isRightRuler: false,
     isBottomRuler: true,
     isTopRuler: false,
+    //backgroundColor: undefined,
```
hkfb, line 95: Remove comments?

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/storybook/layers/Axes2DLayer.stories.tsx
nilscb: I renamed the function. I may be possible to enforce camel case with eslint. But perhaps this is better done in a separate PR as it may trigger cleaup all over the place.

typescript/packages/subsurface-viewer/src/storybook/layers/Axes2DLayer.stories.tsx
nilscb: I thought I would keep this one as otherwise one has to look up d3 documentation to test dofferent cases,

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
nilscb: Removed!

hkfb:
@nilscb if you merge from master now, are you able to update the snapshots?

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
```diff
@@ -144,8 +148,22 @@ const fontInfo = {
...
+        changeFlags,
+    }: UpdateParameters<this>): boolean {
+        return (
+            super.shouldUpdateState({
+                props,
+                oldProps,
+                context,
+                changeFlags,
+            }) ||
+            changeFlags.propsOrDataChanged ||
```
t0oF-azpn, line 164: Probably useless (there is no data and super.shouldUpdateState should handle props).

BTW, we should be consistent with other layers (in particular 3D axes layer)

t0oF-azpn (changes requested):

typescript/packages/subsurface-viewer/src/layers/axes2d/axes2DLayer.ts
nilscb: Yes that's true. They are the same now.

typescript/packages/subsurface-viewer/src/storybook/layers/__image_snapshots__/subsurfaceviewer-map-layer--big-map-with-hole.png
t0oF-azpn, line 1: This one seems strange to me :/

t0oF-azpn (approved):

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/storybook/examples/CameraControlExamples.stories.tsx
```diff
@@ -546,6 +546,7 @@ const ScaleYComponent = ({ verticalScale }: { verticalScale: number }) => {
                 {
                     id: "section",
                     verticalScale,
+                    zoom: 2,
```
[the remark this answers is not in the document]
hkfb: Not a big deal, but it looks quite small now, so it might be harder to see the effect of a verticalScale < 1

hkfb (approved):

PR #1905, opened 2024-02-02
feat: use webworker pool to control number of workers

t0oF-azpn:
Creating too many webworkers is bad and leads to out of memory (OOM) errors (more or less quickly, depending on the web bundle size).

This work is about moving to controller web worker pools, first for triangle layer.

It contains:

* reproducing the issue in storybook
  - Add story generating many TriangleLayers with 1 triangle each
    The goal is to force spanning may webworkers to reach an OOM issue.
    I got the OOM with 1200+ layers (note that client reached it with ~200 layers, probably of bigger web bundles)
  - Add story with 1 TriangleLayers displaying many triangles
    The goal is to check behavior if 1 triangle layer takes a lot of time
  - Add triggerHome to all the stories

* Fix endless rendering (in fact no rendering at all), thus never ending "Loading assets..."
  Issue revealed by the new story...

* Use a workerpool to limit the number of workers to 10
  BTW, this also achieves a good performance raise in the case of this extreme use case (each layer has only 1 triangle)
  Using https://github.com/josdejong/workerpool

### DIFF
--- a/python/webpack.config.js
+++ b/python/webpack.config.js
@@ -42,6 +42,11 @@ module.exports = function (env, argv) {
         externals,
         resolve: {
             extensions: [".ts", ".tsx", ".js", ".jsx", ".json"],
+            fallback: {
+                os: false,
+                child_process: false,
+                worker_threads: false,
+            },
         },
         devtool: "source-map",
         module: {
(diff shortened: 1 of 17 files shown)


hkfb (approved):
LGTM :+1:

PR #1907, opened 2024-02-05
fix: Well head position Z value should match the top of the well trajectory by default if not set explicitly

nilscb:


hkfb (approved):

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.ts
```diff
@@ -50,6 +48,17 @@ export function removeDuplicates(data: FeatureCollection): void {
 
         let coords = lineString.coordinates as Position3D[];
 
+        // If not defined set wellhead z value to top of well string.
+        const wellHead = geometryCollection?.geometries[0] as Point;
+        if (wellHead.coordinates && wellHead.coordinates.length === 2) {
+            wellHead.coordinates.push(coords[0][2]);
```
t0oF-azpn, line 54: I'm checking if we want to take the Z value only, or if we have to apply also the MD also.

t0oF-azpn: Z value is good enough for now

Issue #1888, opened 2024-01-25
WellLogViewer, Python - Prevent cropping of unit in readout


eirik-morken:
There is an issue where a long value in the readout is not scaled properly, leading to the readout extending outside the visible box.
![image](https://github.com/equinor/webviz-subsurface-components/assets/72235758/68958af3-b970-40c9-97e7-0b49b3ebbae9)

Recreated in storybook where MFOA gets too many numbers, here just the outer line disappears. If the combined length of number and unit is too long then the unit starts to disappear. 
![image](https://github.com/equinor/webviz-subsurface-components/assets/72235758/bffcf5fa-5677-4bfc-a406-63a494cefa0a)

This is always prevalent in my viewer. Making the box the viewer exists in larger does not help as the viewer just scales accordingly and same problem occurs.
Is this a bug that could be fixed or is there some workaround?

anders-kiaer:
Thanks for the report @eirik-morken. What would be the preferred general solution here? Limit readout to X significant digits?

eirik-morken:
It could be limit amount of digits, dynamically or manually have more flexibility to increase size of readout or a mix of both I guess.

Issue #1874, opened 2024-01-16
[NGRM] - Modify Legend display of discrete logs


mirisb:
These changes are particularly important for discrete logs when showing classes in the readout.
1. Currently in Legend display there is a varying font size for well log names in the legend readout. The request is to display all logs with the same font size and if the name of the log is too long, truncate it with ... 
2. It is important to show the full name of the class in the legend readout. There is also a request that if class contains multiple words it will not be broken down into multiple lines- keep the class in a single line.

Suggestions:
1. add tooltip to see full name of log if it is truncated
2. provide API to allow external font size (font type is defined in css)

Issue #1715, opened 2023-10-18
[NGRM] - In WellLogViewer, title text is not displayed correctly when component size is reduced


mirisb:
When reducing the size of the component, the titles of the tracks go up and get cropped from the top and barely visible:
![image](https://github.com/equinor/webviz-subsurface-components/assets/18436935/294405d3-b8f9-4f5f-8fce-44b71c5c5820)
If reducing the size some more, the font size of the titles is adjusted such that the titles are visible:
![image](https://github.com/equinor/webviz-subsurface-components/assets/18436935/26940d8e-b9f6-47d2-b2e5-29f595f8404b)

The font adjustments should be associated with the title width

mirisb:
This bug is within Webviz implementation of stackLegendConfig which is responsible for legend display of a stacked track.

Vladimir-Kokin:
The bug is not only in **stack** tracks but in **scale** track also.
![изображение](https://github.com/equinor/webviz-subsurface-components/assets/86053741/22fbd0bc-bc10-4c8c-a161-b2d7e0b3e35c)

mirisb:
Size of title is still not computed correctly - the string is now cut on the width axis. see storybook example: 
![image](https://github.com/equinor/webviz-subsurface-components/assets/18436935/c5e6756f-425f-48e7-beed-0c4db0e36a32)

Vladimir-Kokin:
Too long titles for this width.
What can be done?

mirisb:
If the user makes the track narrow we need to truncate the name (dynamically)

suggestions: 
1. truncate name and add ellipsis (...)
2. add tooltip to show full name when hovering on track title

PR #1941, opened 2024-02-29
feat: Track caption and readout font size and text truncation

Vladimir-Kokin:
- Refactor readout code, set HTML elements CSS props with CSS style when possible to allow it customization: font size, readout and readout elements widthes and so on
- Long track captions, readout names, values and units truncated with ellipsys
- Add tooltips to show not truncated text

See #1715, #1874, #1888


hkfb:
@Vladimir-Kokin you probably need to update the test snapshots. See https://github.com/equinor/webviz-subsurface-components/tree/master/typescript#smoke-tests. Let me know if you need help with running the docker scripts.

Vladimir-Kokin:
@hkfb I have no docker installed on my computer
My Windows do not want to be upgraded. It always undoes an upgrade to new version

hkfb (approved):

PR #1953, opened 2024-03-11
feat(grid3D):  Support discrete properties.

LeonidPolukhin:


typescript/packages/subsurface-viewer/src/storybook/layers/Grid3DLayer.stories.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
hkfb, line 1: Suggest not using this globally, but instead for each instance you want to disable.

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/storybook/layers/Grid3DLayer.stories.tsx
LeonidPolukhin: Ooops, fixed.

hkfb (approved):

hkfb:
I think the linting gets stuck. Any idea why?

PR #1978, opened 2024-03-20
fix: disable manipulation and fix matrix error before camera initialization

t0oF-azpn:
A non invertible matrix error throws when picking before valid camera initialization. This happens before a consistent bounding box could be provided (before any geometry is displayed before loading assets finished).

Changes:
- a valid default camera setting is used when missing data prevents to compute valid camera (widget size of 0, 3d bounding box not yet computed...)
- manipulation is prevented before the valid camera is computed


typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1535,7 +1552,7 @@ function buildDeckGlViews(views: ViewsType | undefined, size: Size): View[] {
         return [
             new OrthographicView({
                 id: "main",
-                controller: { doubleClickZoom: false },
+                controller: null,
```
nilscb, line 1555: Does the "double-click to move camera to point" functionality work after removing this?

nilscb (approved):

typescript/packages/subsurface-viewer/src/components/Map.tsx
t0oF-azpn: This is a dummy view when in case no valid view is provided.
controller = null means no interaction at all.

Whenever anything is displayed (ie. this view is not used), double click still works correctly ;)

typescript/packages/subsurface-viewer/src/storybook/examples/__image_snapshots__/subsurfaceviewer-examples-camera--add-layer.png
hkfb, line 1: Can you reproduce this?

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1591,32 +1608,101 @@ function buildDeckGlViews(views: ViewsType | undefined, size: Size): View[] {
     return result;
 }
 
+/**
+ * Returns true if the camera zoom is set.
+ * @param camera to be tested
+ * @returns true if the camera camera zoom is set.
+ */
+function cameraHasZoom(camera: ViewStateType | undefined): boolean {
+    return typeof camera?.zoom === "number" && !Number.isNaN(camera.zoom);
```
hkfb, line 1617: `zoom` can be `[number, number]` in case of vertical scaling in orthographic view.

t0oF-azpn: Too bad there is no doc reflecting it in `MapProps`, nor story :/
Btw, it seems that `computeCameraZoom` does not handle the zoom correctly...

Since this PR is to fix a critical bug, can we complete it and add a follow-up to handle the `[number, number]` case ?

hkfb (approved):

PR #1980, opened 2024-03-21
test: allow running docker tests on Windows git bash shell

hkfb:


typescript/package.json
```diff
@@ -23,8 +23,8 @@
         "storybook": "storybook dev -p 6006",
         "build-storybook": "storybook build",
         "storybook:test": "test-storybook",
-        "docker:build": "docker build --tag wsc_story_test - < Dockerfile.test",
-        "docker:storybook:test": "docker run -it --rm --mount type=bind,source=.,target=/test --network host --ipc=host wsc_story_test",
+        "docker:build": "docker build --tag wsc_story_test --file Dockerfile.test .",
+        "docker:storybook:test": "docker run -it --rm --mount type=bind,source=./packages,target=/test/packages --network host --ipc=host wsc_story_test",
```
t0oF-azpn, line 27: Bind fails on windows with bash shell:

docker: Error response from daemon: invalid mount config for type "bind": invalid mount path: 'packages' mount path must be absolute.

t0oF-azpn (changes requested):

typescript/package.json
```diff
@@ -23,8 +23,8 @@
         "storybook": "storybook dev -p 6006",
         "build-storybook": "storybook build",
         "storybook:test": "test-storybook",
-        "docker:build": "docker build --tag wsc_story_test - < Dockerfile.test",
-        "docker:storybook:test": "docker run -it --rm --mount type=bind,source=.,target=/test --network host --ipc=host wsc_story_test",
+        "docker:build": "docker build --tag wsc_story_test --file Dockerfile.test .",
+        "docker:storybook:test": "docker run -it --rm --mount type=bind,source=$(pwd)/packages,target=/test/packages --network host --ipc=host wsc_story_test",
```
t0oF-azpn, line 27: $(pwd) does not work on windows with git bash shell :/

t0oF-azpn (changes requested):

typescript/package.json
hkfb: What error do you get?

Does it work if you replace `$(pwd)` with `%cd%`?

typescript/package.json
```diff
@@ -22,10 +22,11 @@
         "validate": "npm run typecheck && npm run lint",
         "storybook": "storybook dev -p 6006",
         "build-storybook": "storybook build",
-        "storybook:test": "test-storybook",
-        "docker:build": "docker build --tag wsc_story_test - < Dockerfile.test",
-        "docker:storybook:test": "docker run -it --rm --mount type=bind,source=.,target=/test --network host --ipc=host wsc_story_test",
-        "docker:storybook:test:update": "npm run docker:storybook:test -- npm run storybook:test -- --updateSnapshot",
+        "storybook:test": "test-storybook --no-index-json",
+        "docker:build": "docker build --tag wsc_story_test --file Dockerfile.test .",
+        "docker:storybook:test": "cross-env-shell CWD=$PWD npm run docker:storybook:test:docker:core -- -it --rm --mount type=bind,source=${CWD}/packages,target=/test/packages --network host --ipc=host wsc_story_test",
```
hkfb, line 27: Does using `${PWD}` directly work?

typescript/package.json
t0oF-azpn: Yes, it works with %cd%.
Thus we can either solve the issue by using `cross-env` package, or by using linux vs. windows specific scripts...

typescript/package.json
t0oF-azpn: No, it does not work

typescript/package.json
```diff
@@ -22,10 +22,11 @@
...
+        "docker:storybook:test:update": "npm run docker:storybook:test -- npm run storybook:test -- --url http://host.docker.internal:6006/ --no-index-json --updateSnapshot",
```
t0oF-azpn, line 28: Shouldn't we use non default port number, to avoid the risk to interfere with other running storybooks ?

typescript/package.json
hkfb: Committed a rename from `CWD` to `PWD` for compatibility with Linux shell. Does it work for you?

typescript/package.json
hkfb: We could run both the storybook and the tests in the docker container. So you wouldn't have to manually run the storybook at all.

We could do something like
1. Build storybook static
2. Copy into docker image when the image is built
3. In the container, serve storybook and test in parallel as in the [typescript workflow](https://github.com/equinor/webviz-subsurface-components/blob/258638ab5a5d56c4a0ed9c4595417f0be31cd7f5/.github/workflows/typescript.yml#L143)

Should we follow this up in a separate PR?

t0oF-azpn: I'm in favor to have that in a separate PR.
The current PR bring snapshot updates to windows users, which is important.

t0oF-azpn (approved):

PR #2001, opened 2024-04-03
fix: allow vertical scaling when camera is defined

hkfb:
Also inverts zoom amplitude according to `verticalScale`

Also removes redundant png files from builds.


typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1689,10 +1689,11 @@ function updateViewState(
     camera: ViewStateType,
     boundingBox: BoundingBox3D,
     size: Size,
-    is3D = true
+    is3D = true,
+    verticalScale = 1
 ): ViewStateType {
     if (isCameraDefined(camera)) {
-        return camera;
+        return scaleCameraZoom(camera, verticalScale, is3D);
```
t0oF-azpn, line 1696: I'm a bit concerned about the impact on memos.
I fear that returning a modified version of the camera here will force re-rendering... :/

I tested it locally, and /story/subsurfaceviewer-examples-camera--synced-subsurface-viewers seems to be broken (at least compared to published storybook).

nilscb:
The 2D Axes are messed up for some reason when changing the controls.

hkfb:
@t0oF-azpn @nilscb perhaps it's better to explicitly set [x, y] zoom directly on the `camera` prop. We need to be able to store / restore the camera, so we need to expose that state.

PR #2002, opened 2024-04-03
feat: Allow camera xy scaling

hkfb:
Followup to https://github.com/equinor/webviz-subsurface-components/pull/2001

Also inverts amplitude of vertical scaling

Also removes unused png files from builds

Note that axis rendering instability persists. I have not found a solution, but we can follow it up.


hkfb:
@nilscb What's interesting is that the axes are stable if `cameraPosition.target` is set to eg. `[1, 1, 1]`, but unstable when set to `[0, 0, 0]`

nilscb:
Ok that is a good observation. I guess target (what the camera looks at) should be set to x,y = middle of data and z some negative value(?) I can look at t further after this PR.

nilscb (approved):

PR #2012, opened 2024-04-09
feat(grid3d): discrete property names are displayed in the readouts.

LeonidPolukhin:
1) prop **propertyNames** of string[] type is added to Grid3DLayer.
2) The prop is used to display discrete property names in the readout.


typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
```diff
@@ -146,6 +146,12 @@ export interface Grid3DLayerProps extends ExtendedLayerProps {
      */
     propertiesData: string | number[] | Float32Array | Uint16Array;
 
+    /**
+     * Discrete propety names to be displayed in cursor readouts.
+     * The property values are used as property name indices.
+     */
+    propertyNames?: string[];
```
t0oF-azpn, line 153: only valid for discrete properties => discretePropertyNames ? discretePropertyValueNames ? discretePropValueNames ?

t0oF-azpn (changes requested):

typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
LeonidPolukhin: renamed to discretePropertyNames

t0oF-azpn (approved):

PR #2013, opened 2024-04-09
fix: honor vertical scale prop in 3D

hkfb:
Note that only the first `verticalScale` value is taken into account. To handle viewport dependent scaling we can look into creating a custom view following [this discussion](https://github.com/visgl/deck.gl/discussions/8675)


typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -504,8 +424,20 @@ const Map: React.FC<MapProps> = ({
...
+    const { zScale: zReScale, divRef: zScaleRef } = useHandleRescale();
+
+    const { shiftHeld, divRef: shiftHeldRef } = useShiftHeld();
+
+    const divRef = mergeRefs(
+        zScaleRef,
+        shiftHeldRef
+    ) as React.Ref<HTMLDivElement>;
+
+    const zScale = verticalScale ?? zReScale;
```
t0oF-azpn, line 440: Does that mean that if any viewport has a `verticalScale` set, the arrow keys are disabled ?
In that case, we probably want to disable the event handling, no ?

Does this allow the client application to set its own handler of key events ?

typescript/packages/subsurface-viewer/src/views/viewport.ts
```diff
@@ -0,0 +1,45 @@
...
+    rotationX?: number;
+    rotationOrbit?: number;
+
+    /** A vertical scale factor, used to scale items in the viewport vertically */
+    verticalScale?: number;
+
+    isSync?: boolean;
+}
+
+export const useVerticalScale = (viewport: ViewportType[] | undefined) => {
```
t0oF-azpn, line 38: Please, rename `viewport` to `viewports`.

So here we use the first found verticalScale from all the viewports, right ?

t0oF-azpn (changes requested):

typescript/packages/subsurface-viewer/src/components/Map.tsx
hkfb: > Does that mean that if any viewport has a `verticalScale` set, the arrow keys are disabled ? In that case, we probably want to disable the event handling, no ?

Yes, the arrow keys will be disabled. Will add a change to disable the events.

> 
> Does this allow the client application to set its own handler of key events ?

That was the plan initially, to add let `verticalScale` be something like `number | () => number` and, if not set, fall back to a default value like the current event handler hook. So probably in an upcoming PR.

typescript/packages/subsurface-viewer/src/views/viewport.ts
hkfb: > Please, rename `viewport` to `viewports`.

Done.
> 
> So here we use the first found verticalScale from all the viewports, right ?

Yes, will make an attempt to make `verticalScale` viewport dependent, like in 2D, in a subsequent PR, and then we can apply a `verticalScale` respectively to each viewport.

t0oF-azpn (approved):

PR #2016, opened 2024-04-15
feat: move vertical scale to SubsurfaceViewer props

hkfb:
Changing `verticalScale` in the `views` prop does not work, as it will reset the camera target when `view` changes.


t0oF-azpn (changes requested):

t0oF-azpn:
@hkfb I would rephrase the title: "move vertical scale to SubsurfaceViewer props"

t0oF-azpn (approved):

Issue #1898, opened 2024-01-30
[NGRM] - SyncLogView initial view for wells with different ranges is wrong if not supplying domain


mirisb:
In the storybook example: 
![image](https://github.com/equinor/webviz-subsurface-components/assets/18436935/41f1c5d4-20b7-479e-8747-5c5302c2b606)
If removing the domain prop and remounting the component, it is expected that the full common range would be visible and that the zoom will be 1. However, it seems that the component is adjusted to show the full range of the first well only and setting zoom according to it.  

Upon first mount, the controller zoom is being adjusted, however this function doesn't fix anything if this.prop.domain is not set, and so the zoom is not fixed.

PR #2022, opened 2024-04-18
fix: Initial view for wells with different ranges in SyncLogViewer

Vladimir-Kokin:
See #1898


w1nklr:
No code enhancement in the PR, except some changes in a story.

PR #2028, opened 2024-04-22
fix: Incorrect depth readout after vertical re-scaling

nilscb:


typescript/packages/subsurface-viewer/src/layers/map/privateMapLayer.ts
```diff
@@ -272,9 +272,15 @@ export default class PrivateMapLayer extends Layer<PrivateMapLayerProps> {
...
-                ? -info.coordinate[2]
-                : info.coordinate[2];
+            const zScale = this.props.modelMatrix
+                ? this.props.modelMatrix[10]
+                : 1;
+
+            const depth =
+                (this.props.ZIncreasingDownwards
+                    ? -info.coordinate[2]
+                    : info.coordinate[2]) / zScale;
```
hkfb, line 282: Can zScale be 0?

nilscb: No, but i can be arbitrarely small if one keeps on scaling down,

nilscb: I added a lower bound to it just to make sure.

t0oF-azpn (approved):

PR #2032, opened 2024-04-24
feat(grid3d): Discrete property value in grid readouts

LeonidPolukhin:
1) Grid3dLayer API takes an array of discrete property value-name pairs.
2) The array is used to display both discrete property names and values in readouts.
3) A grid story updated.
4) minimal npm version is bumped a little because of eslint-plugin requirements.


t0oF-azpn (approved):

typescript/packages/subsurface-viewer/src/layers/grid3d/privateGrid3dLayer.ts
```diff
@@ -22,7 +22,7 @@ import fsShader from "./fragment.fs.glsl";
 import vsLineShader from "./vertex_lines.glsl";
 import fsLineShader from "./fragment_lines.glsl";
 import { localPhongLighting, utilities } from "../shader_modules";
-import { TGrid3DColoringMode } from "./grid3dLayer";
+import { IDiscretePropertyValueName, TGrid3DColoringMode } from "./grid3dLayer";
```
hkfb, line 25: There is a warning here: `Imports "IDiscretePropertyValueName" are only used as type`

typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
```diff
@@ -147,10 +152,10 @@ export interface Grid3DLayerProps extends ExtendedLayerProps {
     propertiesData: string | number[] | Float32Array | Uint16Array;
 
     /**
-     * Discrete propety names to be displayed in cursor readouts.
-     * The property values are used as property name indices.
+     * Discrete propety value-name pairs to be displayed in cursor readouts.
+     * The property values are used as the array indices.
      */
-    discretePropertyNames?: string[];
+    discretePropertyValueNames?: IDiscretePropertyValueName[];
```
hkfb, line 158: Normally this is a breaking change, but maybe no-one else is using this layer?

typescript/packages/subsurface-viewer/src/layers/grid3d/privateGrid3dLayer.ts
LeonidPolukhin: Fixed. IDiscretePropertyValueName is imported as type

typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
LeonidPolukhin: Only Viz uses the prop as far as I know

PR #2033, opened 2024-04-24
feat: add section projection option to wells layer

hkfb:


nilscb (approved):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -254,33 +267,39 @@ export function getSize(
...
-        let data = this.props.data as unknown as FeatureCollection;
-        if (typeof data !== "undefined" && !isEqual(data, [])) {
-            if (this.props.ZIncreasingDownwards) {
-                data = invertPath(data);
-            }
+        let data = this.props.data as FeatureCollection<GeometryCollection>;
+        const refine = this.props.refine;
 
-            checkWells(data);
+        if (typeof data === "undefined" || isEqual(data, [])) {
```
t0oF-azpn, line 273: isn't `data == undefined` or `! data` good enough ?
I'm always wondering the added value of `typeof data === "undefined"` ...

t0oF-azpn (approved):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
hkfb: I agree. Fixed.

t0oF-azpn (approved):

PR #2061, opened 2024-05-27
chore: group some dependencies for dependabot

w1nklr:
The default behavior of dependabot is to generate one PR for each dependency that needs to be updated. This leads to multiple PRs for related dependencies, while related dependencies should be handled together. For example storybook and 7 @storybook/* packages have the same version number, and should be handled as a whole.

This PRs groups some dependencies:
- react
- deckgl
- storybook

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,30 @@ updates:
     commit-message:
       prefix: fix
       prefix-development: chore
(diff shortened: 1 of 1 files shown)


anders-kiaer (approved):
LGTM 👍 I also don't see any way of testing this without merging to main branch.

PR #2087, opened 2024-06-05
fix: Smooth SLV scrolling

### LINKED ISSUE #2035: [NGRM] - Scrolling is erratic when flattening in sync log viewer
Scrolling in sync log view behaves erratically (where it should be smoothed) if the wells are flattened according to well pick. This can also be observed in internal Webviz storybook: 
https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/welllogviewer-demo-synclogviewer--default&args=syncTrackPos:!false;syncContentSelection:!false

Vladimir-Kokin:
Make scrolling with a scroller to be smooth

Closes #2035


typescript/packages/well-log-viewer/src/SyncLogViewer.tsx
```diff
@@ -540,6 +540,7 @@ class SyncLogViewer extends Component<SyncLogViewerProps, State> {
 
         if (this._isMounted) this.setState({ primaryAxis: value });
     }
+    /*
```
mirisb, line 543: If the code is commented out, can it be removed as to not pollute the code?

mirisb (approved):

typescript/packages/well-log-viewer/src/SyncLogViewer.tsx
Vladimir-Kokin: I have removed it

Vladimir-Kokin:
Storybook Smoke Tests are unstable
I had need to rerun to make they passed

PR #2153, opened 2024-07-11
fix: div size in the RTC Tab Layout story

### LINKED ISSUE #2015: [NGRM] - SLV layout conflicts with MUI Grid
Since WLV 1.7.0 which introduced ViewerLayout, there is a bug when the SyncLogViewer is within a MUI Grid container. 
The bug (probably related to Flex) causes the track to resize and grow infinitely until the page hangs.
Recording2024-04-11123645

This is a serious issue for clients who use MUI Grid for their layout and prevents them from upgrading to this version.

I am attaching a storybook file that can be plugged into webviz storybook to recreate the problem:
[example_story.zip](https://github.com/equinor/webviz-subsurface-components/files/14970810/example_story.zip)

Vladimir-Kokin:
Close  #2015


typescript/packages/well-log-viewer/src/Storybook/helpers/MuiComponentsTabHelper.tsx
```diff
@@ -66,7 +66,11 @@ const TestComponentArea: React.FC<React.PropsWithChildren> = ({ children }) => {
     const tab = useSelector(getTabValue);
 
     const renderer = (ref: React.MutableRefObject<null>): React.JSX.Element => {
-        return <div ref={ref}> {children} </div>;
+        return (
```
w1nklr, line 69: Here you apply a workaround at storybook level.
This will not help to fix the bug in applications (which are not using this code)

w1nklr (changes requested):

typescript/packages/well-log-viewer/src/Storybook/helpers/MuiComponentsTabHelper.tsx
Vladimir-Kokin: As I understand this code was writeen to be similar to problematic application code.
So a corresponding modification could be applied in it also.

w1nklr: The goal is not to provide an example of a workaround.
The goal is to fix this buggy behavior.

Vladimir-Kokin:
@w1nklr @nilscb Now I have reverted changes in the story but add some modification to component's CSS

w1nklr (approved):

mirisb (approved):

Issue #2151, opened 2024-07-11
[NGRM] - In WLV, need to reset twice to go back to initial prop


mirisb:
In webviz storybook example : https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/welllogviewer-demo-synclogviewer--custom-layout&args=domain[0]:1000;domain[1]:3000
If I have a domain set up then I need to press twice the reset button to get to the initial display.

mirisb:
Bug is still re-creatable in Viz storybook when a domain is applied.

PR #2173, opened 2024-07-26
fix: need to reset twice to go back (#2151)

Denis-Shamakin:
When the reset button is pressed, the default zoom is set. This causes the scrolling to change and triggers the onScroll event, which sets the overall zoom. This problem was solved by setting the "isDefZoom" flag, which is set to "true" when setting the default scale and does not allow the onScroll event to change the zoom


w1nklr (approved):

mirisb:
not solving the issue

Issue #2054, opened 2024-05-23
[NGRM] - In SLV, provide the option to color log by color function


mirisb:
Currently logs may be colored by solid color or color map (for area coloring). We would like to be able to supply a color function to color the log in addition to the possibility of using color maps.
This is similar to the ability that exists for property maps that allows the map to be colored either by color map or by a color function.

mirisb:
The current implementation overrides the colorTables in a global way, thus adding a function overrides the possibility to have some tracks use regular color tables. 

We would like this implementation to change such that the color function would be provided by TemplatePlotProp. This template props can be applied to a single track. this would allow us to have a different function for different plots\tracks and not one single function at the root.

PR #2191, opened 2024-08-20
test: Color by function groundwork for test

mirisb:
Add a story with required API for supplying color function per plot
Adjust a specific deep copy function for templates to handle copy of functions.


mirisb:
this relates to #2054

typescript/packages/well-log-viewer/src/WellLogViewer.stories.tsx
```diff
@@ -291,6 +291,56 @@ export const Default: StoryObj<typeof StoryTemplate> = {
     render: (args) => <StoryTemplate {...args} />,
 };
 
+export const ColorFunc: StoryObj<typeof StoryTemplate> = {
```
hkfb, line 294: ATM this is not really a color function example, so maybe it can be named as such when it becomes a real example (if I understand the intention correctly)

typescript/packages/well-log-viewer/src/utils/deepcopy.ts
```diff
@@ -1,3 +1,26 @@
+import type { Template } from "../components/WellLogTemplateTypes";
+
 export function deepCopy<T>(source: T): T {
     return JSON.parse(JSON.stringify(source)) as T;
 }
+
+export function deepCopyTemplate(source: Template): Template {
```
hkfb, line 7: This function is more than a deep copy, so maybe it needs a more descriptive name and perhaps a description.

typescript/packages/well-log-viewer/src/WellLogViewer.stories.tsx
mirisb: The color function doesn't work so I commented it. I could add TBD (to be done) to the story name so that for now storybook users will see this example doesn't work yet

typescript/packages/well-log-viewer/src/utils/deepcopy.ts
mirisb: It is exactly deep copy but with addition of specific functions which are not copied over (ever, I researched several methods), I will add a comment. How do you suggest I change the name so that it will be more clear?

hkfb: Ok, got it

hkfb (approved):

PR #2218, opened 2024-09-02
feat: Add hook for vertical scaling

hkfb:
Allows controlling vertical scaling factor through a combination of state and events. For example, this allows storing and restoring a scaling factor that has been altered by keyboard events between sessions.

To help with this a utility hook is added and an `innerRef` is added in order to propagate events.

A story has been added to test the keyboard interaction provided by the new hook.


w1nklr:
can you please add geojson as a subsurface-viewer dependency ?
This package is used in runtime code, but the dependency is missing (it worked before updating Deck.gl to version 9, because geojson was pulled by deck.gl v8 dependencies...)
Or do you prefer doing it in a separate PR ?

hkfb:
> can you please add geojson as a subsurface-viewer dependency ?
> This package is used in runtime code, but the dependency is missing (it worked before updating Deck.gl to version 9, because geojson was pulled by deck.gl v8 dependencies...)

Sure, I'm also wondering if it makes sense to expose 3rd party types that are required implicitly (eg. geojson types such as `Feature`) through the `subsurface-viewer` package? What is the convention here?

hkfb:
> Or do you prefer doing it in a separate PR ?

Will create a separate PR.

w1nklr (approved):

nilscb (approved):

PR #2224, opened 2024-09-05
fix(grid3d): Fixed coloring if no property data proived

LeonidPolukhin:
1) Fixed case when no property data is given to Grid3DLayer coloring by undefined property is performed.
2) Fixed precision issue with undefined property value in case of continuous property.


w1nklr (approved):

LeonidPolukhin:
> LGTM, but why wasn't this bug caught by the tests? Would it make sense to create a new story for this specific case?

The existing examples are not that impressive being displayed in a constant color. Perhaps it would be nice to make a grid with complicated geometry.

PR #2244, opened 2024-09-13
fix: reloading page fails to reset the camera target correctly if Z s…

w1nklr:
…cale factor is provided

Z scale factor is not applied to the camera target while it is applied to point transformation when displaying objects.

### DIFF
--- a/typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
+++ b/typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
@@ -9,6 +9,7 @@ import type {
     BoundsAccessor,
     MapMouseEvent,
     MapProps,
+    Point3D,
     TooltipCallback,
     ViewStateType,
     ViewsType,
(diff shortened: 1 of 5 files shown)


typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
```diff
@@ -41,15 +42,15 @@ export type LightsType = {
     pointLights?: [
         {
             intensity: number;
-            position: [number, number, number];
+            position: Point3D;
```
hkfb, line 45: I wonder if the changed types in the API is causing `dash-generate-components` to fail? Have a look at `SubsurfaceViewer.propTypes`, maybe it needs to be declared?

hkfb (approved):
LGTM :+1:

typescript/packages/subsurface-viewer/src/SubsurfaceViewer.tsx
rubenthoms: I'm afraid we'd still like to have `propTypes` due to it checking prop types during runtime and our props coming in from Dash/Python via JSON.

PR #2253, opened 2024-09-20
fix: color of undefined value for 3D grids is incorrect

w1nklr:
The color was way to clear, and was missing the 0-255 to 0-1 conversion.


typescript/packages/subsurface-viewer/src/layers/grid3d/grid3dLayer.ts
```diff
@@ -430,7 +430,7 @@ export default class Grid3DLayer extends CompositeLayer<Grid3DLayerProps> {
         ) {
             return [colorFunc[0] / 255, colorFunc[1] / 255, colorFunc[2] / 255];
```
fbosquet-azpn, line 431: why do we divide by 255 here.

fbosquet-azpn (approved):

mirisb (approved):

Issue #2240, opened 2024-09-11
[NGRM]: zooming using mouse scroll wheel is limited


w1nklr:
Zooming using mouse scroll wheel is limited.

Using storybook, SubSurfaceViewer/Examples/Camera/ScaleY, when zooming in, the zoom can not exceed the following:

![image](https://github.com/user-attachments/assets/cf15a6b9-5a2a-4a6e-bd63-d76e4e653fcc)

PR #2260, opened 2024-09-23
fix: "[NGRM]: zooming using mouse scroll wheel is limited #2240"

nilscb:
Will fix limited zoom in 2D and also missing ticks on the ruler top left corner.

For large UTM like numbers there is some precision issues in 2D . Hence will limit zoom somewhat in this case. Is a bit hacky but seems to work fine.

A better possible future solution would be to translate all data to around origo. For readout the numbers would have to be translated back. This is similar to how RMS does it.


hkfb:
> For large UTM like numbers there is some precision issues in 2D . Hence will limit zoom somewhat in this case. Is a bit hacky but seems to work fine.
> 
> A better possible future solution would be to translate all data to around origo. For readout the numbers would have to be translated back. This is similar to how RMS does it.

The Layer class supports offset coordinates with a `coordinateOrigin`. We should aim to support it in all layers.

typescript/packages/subsurface-viewer/src/storybook/examples/__image_snapshots__/subsurfaceviewer-examples-camera--scale-y-with-camera-position.png
hkfb, line 1: Looks like some previously missing tick marks and labels are now appearing 👍

hkfb (approved):

w1nklr (approved):

PR #2292, opened 2024-10-04
fix: camera vertical scale issue

w1nklr:
Fix vertical scale not being taken into account on some camera manipulations.

The camera controller now explicitly separates view states sent to DeckGL (containing the vertical scaling) from the ones manipulated by the user (without vertical scale).
Fix target being stuck to Z = 0 when the vertical scale becomes 0.
Update storybook example rotating the camera.


typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1094,20 +1095,20 @@ class ViewController {
...
 
-    // returns the DeckGL views state(s) (ie. camera settings applied to individual views)
-    private readonly getDeckGlViewState = (
+    /**
+     * Returns the scaled DeckGL view state(s) (ie. camera settings applied to individual views)
+     * and the corresponding view states in user space (ie. not scaled by the zScale)
+     * based on the input views and current state.
+     * @param views requested views.
+     * @param state current state.
+     * @returns
```
hkfb, line 1104: Missing return value documentation

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1634,11 +1713,71 @@ function applyZScale(
...
+        target[1] === unscaledTarget?.[1]
+    ) {
+        return [
+            target[0],
+            target[1],
+            target[2] ? target[2] : unscaledTarget[2],
+        ];
+    }
+    return [target[0], target[1], target[2]];
+}
```
hkfb, line 1779: Would be good with some unit tests. YCR

w1nklr: It will be a good try for copilot.
I'll do it in a separate PR.

PR #2364, opened 2024-11-13
feat: Exposing reference to deck.gl instance as `deckGlRef` prop in `Map` component

rubenthoms:
In order to be able to use some of the deck.gl features (e.g. picking) outside of the actual `SubsurfaceViewer` component, it is necessary to have access to the deck.gl instance. Our specific use case is to implement simultaneous multi-view-picking in a custom implementation as deck.gl does not support this feature yet (and does not seem eager to implement it, see: https://github.com/visgl/deck.gl/discussions/9210). This could at some point become a feature in the `subsurface-viewer` component itself.


hkfb:
@rubenthoms Could you add some background for this change?

hkfb:
> Our specific use case is to implement simultaneous multi-view-picking in a custom implementation

Would it make sense to add a storybook example demonstrating this?

rubenthoms:
I can create an example, yes.

rubenthoms:
> > Our specific use case is to implement simultaneous multi-view-picking in a custom implementation
> 
> Would it make sense to add a storybook example demonstrating this?

I added a storybook example called "Multi View Picking".

typescript/packages/subsurface-viewer/src/storybook/examples/MultiViewExamples.stories.tsx
```diff
@@ -80,6 +81,326 @@ export const MultiViewAnnotation: StoryObj<typeof SubsurfaceViewer> = {
...
+                                <div>Y:</div>
+                                <div>
+                                    {pickingInfoPerView["view_1"]?.y ?? "-"}
+                                </div>
+                                {pickingInfoPerView["view_1"]?.properties?.map(
+                                    (el, i) => (
+                                        <React.Fragment key={`${el.name}-${i}`}>
+                                            <div>{el.name}</div>
+                                            <div>{el.value}</div>
+                                        </React.Fragment>
```
hkfb, line 354: Suggest truncating the readout values to only a few decimals

hkfb:
> > > Our specific use case is to implement simultaneous multi-view-picking in a custom implementation
> > 
> > 
> > Would it make sense to add a storybook example demonstrating this?
> 
> I added a storybook example called "Multi View Picking".

Thanks for the story, it's a nice feature!

typescript/packages/subsurface-viewer/src/storybook/examples/MultiViewExamples.stories.tsx
```diff
@@ -80,6 +81,326 @@ export const MultiViewAnnotation: StoryObj<typeof SubsurfaceViewer> = {
...
+    value: number;
+    color?: string;
+};
+
+type PickingInfoPerView = Record<
+    string,
+    {
+        x: number | null;
+        y: number | null;
+        properties: { name: string; value: number; color?: string }[];
```
w1nklr, line 97: Could use PickingInfoProperty here ?

typescript/packages/subsurface-viewer/src/storybook/examples/MultiViewExamples.stories.tsx
```diff
@@ -80,6 +83,308 @@ export const MultiViewAnnotation: StoryObj<typeof SubsurfaceViewer> = {
...
+                            pickingInfoPerView={pickingInfoPerView}
+                        />
+                    </View>
+                }
+            </SubsurfaceViewer>
+        </div>
+    );
+}
+
+export const MultiViewPicking: StoryObj<typeof SubsurfaceViewer> = {
```
hkfb, line 368: Consider adding a `play` function that moves the mouse cursor on top of one of the maps. This should give a more useful story snapshot for catching regressions.

See an example `play` function here:
https://github.com/equinor/webviz-subsurface-components/blob/d7ae8dbecf9e4134aa55fef4fe35030258ed9491/typescript/packages/subsurface-viewer/src/storybook/examples/CameraControlExamples.stories.tsx#L722

rubenthoms: I implemented a `play` function that moves the cursor to the left-top corner and then to the center of the first viewport.

hkfb:
You'll need to update the snapshots. See https://github.com/equinor/webviz-subsurface-components/tree/master/typescript#smoke-tests

w1nklr (approved):

Issue #2432, opened 2025-02-03
well layer: wellNameAtTop control is inverted


w1nklr:
Using the published storybook:

wellNameAtTop: true
![Image](https://github.com/user-attachments/assets/03546cc4-698f-47ae-a609-3149e02af2bd)

wellNameAtTop : false

![Image](https://github.com/user-attachments/assets/b2765975-3ea3-4d57-86a0-01a0ec74b556)

PR #2433, opened 2025-02-03
fix: well layer: enhance well name display to avoid cluttering

nilscb:
- Simplified some code
- Make sure label stays visible when camera changes.
- _fix bug "well layer: wellNameAtTop control is inverted #2432"_ already done by #2434


typescript/packages/subsurface-viewer/src/storybook/layers/WellsLayer.stories.tsx
```diff
@@ -769,6 +834,73 @@ export const ReducedWellNameClutter2D: StoryObj<
...
+                defaultValue={1}
+                step={1}
+                onChange={handleChange}
+                valueLabelDisplay={"auto"}
+            />
+        </Root>
+    );
+};
+
+export const SyntheticWells: StoryObj<typeof SyntheticWellsComponent> = {
```
w1nklr, line 881: Rename to explain the goal of the example (end user does not care that the wells are synthetic, but that it is a well trajectory refinement example).

typescript/packages/subsurface-viewer/src/storybook/layers/WellsLayer.stories.tsx
```diff
@@ -769,6 +834,73 @@ export const ReducedWellNameClutter2D: StoryObj<
     render: (args) => <ReducedWellNameClutterComponent2D {...args} />,
 };
 
+const SyntheticWellsComponent: React.FC<SubsurfaceViewerProps> = (props) => {
```
w1nklr, line 837: Use a name explaining the reason of the component, not how one of its data is built.

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -417,50 +417,30 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
...
+        // using z=0 for orthographic view to keep label above other other layers
+        if (pos) {
+            return view_is_3d ? pos : [pos[0], pos[1], 0];
         }
         return null;
     }
 
-    getTrajMidPoint(
+    // Return angle and position a given percentage down the trajectory.
+    getTrajPontAtPercentage(
```
w1nklr, line 443: => getTrajPo*i*ntAtPercentage

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -472,33 +452,50 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
         }
 
         const well_xyz = getTrajectory(well_object, undefined);
+
+        const w = this.context.viewport.width;
+        const h = this.context.viewport.height;
+
+        const percentages = [50, 25, 75, 12.5, 87.5, 25, 75];
```
w1nklr, line 459: Are the duplicated 25 and 75 expected ?

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -472,33 +452,50 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
...
+
         const n = well_xyz?.length ?? 2;
         if (well_xyz && n >= 2) {
```
w1nklr, line 462: Add some comments to explain the algorithm (are a ref to original if it comes from a known algo)

w1nklr: BTW, I don't get the logic. If I understand correctly, if the algo fail with the provided percentage, a hard-coded list of percentages completely unrelated to the requested one is used.

w1nklr (changes requested):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
nilscb: Please wait a bit with this code review. The Wellslayer.stories file is included by mistake

nilscb: This is a regular sampling along the trajectory in search of a camera visible location. It is shuffled to minimize large jumps and to tend towards the middle.

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
nilscb: Fixed

nilscb:
The auto positioning of lables  is now default off and controlled by the well property "wellNameAutoPosition"

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -118,6 +118,10 @@ export interface WellsLayerProps extends ExtendedLayerProps {
      *  If given as a number between 0 and 100,  will place name at this percentage of trajectory from top.
      */
     wellNameAtTop: boolean | number;
+    /** It true label position will be auto calculated if possible to be inside view volume at all times.
+     *  default false.
+     */
+    wellNameAutoPosition: boolean;
```
w1nklr, line 124: I would rename that to adjustWellNamePosition

w1nklr (changes requested):

hkfb:
If I disable `hideOverlappingWellNames` then all labels disappear, which is unexpected:
image

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
hkfb: > What bothers me is that it does absolutely not take into account requested position.
> I would use an approach to shift away from the requested position , not to ignore it

I agree, it could be better usability if the labels shifted more gradually from the requested position instead of popping up at a random position.

hkfb (changes requested):
It also appears that if you provoke a 90 degree angle on the labels by rotating the camera, then the labels disappear.

hkfb:
@nilscb ok to close this PR, since it was superceded by #2487?

w1nklr:
> @nilscb ok to close this PR, since it was superceded by #2487?

Fine for me

PR #2440, opened 2025-02-12
fix: Fixed string type logs not being rendered

Anders2303:
The `tracks.ts/setupPlot` had an error that caused all curves `valuetype: "string"` to be skipped. Fixed it so that stacked curves once again support string values. Also updated the "Discrete" storybook example to include a string-value curve, so that future smoke-tests can notice this type of error.

I also tweaked some internals in tracks.ts while I was at it; renamed some internal functions and variables, and made sure that styles are applied before `setupPlot` is called (this is to ensure we have the full props when validating them). _Note: a side-effect of this is that `generateColor()` is called a little more often, so the smoke-tests screenshots have been updated to accommodate for new colors_


w1nklr (approved):

w1nklr:
Are the color changes in the images expected ?

Anders2303:
> Are the color changes in the images expected ?

>Note: a side-effect of this is that generateColor() is called a little more often, so the smoke-tests screenshots have been updated to accommodate for new colors
Yes

PR #2463, opened 2025-03-07
fix: "74736: PPA: well trajectory artifacts"

nilscb:
There are some artifacts when drawing wells with many nodes/legs.

This is caused by deck.gl not being able to handle to many nodes/legs in a well path. Also there is z-buffer fighting between two sub layers in the well layer implementation.

The solution is to resample and reduce number of nodes in a well path in such a manner that its shape is retained.
This is done automatically in the well layer except when the standard "dataTransform" property is set in which case the user is responsible of any transform done to the path via this property.

A storybook example will be provided with an example of "dataTransform" property usage.


typescript/packages/subsurface-viewer/src/storybook/layers/WellsLayer.stories.tsx
```diff
@@ -769,6 +776,117 @@ export const ReducedWellNameClutter2D: StoryObj<
...
+                    </ViewFooter>
+                </View>
+            }
+        </SubsurfaceViewer>
+    );
+};
+
+export const CoarseWellFactor: StoryObj<typeof CoarseWellFactorComponent> = {
+    args: {
+        coarseWellsToleranceFactor: 0.01,
```
hkfb, line 885: Suggestion: use `argTypes` to set this control to a slider

typescript/packages/subsurface-viewer/src/layers/utils/simplify.ts
hkfb, line 1: If the code is copied, then we should honor its license, which says:

```
   1. Redistributions of source code must retain the above copyright notice, this list of
      conditions and the following disclaimer.

   2. Redistributions in binary form must reproduce the above copyright notice, this list
      of conditions and the following disclaimer in the documentation and/or other materials
      provided with the distribution.
```

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -296,14 +296,14 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
 
         const coarseData = coarsenWells(data);
 
-        const doRefine =
-            typeof refine === "number" ? refine > 1 : (refine as boolean);
+        const isDataTransformSet = this.props.dataTransform !== null;
```
hkfb, line 299: Would it work to set the default value of `dataTransform` to coarsenWells instead?

typescript/packages/subsurface-viewer/src/layers/utils/simplify.ts
nilscb: Done

typescript/packages/subsurface-viewer/src/storybook/layers/WellsLayer.stories.tsx
nilscb: Made into slider.

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
nilscb: Not really as dataTransform may do anything but here we need a coarse version for the simplified rendering,

typescript/packages/subsurface-viewer/src/storybook/layers/WellsLayer.stories.tsx
```diff
@@ -865,18 +861,30 @@ const CoarseWellFactorComponent: React.FC<CoarseWellFactorProps> = (
...
+                </SubsurfaceViewer>
+            </div>
+            <Slider
+                min={0.1}
+                max={70}
+                defaultValue={0.1}
+                step={0.1}
+                onChange={handleChange}
+                valueLabelDisplay={"auto"}
+            />
```
hkfb, line 886: I'd suggest using [`argTypes`](https://storybook.js.org/docs/api/arg-types) on the story parameters instead. You can override the default control to a [`range`](https://storybook.js.org/docs/api/arg-types#controltype).

hkfb (changes requested):

w1nklr:
Any chance to start a new review ?
It's impossible to cherry pick in the 42 modified files what this PR is :/

hkfb (changes requested):
When I change the tolerance factor, then the camera resets

typescript/packages/subsurface-viewer/src/storybook/examples/__image_snapshots__/subsurfaceviewer-examples-mutiview--multi-view-picking.png
hkfb, line 1: This change is suspicious

hkfb (changes requested):

hkfb:
> When I change the tolerance factor, then the camera resets

Fixed in https://github.com/equinor/webviz-subsurface-components/pull/2463/commits/abf66d2536b91db61e187ad51f31796873b4c2a5

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.ts
```diff
@@ -329,30 +333,17 @@ export function coarsenWells(data_in: FeatureCollection): FeatureCollection {
...
-            lineString.coordinates =
-                coordsSimplified.coordinates as Position3D[];
+        const properties = data.features[well_no]
+            .properties as GeoJsonProperties;
+        if (properties) {
+            const mds = properties["md"][0];
+            const [newPoints, newMds] = simplify(
+                lineString.coordinates as Position3D[],
+                mds,
+                0.01
```
w1nklr, line 343: A comment for the meaning of this value ?

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
hkfb: Fixed in https://github.com/equinor/webviz-subsurface-components/pull/2463/commits/5f8947552a7bfdf82a1534d0a8f0fbe49168ddf4

typescript/packages/subsurface-viewer/src/layers/wells/utils/spline.ts
hkfb: I improved the documentation for the tolerance argument

typescript/packages/subsurface-viewer/src/storybook/examples/__image_snapshots__/subsurfaceviewer-examples-mutiview--multi-view-picking.png
hkfb: Fixed in https://github.com/equinor/webviz-subsurface-components/pull/2463/commits/8a210b5c09215ef668f6b4ba2b3cd1d6e7bac6f9

w1nklr (approved):

PR #2482, opened 2025-03-20
fix: Fixed wrong default fill-opacity value

Anders2303:
Fixed a small bug that caused area plots to render with zero opacity if no fill is specified


w1nklr (changes requested):
Please remove second commit of #2480.

w1nklr (approved):

PR #2488, opened 2025-03-24
feat: Add tooltip optional prop to track template

mirisb:
It is now possible to customize the tooltip that appear when mouse hovers on top of the title of the track. Instead of showing the title of the track in the tooltip it is now possible to supply a custom tooltip in the track template, that replaces the default text.

image


hkfb (approved):
An improvement to this change could be to create a specific story for the tooltip feature, using the `[play function](https://storybook.js.org/docs/writing-stories/play-function#writing-stories-with-the-play-function)` to trigger the tooltip.

Thus in the future if this feature breaks, then it should be caught by the snapshot tests.

typescript/packages/well-log-viewer/src/components/WellLogTemplateTypes.ts
```diff
@@ -47,6 +47,7 @@ export interface TemplatePlot extends TemplatePlotProps {
 
 export type TemplateTrack = {
     title?: string; // if it is missed then a name or description of the first plot is used (see makeTrackHeader())
+    tooltip?: string; // tooltip for mouse hover over title
```
w1nklr, line 50: might be renamed to titleTooltip for clarity.

w1nklr (changes requested):

w1nklr (approved):

typescript/packages/well-log-viewer/src/WellLogViewer.stories.tsx
```diff
@@ -265,6 +267,108 @@ export const ColorByFunction: StoryObj<typeof StoryTemplate> = {
...
+        await expect(titleDiv.title).toContain("example continuous");
+
+        if (titleDiv) {
+            await userEvent.pointer({
+                target: titleDiv,
+                node: titleDiv,
+                offset: 2,
+            });
+        }
+    },
```
hkfb, line 368: Getting this error when I open the story: 
image

hkfb (changes requested):

hkfb (approved):

PR #2491, opened 2025-03-27
fix: Subsurface viewer - Assorted fixes to wells layer and multi pick hook

Anders2303:
* Made the `useMultiViewPicking` hook able to return picks from multiple wells if hovering on overlapping wells
  * Generally makes it so the hook  is able to pick multiple items from the same layer, whereas before it would only take the first
  * Added a new multi-pick example with wells to show it
* Adds a more descriptive typing for the GeoJson feature we use in the WellsLayer data prop -- `WellFeatureCollection`.
  * Most notably, we now specify what should be in the properties object.
  * Made most utils take this new specified typing, and removed/clean up a buuunch of spots where we were just assuming validity and casting any-values
* Exposed or specified some extra types that made sense to export as well
* Updated the `colorMappingFunction` prop to match the `createColorMapFunction` from emerson table.
  * The only difference is that it now specifies a couple of extra (optional) params. We were actually using these params, so the typing should now reflect that 
* Refactored some messy code in the wells layer class


typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -84,7 +84,31 @@ function onDataLoad(
...
+
+export interface WellsPickInfo extends LayerPickInfo<WellFeature> {
+    featureType?: string;
+    logName: string;
+}
+
 export interface WellsLayerProps extends ExtendedLayerProps {
+    // String, binary and promise is handled internally in Deck.gl to load external data.
+    // It's expected that a correctly structured GeoJSON is loaded
+    data:
```
hkfb, line 106: Use the upstream deck.gl types instead, eg. `LayerDataSource<WellFeature>`?

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -318,18 +352,19 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
             ) ||
             !isEqual(props.refine, oldProps.refine);
         if (needs_reload) {
+            // ? Why are we looping back to initializeState here? Deck.gl life-cycle intends for
+            // ? this to only be called *once*, so this is a confusing choice. (@anders2303)
```
hkfb, line 356: Indeed, the path transformations in `initializeState` should be moved here as far as I understand.

hkfb (changes requested):

nilscb (approved):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
Anders2303: LayerDataSource does not match the typing used by the GeoJsonLayer, which instead supports any valid GeoJson. My thought was to keep it mostly as is, but limit JSON structure to only allow a root `FeatureCollection`, since that's what the data transforms were assuming. So I think I want to keep this as is, for now!
 
See the remark about alternative formats for the data prop [here](https://deck.gl/docs/api-reference/core/layer#basic-properties), and the GeoJsonLayer data description [here](https://deck.gl/docs/api-reference/layers/geojson-layer#data)

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -936,17 +979,14 @@ function getLogColor(
         const max = Math.max(...log_data);
         const max_delta = max - min;
         log_data.forEach((value) => {
+            const adjustedVal = (value - min) / max_delta;
+
             const rgb = colorMappingFunction
-                ? colorMappingFunction((value - min) / max_delta)
-                : rgbValues((value - min) / max_delta, logColor, colorTables);
-            rgbValues(value - min / max_delta, logColor, colorTables, isLog);
```
w1nklr, line 942: Good catch.
you can resolve

w1nklr (changes requested):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
Anders2303: Moved

typescript/packages/subsurface-viewer/src/hooks/useMultiViewPicking/MultiViewPickingInfoAssembler.ts
```diff
@@ -231,11 +230,11 @@ export class MultiViewPickingInfoAssembler {
 
                     collectedPickingInfo[viewport.id] = {
                         coordinates: worldCoordinate,
-                        layerPickingInfo: collectedLayerPickingInfo,
+                        layerPickingInfo: Object.values(layerInfoDict),
                     };
                 } else {
                     collectedPickingInfo[viewport.id] = {
-                        coordinates: null,
+                        coordinates: worldCoordinate,
```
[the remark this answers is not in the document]
Anders2303: We give world-coordinates in all other readouts, even if the user is not hovering over anything; this is intended to mimic that behavior, for consistency. I'd wanna keep this as is for that reason.
Following, we might want a larger discussion for showing coordinates without picks for the project in general, but that's outside of this PRs scope.

Hovering without a pick in a normal, single, readout:
image

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -691,50 +631,63 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
...
-            this.props.lineStyle?.color,
-            (info as WellsPickInfo).featureType
-        );
-        if (!md_property) {
+        let md_property: PropertyDataType | null = null;
+        let tvd_property: PropertyDataType | null = null;
+        let log_property: PropertyDataType | null = null;
+
+        // ! This needs to be updated if we ever change the sub-layer id!
+        if (info.sourceLayer?.id === `${this.props.id}-log_curve`) {
```
hkfb, line 647: Could we use the class name instead?

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
Anders2303: What do you mean? The class name of the curve layer, Like this?
`info.sourceLayer instaceof PathLayer`

hkfb: I had something like that in mind, but I realize it could collide with other `PathLayer`s used as sublayers. It could work if we created a new `LogCurveLayer` based on `PathLayer`.

A second option to prevent future regression could be to extract the "log_curve" text used in both the template and as the layer id to a constant.

Anders2303: I'll go with the constant for now, I think that's the cleanest

hkfb (approved):
Thanks for the refactoring!

typescript/packages/subsurface-viewer/src/hooks/useMultiViewPicking/MultiViewPickingInfoAssembler.ts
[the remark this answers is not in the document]
w1nklr: OK, forget about my comment, it seems to be wrong.

Current implementation already displays the position even if not picking an object.

BTW, this file is NOT used by the Map, it is only provided as a hook and only used in a MultiViewExample.
So it seems that the image that you show is not going through that code... (and it seems that the bundler does not even provide that file in viewers).

So in the end, viewers is not impacted by these changes.
I'll just resolve this...

typescript/packages/subsurface-viewer/src/hooks/useMultiViewPicking/MultiViewPickingInfoAssembler.ts
```diff
@@ -114,14 +119,38 @@ export class MultiViewPickingInfoAssembler {
...
+            // ! For some reason, multi-pick pads the array up to pick-depth length, repeating the last element
+            const multPickResult = deck.pickMultipleObjects({
+                depth: this._options.pickDepth,
+                unproject3D: true,
+                x,
+                y,
+            });
+
+            // Ensure the top-most element is processed first by sorting on z-coordinate
+            return _.sortBy(multPickResult, (pick) => pick.coordinate?.[2]);
```
w1nklr, line 136: You have to sort relative to to the z-buffer distance, not plain z.

This can be approximated by square distance from camera position to coordinates.

BTW, let me please resolve my comments...

w1nklr (changes requested):

typescript/packages/subsurface-viewer/src/hooks/useMultiViewPicking/MultiViewPickingInfoAssembler.ts
Anders2303: Fixed, went with `mathjs.distance()` for the computation

w1nklr (approved):

Issue #2496, opened 2025-03-31
[NGRM] Line layer picking is inconsistent with Map/Grid3D layer


w1nklr:
Cursor readout coordinates are wrong when the cursor is over a line.

How to reproduce:
- Open a story like http://localhost:6007/?path=/story/subsurfaceviewer-grid3d-layer--simgrid or http://localhost:6007/?path=/story/subsurfaceviewer-map-layer--small-map
- zoom in to separate clearly lines from cells
- move the mouse over a line a see the readout

For example:

![Image](https://github.com/user-attachments/assets/6373db2e-b6df-415a-a65e-06bf46b80be0)

or

![Image](https://github.com/user-attachments/assets/8ff89b4b-1316-459d-aa1b-13e54082c991)

hkfb:
@nilscb I also observe that if I zoom out so far that the grid is covered with only lines, and then double click the grid to set the rotation point, then zooming doesn't work properly any longer. Could it be related?

PR #2501, opened 2025-04-02
fix: [NGRM] Line layer picking is inconsistent with Map/Grid3D

nilscb:


hkfb (approved):

hkfb:
Does it resolve the issue mentioned here: https://github.com/equinor/webviz-subsurface-components/issues/2496#issuecomment-2772343341

Issue #2510, opened 2025-04-11
[NGRM] When zoomed in text labels on 3D axes shrinks.


nilscb:
See this discussion: https://github.com/visgl/deck.gl/issues/7992

PR #2511, opened 2025-04-11
fix: [NGRM] When zoomed in text labels on 3D axes shrinks. #2510

nilscb:
See this discussion: https://github.com/visgl/deck.gl/issues/7992


typescript/packages/subsurface-viewer/src/extensions/fixed-size-extension.ts
```diff
@@ -0,0 +1,13 @@
+import { LayerExtension } from "@deck.gl/core";
+
+// Note: See this discussion for more details: https://github.com/visgl/deck.gl/issues/7992
+
+export class FixedSizeExtension extends LayerExtension {
+    getShaders() {
+        return {
+            inject: {
+                "vs:DECKGL_FILTER_SIZE": "size *= 0.7 * gl_Position.w;",
```
hkfb, line 9: What is the 0.7 factor?

Is it possible to get the `minSize * w` and `maxSize * w` and clamp the `size` instead? This way we can still optionally keep perspective sizes but prevent growing or shrinking too much.

nilscb: This is directly taken from the deck.gl discussion mentioned so I dont know the origin of this factor. I think we should use as is until they come up with a more permanent solution.

hkfb: > Is it possible to get the minSize * w and maxSize * w and clamp the size instead? This way we can still optionally keep perspective sizes but prevent growing or shrinking too much.

I think you can disregard this suggestion as we don't know the min and max sizes in the DECKGL_FILTER_SIZE macro.

Still I'm not sure about the 0.7 factor. It looks like all labels are scaled down unnecessarily.

typescript/packages/subsurface-viewer/src/extensions/fixed-size-extension.ts
```diff
@@ -0,0 +1,13 @@
+import { LayerExtension } from "@deck.gl/core";
+
+// Note: See this discussion for more details: https://github.com/visgl/deck.gl/issues/7992
+
+export class FixedSizeExtension extends LayerExtension {
```
w1nklr, line 5: Add some doc

typescript/packages/subsurface-viewer/src/storybook/examples/__image_snapshots__/subsurfaceviewer-examples--user-defined-layer-2.png
w1nklr, line 1: Wrong snapshot

typescript/packages/subsurface-viewer/src/extensions/fixed-size-extension.ts
nilscb: I changed the 0.7 magic factor to 1 and it seems better.

w1nklr: I would say we should try to be consistent. For example, the displayed labels should be "of same size" than well names with the size value.

w1nklr (changes requested):
See comments

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -398,6 +399,7 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
             background:
                 this.props.wellLabel?.background ||
                 this.props.hideOverlappingWellNames,
+            extensions: [new FixedSizeExtension()],
```
hkfb, line 402: Suggest moving this to the default props of `WellLabelLayer`

hkfb (changes requested):

typescript/packages/subsurface-viewer/src/storybook/examples/__image_snapshots__/subsurfaceviewer-examples--user-defined-layer-2.png
hkfb: Fixed

typescript/packages/subsurface-viewer/src/extensions/fixed-size-extension.ts
hkfb: Fixed

w1nklr (approved):

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
hkfb: I think it's in the right place, because we only want to modify the `TextLayer`

hkfb (approved):

PR #2535, opened 2025-05-05
fix: Component for displaying performance metrics

nilscb:
Added a component to show metrics from deck.gl (like frames pr sec etc)
Demonstrated in story Map/BigMap3D

### DIFF
--- a/typescript/packages/subsurface-viewer/src/components/Map.tsx
+++ b/typescript/packages/subsurface-viewer/src/components/Map.tsx
@@ -69,7 +69,11 @@ import StatusIndicator from "./StatusIndicator";
 import type { Unit } from "convert-units";
 import IntersectionView from "../views/intersectionView";
 
-import type { LightsType, TLayerDefinition } from "../SubsurfaceViewer";
+import type {
+    LightsType,
+    TLayerDefinition,
+    DeckMetrics,
+} from "../SubsurfaceViewer";
 import { getZoom, useLateralZoom } from "../utils/camera";
 import { useScaleFactor, useShiftHeld } from "../utils/event";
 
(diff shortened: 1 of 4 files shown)


w1nklr (changes requested):
I (with my current react knowledge) am wondering about this usage of ref/imperative handle.
I don't know all the benefits/drawbacks, but I feel it's a bit overkill for the need.

Curious, I checked and currently found only 1 (resp. 2) usage of forwardRef (resp. useImperativeHandle) in Deck.gl repo, 0 (resp. 1) in webviz and about 15 (I did not count duplicated code) (resp. 0)  in the whole NGR code.

nilscb:
The reason for using forward ref etc is that I want only the metrics component to redraw when a new metrics update occurs. Alternatively it could be a state in the story component and the metrics info could be sent to the metrics component trough a property. However that would cause also the deck component to redraw which is undesirable. But this rest on my understanding that if a react components state is changed then all its children will also be redrawn, I could be wrong here (Håvard disagrees I think) I don't have any strong feelings about it but redrawing deck may alter its statistics which is the very thing we want to measure.

w1nklr (approved):
Let's discuss the forwardref/imperative handle with @hkfb .

Testing if other approach forces a redraw will be easy: this would cause endless rendering, wouldn't it ?

w1nklr:
Here a snapshot:
image

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -399,6 +403,11 @@ export interface MapProps {
      * The reference to the deck.gl instance.
      */
     deckGlRef?: React.ForwardedRef<DeckGLRef>;
+
+    /**
+     * Callback called from deck.gl whith metrics data.
+     */
+    onMetrics?: ((m: DeckMetrics) => void) | null;
```
hkfb, line 410: Suggest using something like `DeckGL["_onMetrics"]`, so that we don't need to duplicate the type

hkfb (changes requested):

PR #2543, opened 2025-05-14
chore: bump node to 20.18.3

w1nklr:
Bump node from 18.19.0 to 20.18.3


hkfb:
Is "fix" the right prefix here? The change doesn't affect production right?

w1nklr:
> Is "fix" the right prefix here? The change doesn't affect production right?

Correct, that should not impact the produced package.
Personally, I nevertheless prefer changing the patch version (it's no so uncommon to have side-effects for things that are considered as safe).

But I can change it to chore, if you prefer.

hkfb:
> But I can change it to chore, if you prefer.

Please do 👍

hkfb (approved):

PR #2550, opened 2025-05-19
Revert node update

w1nklr:
Revert #2543 

Starting with node 20.0.0, npm does not report npm script error status in workspaces :/


w1nklr:
The issue was that @theunderscorer/nx-semantic-release did pull in a dependency on npm 8.19.4

>$ npm ls npm
@webviz/subsurface-components@ C:\dev\AzDevOps\github\equinor\webviz-subsurface-components\typescript
└─┬ @theunderscorer/nx-semantic-release@2.12.0
  └─┬ @semantic-release/npm@9.0.2
    └── npm@8.19.4

It did take precedence on the global npm version handled by volta, leading to the bug

PR #2559, opened 2025-05-21
chore: tsconfig not excluding all the tests.

w1nklr:
Update tsconfig to exclude *.test.ts from build and package


typescript/packages/subsurface-viewer/package.json
hkfb, line 1: https://github.com/equinor/webviz-subsurface-components/pull/2557 will change this again, so maybe we can merge it first?

PR #2564, opened 2025-05-23
chore: "Typed Array Input" triangle story crash when editing layer settings

w1nklr:
This is a known issue of storybook. Story controls are handled as Json, with stringified version used in the control panel. When edited, the modified string version is transformed to a Json structure and send as arguments to render the story.

This breaks the non Json compatible data, like arrays.

The triangle story was not using the workaround used to avoid crashes.


hkfb (approved):
LGTM 👍

Issue #2479, opened 2025-03-19
`SubsurfaceViewer`: colorAbove/colorBelow/colorNaN possibly not working?


vegardkv:
Background for this issue is that the maps visualized in the CO2 leakage part of webviz-subsurface looks incorrect. Most likely coming from replacing `DashSubsurfaceViewer` with `SubsurfaceViewer` (https://github.com/equinor/webviz-subsurface/pull/1312):

Upper image is before, lower is after. The red area on the "outskirts" on the lower should be blue as in the upper:

![Image](https://github.com/user-attachments/assets/2418e906-b613-4ccc-ad14-19da31ea87db)

SubsurfaceViewer accepts a colorScales argument, which is a list of color scales, and where each color scale (according to the documentation) can have optional colorAbove/colorBelow/colorNaN properties. However, these setting does not seem to have any effect.

My suspicion is that this might be due to the `getRgbData` in a different repo, which I think should reference these properties somehow: https://github.com/emerson-eps/color-tables/blob/bb3d1e866c3fa0a7cedc71159169046b9703d6b3/react-app/src/component/Utils/legendCommonFunction.ts#L62

This function is invoked from: https://github.com/equinor/webviz-subsurface-components/blob/ca5e89beb6be2ec83fa2fc09ad6bac91bfa3bac3/typescript/packages/subsurface-viewer/src/layers/colormap/colormapLayer.ts#L23

Even though this is in a different repo, this is just my suspicion based on a brief investigation with limited react/tsx experience.

Apologize in advance for not producing a reproducible issue, but I'll get more info if needed and create one if worst comes to worst.

hkfb:
Would you consider trying the `MapLayer` instead? Any issues you might find with it will likely affect also us, so it's easier to justify prioritizing that.

vegardkv:
Yes, got an hint that this was an alternative. At first glance, this seems to work quite well (https://github.com/AudunSektnanNR/webviz-subsurface-as/pull/70). Will report back if we run into any issues layer, otherwise, I think we will just stick to MapLayer, since I don't think we lose anything by switching.

hkfb:
Please let us know how it works out. If no-one uses the Colormap and Hillshading layers and the MapLayer works sufficiently well as a replacement, then I would like to deprecate those. @HansKallekleiv

vegardkv:
> Please let us know how it works out

Will do!

hkfb:
[...]

PR #2569, opened 2025-05-30
fix: "SubsurfaceViewer: colorAbove/colorBelow/colorNaN possibly not working? #2479"

nilscb:
Added a few more properties to colormap layer. 
See new stories under HillshadingLayer folder.
This will allow contour lines and hillshading directly in colormap layer.

Also added option to set clamp color.  The previous behavior would clamp to colormap min and max value. This will give the option to clamp to a given fixed color. (or transparent)

New properties:

```
    // If true, draw contour lines.  Default false.
    contours: boolean;

    // If true, apply hillshading. Default false.
    hillshading: boolean;

    // Contour reference height. Default 0.
    contourReferencePoint: number;

    // Height between contour lines. Default 50.
    contourInterval: number;

    /**  Clamp colormap to this color at ends.
     * Given as array of three values (r,g,b) e.g: [255, 0, 0]
     * If not set or set to true, it will clamp to color map min and max values.
     * If set to false the clamp color will be completely transparent.
     */
    colorMapClampColor: Color | undefined | boolean;

    // Optional height map. If set hillshading and contourlines will be based on this map.
    heightMapUrl: string;

    // Min and max values of optional height map.
    // Defaults to "valueRange".
    heightValueRange: [number, number];
```


nilscb:
image

PR #2572, opened 2025-06-05
fix: `Grid3DLayer` not including default `shaderModules` and `layer.opacity` not used in fragment shader

rubenthoms:
Using `super.getShaderModules({})` injects the default layer modules, e.g. `layerUniforms` which is required to access `props.opacity` in the fragment shader via `layer.opacity`.

### DIFF
--- a/typescript/packages/subsurface-viewer/src/layers/grid3d/nodeProperty.fs.glsl.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/grid3d/nodeProperty.fs.glsl.ts
@@ -50,7 +50,7 @@ void main(void) {
 
    // Use two sided phong lighting. This has no effect if "material" property is not set.
    vec3 lightColor = lighting_getLightColor(color.rgb, cameraPosition, position_commonspace.xyz, normal);
-   fragColor = vec4(lightColor, 1.0);
+   fragColor = vec4(lightColor, layer.opacity);
    DECKGL_FILTER_COLOR(fragColor, geometry);
 }
 `;
--- a/typescript/packages/subsurface-viewer/src/layers/grid3d/cellProperty.fs.glsl.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/grid3d/cellProperty.fs.glsl.ts
@@ -87,7 +87,7 @@ void main(void) {
 
    // Use two sided phong lighting. This has no effect if "material" property is not set.
    vec3 lightColor = lighting_getLightColor(color.rgb, cameraPosition, position_commonspace.xyz, normal);
-   fragColor = vec4(lightColor, 1.0);
+   fragColor = vec4(lightColor, layer.opacity);
    DECKGL_FILTER_COLOR(fragColor, geometry);
 }
 `;
--- a/typescript/packages/subsurface-viewer/src/layers/grid3d/privateGrid3dLayer.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/grid3d/privateGrid3dLayer.ts
@@ -159,8 +159,20 @@ export default class PrivateLayer extends Layer<PrivateLayerProps> {
         const colormap = this.getColormapTexture(context);
         const mesh_model = new Model(context.device, {
             id: `${this.props.id}-mesh`,
-            vs: geometricShading ? linearVertexShader : flatVertexShader,
-            fs: geometricShading ? linearFragmentShader : flatFragmentShader,
+            ...super.getShaders({
+                vs: geometricShading ? linearVertexShader : flatVertexShader,
+                fs: geometricShading
+                    ? linearFragmentShader
+                    : flatFragmentShader,
+                modules: [
+                    project32,
+                    picking,
+                    lighting,
+                    phongMaterial,
+                    utilities,
+                    gridUniforms,
+                ],
+            }),
             geometry: new Geometry({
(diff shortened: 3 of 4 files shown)


rubenthoms:
> BTW, shouldn't the 2 last lines of `cellProperty.vs.glsl.ts` and `nodeProperty.vs.glsl.ts` be removed ?
> 
> ```
>    vec4 color = vec4(0.0);
>    DECKGL_FILTER_COLOR(color, geometry);
> ```
> 
> This should not be part of a vertex shader, right ?

Good question. This seems to apply a highlighting color to picked entities. In deck.gl layers, this is also called within the vertex shader. So either we are doing it right or they are doing it wrong.

w1nklr:
> > BTW, shouldn't the 2 last lines of `cellProperty.vs.glsl.ts` and `nodeProperty.vs.glsl.ts` be removed ?
> > ```
> >    vec4 color = vec4(0.0);
> >    DECKGL_FILTER_COLOR(color, geometry);
> > ```
> >    
> >   
> > This should not be part of a vertex shader, right ?
> 
> Good question. This seems to apply a highlighting to entities. In deck.gl layers, this is also called within the vertex shader. So either we are doing it right or they are doing it wrong. I suggest we look at this later and not as part of this PR.

I gave a look and understand better.
These DECKGL_FILTER_COLOR, be it in vertex or fragment shader, allow users of the layer to inject custom behavior.
In particular, this is used by deck.gl for picking !
This mean that DECKGL_FILTER_COLOR should be in (nearly) each and every shader file.

I counted 14 shader files without DECKGL_FILTER_COLOR.
We probably don't want to apply that on the 4 axes2D shaders (axis2d is not really meant to display data and should be eventually be implemented not as a layer).
For other shaders (like for the north arrow or piechart ?), it is not sure that we want to pick them.

In any case, all (most) of them should apply the layer.opacity.

Here is the list:
```bash
# should support layer.alpha and picking (thus DECKGL_FILTER_COLOR)
src/layers/axes/box.fs.glsl.ts
src/layers/axes/box.vs.glsl.ts
src/layers/gpglLayers/triangleLine.fs.glsl.ts
src/layers/gpglLayers/triangleLine.vs.glsl.ts
src/layers/grid3d/line.vs.glsl.ts
src/layers/map/line.vs.glsl.ts
src/layers/triangle/line.fs.glsl.ts
src/layers/triangle/line.vs.glsl.ts
# should support layer.alpha
# and probably also not picking (thus no DECKGL_FILTER_COLOR)
src/layers/northarrow/northarrow.fs.glsl.ts
src/layers/northarrow/northarrow.vs.glsl.ts
src/layers/piechart/piechart.fs.glsl.ts
src/layers/piechart/piechart.vs.glsl.ts
# should not support picking (thus no DECKGL_FILTER_COLOR)
# and probably no layer.alpha (should be considered like the distance ruler)
src/layers/axes2d/label.fs.glsl.ts
src/layers/axes2d/label.vs.glsl.ts
src/layers/axes2d/line.fs.glsl.ts
src/layers/axes2d/line.vs.glsl.ts
# utility, should not be changed at all
src/layers/shader_modules/decoder.fs.glsl.ts
```

Do you feel to handle that in this PR, or do you prefer to handle that appart ?

typescript/packages/subsurface-viewer/src/layers/grid3d/nodeProperty.fs.glsl.ts
```diff
@@ -42,15 +43,10 @@ void main(void) {
       fragColor = encodeVertexIndexToRGB(vertexIndex);      
       return;
    }
-       
-   // This may happen due to GPU interpolation precision causing color artifacts.
-   float propertyValue = clamp(property, grid.valueRangeMin, grid.valueRangeMax);
```
w1nklr, line 47: keep clamping in fragment shader. we fought hard to have it working correctly :/ (for record, we never had issues on own machines, but it was hard getting ci/cd machines working, possibly due to precision / software OpenGL).

typescript/packages/subsurface-viewer/src/layers/grid3d/nodeProperty.vs.glsl.ts
```diff
@@ -49,7 +50,11 @@ void main(void) {
 
    DECKGL_FILTER_GL_POSITION(gl_Position, geometry);
 
-   vec4 color = vec4(0.0);
-   DECKGL_FILTER_COLOR(color, geometry);
+   // This may happen due to GPU interpolation precision causing color artifacts.
+   float propertyValue = clamp(property, grid.valueRangeMin, grid.valueRangeMax);
+
+   vColor = getPropertyColor(propertyValue);
```
w1nklr, line 56: Nope.
There is a huge difference between value interpolation and color interpolation.

Say we interpolation between 2 vertices, the value between 0 and 1 with a colormap being black for 0, white for 1 and red for 0.5.

Value interpolation (ie. previous implementation) will interpolate the value between 0 and 1, and the fragment shader will pick the corresponding color => color will start black (for 0), evolving to red (for 0.5) and to white (for 1).

Color interpolation (new impl) will compute the color for 0 which is black and for 1, which is white.
These 2 colors are used for the interpolation: in the end, the color will range from black to white, completely skipping the red.

w1nklr: So please, do the same as for cellProperty version ;)

PR #2587, opened 2025-06-23
fix: make triangle line shader consistent with triangle shader

w1nklr:
Add DECKGL_FILTER_GL_POSITION to triangle line vertex shader for consistency with triangle vertex shader


rubenthoms (approved):
LGTM :+1:

Issue #2631, opened 2025-10-08
fix readout for cross section view


nilscb:
See story Wells Layer/unfolded projection

Chage readout from x,y z to z and distance for unfolded views.

hkfb:
@nilscb is this done? Can we close this issue?

PR #2632, opened 2025-10-08
fix: "fix readout for cross section view #2631"

nilscb:


typescript/packages/subsurface-viewer/src/viewports/sectionViewport.ts
```diff
@@ -0,0 +1,22 @@
+import { OrthographicViewport } from "@deck.gl/core";
+import type { OrthographicViewportOptions } from "@deck.gl/core/dist/viewports/orthographic-viewport";
+
+export type Padding = {
+    left?: number;
+    right?: number;
+    top?: number;
+    bottom?: number;
+};
```
hkfb: Is this needed?

PR #2639, opened 2025-10-14
fix: readout for section view

hkfb:


nilscb (approved):

typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -409,6 +409,25 @@ export interface MapProps {
...
+
+// Helper function to handle deprecated "show3D" property of ViewportType
+function show3D(viewPort: ViewportType): boolean {
+    if (typeof viewPort.show3D === "boolean") {
+        console.warn("show3D is deprecated. Use viewType");
+        return viewPort.show3D;
+    }
+
+    if (viewPort.viewType !== undefined) {
+        return viewPort.viewType === OrbitView;
```
w1nklr: This does return false if viewPort.viewType is undefined, right ?
You can remove test on line 425

YCR

PR #2651, opened 2025-10-24
fix: map moves to origin when double-clicking

### LINKED ISSUE #2642: SubsurfaceViewer / Map - double moves anchor to origin
https://github.com/user-attachments/assets/7425ff25-3aef-426e-8548-83f525465628

Double clicking in the map view no longer repositions to whatever you clicked, instead moving the camera back to the world origin. 

To recreate, go to any of the map storybooks (for example [MapLayer2D](https://equinor.github.io/webviz-subsurface-components/storybook-static/?path=/story/subsurfaceviewer-map-layer--map-layer-2-d), as in the video) and try double-clicking

Anders2303:
Resolves #2642 

My changes done in #2621 accidentally forgot to return the scaled coordinate after picking, which was causing the map to move to origin on double-clicks


Anders2303:
> A simple test could be a new story displaying a map and using the `play` function to implement a double click. There are several example stories using `play` to test story behaviour.

Already tried that; I am unable to make an play function that triggers deck.gl's click event. For instance, I've tried this:
```typescript
play: async (args) => {
        const delay = 500;
        const canvas = args.canvasElement.querySelector("#map-wrapper");

        if (!canvas) {
            return;
        }

        await userEvent.dblClick(canvas, { delay });
}
```

But this does not trigger deck.gl's click event, only a "hover" event. I also tried `userEvent.click`, `userEvent.pointer` and `fireEvent`, and none of those worked either...

Unless any of you can write a test that works, I'd say should leave it without a test for now, as this is an important bug to fix. We can open a dedicated issue for general interaction tests for camera movement (drag, recenter, 3d-rotate, and so on should all be tested)

hkfb:
> > A simple test could be a new story displaying a map and using the `play` function to implement a double click. There are several example stories using `play` to test story behaviour.
> 
> Already tried that; I am unable to make an play function that triggers deck.gl's click event. 

Thanks for giving it a shot!

Issue #2254, opened 2024-09-20
Viewport with empty layer ids list shows all layers


rubenthoms:
I would expect a viewport with an empty layer ids list to not show anything.

w1nklr:
It is for convenience.
If you display a single view (the usual case), you probably don't want to specify all the displayed layers on one side plus all the corresponding layer ids for the viewport...

rubenthoms:
I understand the use case. However, in my opinion, sending in an empty list should result in showing no layers, as specified. Why can't `layerIds` be left `undefined` if one wants all layers to be displayed?

w1nklr:
I don't know. I was not there at the initial design...
I don't really see the use case needed to display nothing. 

But if you have such an use case and if you can provide an easy way to display all the layers (and set that as DEFAULT_VIEWS), I don't object to a change.

BTW, this would be closer to Deck.gl approach: https://github.com/visgl/deck.gl/blob/4dd814805121ab70c16ea3b458593ecc92f92f1a/docs/api-reference/core/layer-manager.md?plain=1#L46

rubenthoms:
We have a component where individual viewports can be added. To each of the latter custom layers can be added later on (in some sort of tree arrangement). When adding a layer to the first viewport, it becomes visible in all (empty) viewports although they should not have any children. We would be glad to have the possibility to temporarily have views without any layers.

hkfb:
Agree, it makes sense to be able to show empty views, especially in the multi-viewport case.

danielsantiago1002:
Hey @hkfb, I've worked on the issue and have a few questions before moving forward.

I found the part of the code that creates the displayed behavior:

```
    const layerFilter = useCallback(
        (args: { layer: Layer; viewport: Viewport }): boolean => {
            // display all the layers if views are not specified correctly
            if (!views?.viewports || !views?.layout) {
                return true;
            }
            const cur_view = views.viewports.find(
                ({ id }) => args.viewport.id && id === args.viewport.id
            );
            if (cur_view?.layerIds && cur_view.layerIds.length > 0) {
                const layer_ids = cur_view.layerIds;
                return layer_ids.some((layer_id) => {
                    const t = layer_id === args.layer.id;
                    return t;
                });
            } else {
                // Changing this to false to avoid displaying all layers in all views by default
                return false;
            }
        },
        [views]
    );
```

However, I found one possible Edge case that might be worth checking out:

[...]

PR #2655, opened 2025-11-11
fix: subsurface viewer, viewport with empty layer ids list should sho…

danielsantiago1002:
…w an empty view

solves: https://github.com/equinor/webviz-subsurface-components/issues/2254

Updated tests, components, and storybook examples to consistently handle cases where viewports have empty or undefined layerIds. Improved layer filtering logic in Map.tsx to only consider layers present in the current views, and adjusted conditional checks to ensure correct behavior when layerIds is undefined or empty.

### DIFF
--- a/typescript/package-lock.json
+++ b/typescript/package-lock.json
@@ -17,7 +17,7 @@
         "@mdx-js/react": "^3.0.1",
         "@mui/material": "^5.15.15",
         "@mui/system": "^5.15.14",
-        "@nx/jest": "^22.0.2",
+        "@nx/jest": "^22.0.3",
         "@storybook/addon-essentials": "^8.6.14",
         "@storybook/addon-interactions": "^8.6.14",
         "@storybook/addon-links": "^8.6.14",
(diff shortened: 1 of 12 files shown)


danielsantiago1002:
@hkfb I was able to run the snapshots, and the two new images were created (for the two new stories).

However, two additional images were modified. The change seems minimal (only a few bits), but it's there, and I thought it was worth mentioning: 

- typescript\packages\subsurface-viewer\src\storybook\examples\__image_snapshots__\subsurfaceviewer-examples--map-material.png
- typescript\packages\subsurface-viewer\src\storybook\layers\__image_snapshots__\subsurfaceviewer-gpgl-layers--gpgl-texture-with-explicit-mesh.png

Additionally, I had some problems running "npm run format." It seems that for some reason, when I run it, it "modifies" all the files (it seems to be related to 'LF' to 'CRLF' (again, Windows). I'm investigating). However, I did submit the changes for the Map component.

Thanks! I'll be waiting if anything else is needed.

hkfb (approved):

Issue #2656, opened 2025-11-12
Improve default camera controls


hkfb:
Suggestions to improve default mouse camera control behaviour in `SubsurfaceViewer`

- [ ] Increase default zoom granularity. Currently using the scrollwheel, each scroll increment results in a large jump in the zoom level. (Note that currently the zoom increment can be reduced by holding the shift key)
- [ ] Smooth zooming. Add easing to animate smooth transitions between zoom levels.
- [ ] Add inertia to pan and rotate. Continue to pan or rotate after releasing the mouse button. Cf. panning behaviour in [Google maps](https://www.google.com/maps).

Preferably these controls should be configurable through the API, but with sensible default values if not configured.

These suggestions are subjective. Feel free to comment.

PR #2657, opened 2025-11-14
feat: Subsurface Viewer, Improve default camera controls

danielsantiago1002:
Introduces the defineController function to centralize controller configuration for views, applying default options and merging custom props. Refactors Map.tsx to use this utility. 

https://github.com/equinor/webviz-subsurface-components/issues/2656


typescript/packages/subsurface-viewer/src/components/Map.tsx
```diff
@@ -1708,16 +1708,11 @@ function newView(
     const [ViewType, Controller] = getViewType(viewport);
     return new ViewType({
         id: viewport.id,
-        controller: {
-            type: Controller,
-            doubleClickZoom: false,
-        },
-
+        controller: defineController(Controller, viewport?.controller),
```
w1nklr: What is the behavior if defineController returns null or false ?

typescript/packages/subsurface-viewer/src/views/viewport.ts
```diff
@@ -54,3 +76,25 @@ export const useVerticalScale = (viewports: ViewportType[] | undefined) => {
...
     }, [viewports]);
 };
+
+export const defineController = (
+    controllerClass: ConstructorOf<Controller<any>>,
+    controllerProps?: ControllerProps
+) => {
+    const controllerPropsIsNullOrFalse =
+        controllerProps === null || controllerProps === false;
+    if (controllerPropsIsNullOrFalse) {
```
w1nklr: Replace controllerPropsIsNullOrFalse by controllerProps === null || controllerProps === false

typescript/packages/subsurface-viewer/src/views/viewport.ts
```diff
@@ -54,3 +76,25 @@ export const useVerticalScale = (viewports: ViewportType[] | undefined) => {
...
+        return controllerProps;
```
w1nklr: This returns either null or false, right ?

typescript/packages/subsurface-viewer/src/views/viewport.ts
```diff
@@ -1,5 +1,22 @@
 import React from "react";
 import type { ViewTypeType } from "../components/Map";
+import { Controller, View } from "@deck.gl/core";
+import { ConstructorOf } from "@deck.gl/core/dist/types/types";
+import { ControllerOptions } from "@deck.gl/core/dist/controllers/controller";
+
+type ControllerProps = Exclude<
```
w1nklr: What is the difference with ControllerOptions ?

typescript/packages/subsurface-viewer/src/components/Map.tsx
danielsantiago1002: If that happens, the view shouldn't be interactive. This is a use case defined by Deck GL (I'm not sure if it's relevant to consider it).

reference: https://deck.gl/docs/api-reference/core/view#controller

typescript/packages/subsurface-viewer/src/views/viewport.ts
danielsantiago1002: With this typing, we're including false or null (for the non-interactive view) and excluding the user's ability to define the class (because we're doing it ourselves). This is a draft; I'm open to suggestions for a better approach.

https://github.com/equinor/webviz-subsurface-components/issues/2656#issuecomment-3533585094

typescript/packages/subsurface-viewer/src/views/viewport.ts
```diff
@@ -1,5 +1,22 @@
...
+    View["props"]["controller"],
+    ConstructorOf<Controller<any>>
+>;
+
+export const DEFAULT_CONTROLLER_OPTIONS = {
```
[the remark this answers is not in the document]
danielsantiago1002: done, Thanks for the recommendation.

hkfb:
> @hkfb I ran additional tests and searched for similar issues. Found this issue, which may have some similarities: [visgl/deck.gl#7158](https://github.com/visgl/deck.gl/issues/7158).
> 
> It seems the bug might be due to a conflict between the React viewState implementation and DeckGL's smoothing option. I did the tests using the DeckGl component as an uncontrolled component and the "smooth" option seems to work fine.
> 

Yes, the issue might be caused by the manipulation of the view state in Map.tsx

> 
> I'll keep an eye on it in case there are more changes. Should we create a test for these changes, or will they be fine as they are?

It would be useful to have a story to demonstrate the behaviour of overridden controller settings.

Eg. a simple story showing a map, while using the [`play` function](https://storybook.js.org/docs/writing-stories/play-function) to demonstrate the effect of zooming (see eg. the "ScaleFactorHook" story in CameraControlExamples.stories.tsx).

Check also [args](https://storybook.js.org/docs/writing-stories/args) and  [ArgTypes](https://storybook.js.org/docs/api/arg-types) for exposing the controller settings in the story controls (also used in the ScaleFactorHook story).

w1nklr (approved):

typescript/packages/subsurface-viewer/src/views/viewport.ts
w1nklr: boolean is good for me 👍

hkfb:
It passed finally. I think we can merge it. If it remains too flaky we can adjust or disable it post-hoc. Thanks @danielsantiago1002

hkfb:
Created a follow-up issue for the smooth zooming issue: https://github.com/equinor/webviz-subsurface-components/issues/2665

Issue #2661, opened 2025-11-18
Change WellsLayer data-model to use WellLogJson instead


Anders2303:
As discussed; we want to change the wells layer to use WellLogJson instead og GeoJson, so the data types are more unified across components. I'll start working on this; we can use this issue for further discussion if needed

w1nklr:
What does `getTrajectoryCurveIndices: CurveIndicesAccessor` return ? 

How is the returned value decided ?
I guess `{x: number, y: number, z: number}` (BTW, missing closing bracket) is the XYZ coordinates.
But are the `[number, number, number]` really indices ?

PR #2674, opened 2025-11-28
fix: Subsurface viewer - wells layer separate log layers

Anders2303:
In preparation for the data model change for the wells-layer (#2661 ), I did some preemptive refactoring and cleanup in the wells-layer to make it easier to work with. Most notably, I separated the logic for visualizing Well-log curves along the track as a dedicated sub-layer. This also fixes assorted bugs I discovered with the log layer as I was working on it.

## Changes
* Halves the size of the `wellsLayer` file; many of the local utilities have been moved to the layer's utility files
* Added the LogCurveLayer class; this is now used as a sub-layer to handle log curve visualizations
  * I think the wellslayer's properties that related to the log layer are terribly unintuitive, so the log layer has different property names that I think are clearer. We might want to consider using these in the wells layer as-well, and deprecating the old properties....
  * **NOTE:** I also moved the wells-layer's "selection" layer into the new LogCurveLayer, as I understood this to be intended as a way of highlighting log tracks specifically; this is technically a regression, as the new selection layer is not able to highlight outside of it's related logs MD-domain
* Fixed the selection layer not being able to follow log curve width if logRadius was set to 0 
* Fixes layer auto-highlights not always picking paired logs/trajectories
* Made the well-log layer better align with the trajectory path. In short terms, there are now interpolated points that match the trajectories points, so the log curve has the same curvature. It looks nice in 3D now too!
* Fixed the MapAndWellLogViewer storybook example, clicking wells, and domain selection now works

### DIFF
--- a/typescript/packages/subsurface-viewer/src/layers/wells/utils/features.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/wells/utils/features.ts
@@ -38,7 +38,10 @@ export function getSize(
 
     if (type == LINE) return DEFAULT_LINE_WIDTH + offset;
     else if (type == POINT) return DEFAULT_POINT_SIZE + offset;
-    else throw `Unrecognized feature type: ${type}`;
+    else {
+        console.warn(`Unrecognized feature type: ${type}`);
+        return 0;
+    }
 }
 
 function getPointGeometry(well_object: WellFeature): Point | undefined {
--- a/typescript/packages/subsurface-viewer/src/layers/wells/utils/wells.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/wells/utils/wells.ts
@@ -71,6 +71,6 @@ export function getPositionByMD(
 function getNeighboringMdIndices(mds: number[], md: number): number[] {
     const idx = mds.findIndex((x) => x >= md);
 
-    if (idx === -1) throw Error(`md ${md} is outside of md range`);
+    if (idx === -1) return [];
     return idx === 0 ? [idx, idx + 1] : [idx - 1, idx];
 }


typescript/packages/subsurface-viewer/src/utils/arrays.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Scales all elements in a numeric array by a given factor
+ * @param array The array to scale
+ * @param factor The scaling factor
+ * @returns A new array with all elements scaled
+ */
+export function scaleArray<T extends number[]>(array: T, factor: number): T {
```
w1nklr: Please add unit tests.

typescript/packages/subsurface-viewer/src/storybook/layers/__image_snapshots__/subsurfaceviewer-wells-layer--callback-styled-wells.png
w1nklr: Are the style changes expected ?

typescript/packages/subsurface-viewer/src/layers/wells/wellsLayer.ts
```diff
@@ -308,10 +289,15 @@ export default class WellsLayer extends CompositeLayer<WellsLayerProps> {
...
+
+        if (!well || !_selection) {
             this.setState({
-                well: well,
-                selection: _selection,
+                logDomainSelection: undefined,
+            });
+        } else {
+            this.setState({
+                logDomainSelection: { well: well, selection: _selection },
```
w1nklr: { well, selection }

typescript/packages/well-log-viewer/src/Storybook/examples/MapAndWellLogViewer.tsx
```diff
@@ -256,12 +257,17 @@ export class MapAndWellLogViewer extends React.Component<
     }
 
     onMapMouseEvent(event: MapMouseEvent): void {
-        console.warn("! onMapMouseEvent event is not working...");
+        if (event.type !== "click") return;
```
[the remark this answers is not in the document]
Anders2303: Oh yeah, it was. I must've miss-read the logic on my first go, will fix

typescript/packages/subsurface-viewer/src/layers/wells/utils/features.ts
```diff
@@ -0,0 +1,42 @@
...
+    offset = 0
+): Accessor<Feature, number> {
+    if (typeof accessor == "function") {
+        return (object, objectInfo) => {
+            const width = getFromAccessor(accessor, object, objectInfo);
+            return (width as number) + offset;
+        };
+    }
+
+    if ((accessor as number) == 0) return 0;
```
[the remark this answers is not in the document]
Anders2303: This function was just extracted from the [wellsLayer file](https://github.com/equinor/webviz-subsurface-components/pull/2674/changes/BASE..c491dd3e813e5f1a88db7ee835efc9f707dc2b22#diff-bc033809fe18e20467b8336176685b60274895af320bb9aac678a05656f17d16L200-L219), so it's been left as is.
Unsure why the offset is ignored, but I think the idea is that a size of 0 is probably intended to be hidden, and should remain hidden even with an offset.

So I'll keep that as is, but I'll add better type checks

PR #2689, opened 2026-01-19
fix: bump the deckgl group to 9.2.6

w1nklr:


typescript/package.json
```diff
@@ -135,5 +135,15 @@
...
+  ],
+  "dependencies": {
+    "@deck.gl/aggregation-layers": "^9.2.5",
+    "@deck.gl/core": "^9.2.5",
+    "@deck.gl/extensions": "^9.2.5",
+    "@deck.gl/geo-layers": "^9.2.5",
+    "@deck.gl/json": "^9.2.5",
+    "@deck.gl/layers": "^9.2.5",
+    "@deck.gl/mesh-layers": "^9.2.5",
+    "@deck.gl/react": "^9.2.5"
```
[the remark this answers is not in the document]
w1nklr: My fault, sorry.

Reverting to draft...

typescript/packages/subsurface-viewer/package.json
```diff
@@ -32,15 +32,15 @@
...
-    "@deck.gl/react": "~9.1.15",
+    "@deck.gl-community/editable-layers": "^9.1.1",
+    "@deck.gl/aggregation-layers": "^9.2.6",
+    "@deck.gl/core": "^9.2.6",
+    "@deck.gl/extensions": "^9.2.6",
+    "@deck.gl/geo-layers": "^9.2.6",
+    "@deck.gl/json": "^9.2.6",
+    "@deck.gl/layers": "^9.2.6",
+    "@deck.gl/mesh-layers": "^9.2.6",
+    "@deck.gl/react": "^9.2.6",
```
hkfb: Is the change from the `~` to `^` prefix intentional? I suspect staying with the `~` prefix is better, knowing that minor deck.gl versions sometimes have breaking changes.

hkfb (changes requested):

typescript/packages/subsurface-viewer/package.json
w1nklr: Initially the move to ~ was to stick to line 9.1, because of deck.gl's wrong handling of their own deps with respect of their breaking minor versions. 
Now that they handle their deps correctly, it does not really matter what we use: the deck.gl/luma.gl libs that will be installed in our apps will be consistent.

Now, I can move back to ~.

Notes:
- that I did not see this change ;)
- it will have no real impact, as deck.gl does now ensure all libs are on the same ~line
- dependabot will anyhow try to bump minor/major versions...

Issue #2659, opened 2025-11-17
Subsurface Viewer: WellLabelLayer has significant slowdown on larger data sets


Anders2303:
The `WellLabelLayer` causes notable frame drops during panning/zooming when the loaded set has a lot of trajectories (In this case, it's with Troll's set of 900+ trajectories). I havent had the chance to dig into exactly what is causing it, but I'm assuming this is related to some of the position or merge logic running on each viewport movement.

However, with my current settings, the labels are static and alway at the end, so no computations should be necessary...

The settings are:
```ts
new WellLabelLayer({
  mergeLabels: false,
  autoPosition: false,
  getPositionAlongPath: 1,
  getAlignmentBaseline: "top",
  getTextAnchor: "end",
  orientation: LabelOrientation.TANGENT,
  // ... other props that don't affect positioning
}),
```

PR #2695, opened 2026-01-22
fix: Revert fix

nilscb:
This is a revert of https://github.com/equinor/webviz-subsurface-components/issues/2659
This may be picked up later but for now I can not see any good solution.


hkfb (approved):
Remember to re-open https://github.com/equinor/webviz-subsurface-components/issues/2659 once this fix is merged.

PR #2752, opened 2026-04-22
feat: add support of an array of domains and selections for the Well Log Viewer

w1nklr:
Accept an array of domains and selections, one for each log view.

Enhance the API to accept an array of domains or of selections to apply individually on each log view
(the current API accepts only one single domain or selection, which are applied to all the log views).

Add a SyncLogViewer story demonstrating it.
Update definition of domain and selection for the SyncLogViewer.
Export some helper functions (isEqDomains, isEqualArrays, isEqualRanges)

### DIFF
--- a/typescript/packages/well-log-viewer/src/SyncLogViewer.stories.tsx
+++ b/typescript/packages/well-log-viewer/src/SyncLogViewer.stories.tsx
@@ -366,6 +366,28 @@ export const DiscreteLogs: StoryObj<typeof TemplateWithSelection> = {
     render: (args) => <TemplateWithSelection {...args} />,
 };
 
+export const DiscreteLogsWithIndividualDomains: StoryObj<
+    typeof TemplateWithSelection
+> = {
+    args: {
+        ...facies3Wells,
+        domain: [
+            [3000, 5500],
+            [2500, 4000],
+            [1000, 4000],
+        ],
+        selection: [
+            [3500, 3600],
+            [2700, 2900],
+            [3300, 3400],
+        ],
+        syncContentDomain: false,
+        syncContentSelection: false,
+        syncTrackPos: false,
+    },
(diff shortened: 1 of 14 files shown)


typescript/packages/well-log-viewer/src/SyncLogViewer.stories.tsx
w1nklr, line 74: This disables all the snapshot tests :/

hkfb: We could try enabling. Could be more stable now. YCR

typescript/packages/well-log-viewer/src/SyncLogViewer.tsx
```diff
@@ -139,14 +148,18 @@ export interface SyncLogViewerProps {
...
 
     /**
-     * Initial selected range
+     * Initial selected range. A single selection applies to all the tracks,
+     * an array of selections applies to the tracks in corresponding views.
      */
-    selection?: [number | undefined, number | undefined];
+    selection?:
+        | [number | undefined, number | undefined]
+        | [number | undefined, number | undefined][];
```
hkfb: Suggestion: extract `[number, number]` and `[number | undefined, number | undefined]` to eg. `NumericRange` and `OpenNumericRange` respectively.

hkfb (approved):

typescript/packages/well-log-viewer/src/SyncLogViewer.tsx
w1nklr: Done as Range and OpenRange.

typescript/packages/well-log-viewer/src/SyncLogViewer.stories.tsx
w1nklr: I prefer doing it in a followup.

PR #2757, opened 2026-04-27
feat: Option to give separate color for each line in PolylinesLayer.

nilscb:
Ammended the "color" property of this layer so that it may be given as an array of colors. One for each polyline.


typescript/packages/subsurface-viewer/src/storybook/layers/PolylinesLayer.stories.tsx
hkfb, line 1: Suggest showing examples of respective single and multi color use cases

nilscb: I changed the small layer example to be multi color but that i perhaps not sufficient

w1nklr (changes requested):
I think it would make more sense to support a list of polylines, each with its own props, and picking should differentiate each polyline.

The use case I see, is to lower the number of deck.gl layers: picking really works with less that 256 layers (if I remember correctly), which is ridiculously small.

nilscb:
Putting this on hold for now. We need to agree if it is needed or not.

PR #2766, opened 2026-05-14
feat: polyline group layer

### LINKED ISSUE #2765: Surface (horizon, fault) sections

hkfb:
## Summary

Adds `PolylineGroupLayer`, a new deck.gl `CompositeLayer` for rendering collections of polylines organised into named groups, with flexible styling, efficient binary data ingestion, discontinuous geometry, dash patterns, GPU-side visibility filtering, and cross-section projection.

### Grouped polylines with a styling cascade

Polylines are organised under `PolylineGroup` objects. Color, width, and dash pattern are resolved per-path through a five-level cascade:

1. polyline-level accessor (`getPolylineColor` / `getPolylineWidth` / `getPolylineDashArray`)
2. `Polyline` field (`color` / `width` / `dashArray`)
3. group-level accessor (`getGroupColor` / `getGroupWidth` / `getGroupDashArray`)
4. `PolylineGroup` field (`color` / `width` / `dashArray`)
5. layer default (`defaultGroupColor` / `defaultGroupWidth` / `defaultGroupDashArray`)

### Discontinuous polylines

A `Polyline.path` may be a nested `PolylineGroup` instead of a flat `Position[]`. This lets one logical polyline consist of disjoint segments — the canonical use-case is a seismic horizon cut by faults. All segments share the root `Polyline.id`, so they are picked, shown, and hidden as a single unit.

### Binary format for large datasets

`PolylineGroup.polylines` accepts a `BinaryPolylines` object (`Float32Array` positions + `Uint32Array` start indices) in addition to `Polyline[]`. This avoids per-object JavaScript overhead for large datasets loaded from binary sources. Group-level styling applies; per-polyline overrides are not supported in this path.

### Dash style

Dash patterns (`[dashLength, gapLength]`) can be set at the group or polyline level via accessors or data fields. When any dash prop is provided, `PathStyleExtension` is activated automatically. A `highPrecisionDash` flag is available to avoid dash-length variation at segment joins.

### GPU-side visibility filtering

`hiddenGroups` and `hiddenPolylines` are `Set` props processed entirely on the GPU via `DataFilterExtension`. Toggling visibility never rebuilds the flattened vertex buffer.

### Cross-section projection

When `sectionPath` (a fence line as 2-D world-space XY coordinates) is supplied, path coordinates are interpreted as `[abscissa, depth]`. The layer automatically renders two sub-layers: one in abscissa/depth space routed to `SectionViewport`s, and one with abscissa values projected back onto the fence for 3-D viewports. The projection index is built once and cached until `sectionPath` changes.

### Rich pick info

`getPickingInfo` exposes the originating `PolylineGroup` and root `Polyline` on the pick result, together with group name, polyline id, and depth-corrected Z value as display properties.

## Storybook

Stories covering all of the above are added under `PolylineGroupLayer.stories.tsx`:

- **Basic** — single group, `Polyline[]` format
- **GroupLevelStyling** — multiple groups with per-group color overrides
- **BinaryPolylines** — typed-array input path
- **DiscontinuousPolylines** — nested `PolylineGroup` path (fault-cut horizons)
- **DashStyle** — `dashArray` at group and polyline level
- **HiddenGroups / HiddenPolylines** — GPU-side visibility toggle
- **SectionViewRendering** — `sectionPath` with side-by-side 3-D and section viewports
- **PickablePolylines** — `pickingRadius`, `onHover`, and an automated play function that simulates a pick


typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,848 @@
...
+            !isEqual(props.data, oldProps.data) ||
+            props.getGroupColor !== oldProps.getGroupColor ||
+            props.getGroupWidth !== oldProps.getGroupWidth ||
+            props.getPolylineColor !== oldProps.getPolylineColor ||
+            props.getPolylineWidth !== oldProps.getPolylineWidth ||
+            props.getGroupPolylines !== oldProps.getGroupPolylines ||
+            props.getPolylinePath !== oldProps.getPolylinePath ||
+            props.defaultGroupColor !== oldProps.defaultGroupColor ||
+            props.defaultGroupWidth !== oldProps.defaultGroupWidth ||
+            props.ZIncreasingDownwards !== oldProps.ZIncreasingDownwards;
```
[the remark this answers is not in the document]
rubenthoms: This is an issue I also wanted to point out. Consider adding
```typescript
props.getGroupDashArray !== oldProps.getGroupDashArray ||
props.getPolylineDashArray !== oldProps.getPolylineDashArray ||
props.defaultGroupDashArray !== oldProps.defaultGroupDashArray ||
```

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,848 @@
...
+            defaultGroupDashArray,
+            getGroupDashArray,
+            getPolylineDashArray,
+            highPrecisionDash,
+        } = this.props;
+
+        const hasDash =
+            defaultGroupDashArray != null ||
+            getGroupDashArray != null ||
+            getPolylineDashArray != null;
```
rubenthoms: Aren't we neglecting directly set dash properties (no acessors)?

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,848 @@
...
+
+        return [
+            new PathLayer(
+                this.getSubLayerProps({
+                    ...sharedProps,
+                    id: "paths",
+                    getPath: (d: FlatEntry) => {
+                        if (!ZIncreasingDownwards) return d.path;
+                        return d.path.map(
+                            ([x, y, z]) => [x, y, -z] as Position
```
rubenthoms: In theory, destructuring `Position` could give `z === undefined`, and, hence, `z = NaN`. Are we sure we never get 2D points here? Consider adding a short check: `return d.path.map(([x, y, z]) => [x, y, -(z ?? 0)] as Position);`

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,848 @@
...
+): FlatEntry[] {
+    const segPolylines = subGroup.polylines;
+    if (isBinaryPolylines(segPolylines)) {
+        console.warn(
+            "PolylineGroupLayer: BinaryPolylines inside a sub-group path is not supported. Skipping."
+        );
+        return [];
+    }
+    return segPolylines.map((segment) => ({
+        path: getPath(segment, group),
```
rubenthoms: In line 45/46, it states that nested polyline groups inside segments are not supported. However, in theory, a user could pass nested polyline groups as there is no guard against that. Consider adding a simple check here, something like:
```typescript
return segPolylines.flatMap((segment) => {
    if (!Array.isArray(segment.path)) {
        console.warn(
            "PolylineGroupLayer: nested PolylineGroup inside a sub-group segment is not supported. Skipping segment."
        );
        return [];
    }
    return [{
        path: getPath(segment, group),
        ...
    }];
});
```

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,848 @@
...
+        props.getGroupPolylines ??
+        ((g: PolylineGroup): Polyline[] | BinaryPolylines => g.polylines);
+    const getPath =
+        props.getPolylinePath ?? ((p: Polyline) => p.path as Position[]);
+    const result: FlatEntry[] = [];
+
+    for (const group of data) {
+        const polylines = getPolylines(group);
+
+        if (isBinaryPolylines(polylines)) {
```
rubenthoms, line 671: Isn't this contradicting your original intention of improving performance with large data sets as binary data? A better approach would be to preserve the binary data and pass it directly to the `PathLayers` instead of reallocating it here.

deck.gl's `PathLayer` accepts binary data natively via `data.startIndices` and `data.attributes.getPath: { value: Float32Array, size: 3 }`. Binary groups should be routed to a separate `PathLayer` using that format, bypassing `FlatEntry` entirely, so the typed arrays are transferred to the GPU without any JS allocation. As-is, `BinaryPolylines` is a data-model convenience type only - not a performance optimization - which is misleading given its documented intent.

rubenthoms (changes requested):
All in all, great job! :+1: I think this layer will come in very handy. A couple of minor issues and improvements to consider. One bigger issue regarding your intention with binary data.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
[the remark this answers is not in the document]
hkfb: Fixed

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,848 @@
...
+                        ? startIndices[i + 1]
+                        : positions.length / 3;
+                const path: Position[] = [];
+                for (let v = start; v < end; v++) {
+                    path.push([
+                        positions[v * 3],
+                        positions[v * 3 + 1],
+                        positions[v * 3 + 2],
+                    ]);
+                }
```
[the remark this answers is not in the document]
hkfb: Fixed

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
hkfb: Fixed. The check was not needed at all.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
hkfb: Fixed.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
hkfb: Fixed

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
hkfb: Good catch! Fixed.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1136 @@
...
+     * Extract the path positions from a polyline object. Defaults to `p.path as Position[]`.
+     * Only called on leaf polylines whose `path` is `Position[]` (i.e. not on container
+     * polylines whose `path` is a `PolylineGroup`).
+     */
+    getPolylinePath?: (polyline: Polyline, group: PolylineGroup) => Position[];
+
+    // -- Fallback defaults ---------------------------------------------------
+
+    /** Fallback color used when no group or polyline color is resolved. Default: `[0, 128, 255, 255]`. */
+    defaultGroupColor?: Color;
```
rubenthoms: You could consider combining the three default values into a single `defaultGroupStyle` to be more symmetric/coherent. This might also prove useful regarding potential future changes to the styles object.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1136 @@
...
+            depthTest,
+            ZIncreasingDownwards,
+            hiddenGroups,
+            hiddenPolylines,
+            highPrecisionDash,
+        } = this.props;
+
+        // Shared sub-layer props for the non-binary `paths` sub-layer (and the
+        // section-mode sub-layers).
+        const sharedProps = {
```
rubenthoms: I find this comment a bit misleading as `sharedProps` are also spread on the binary layer (see line 1038).

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1136 @@
...
+            layers.push(
+                new PathLayer(
+                    this.getSubLayerProps({
+                        id: "paths-binary",
+                        ...sharedProps,
+                        extensions: [
+                            new DataFilterExtension({ filterSize: 1 }),
+                        ],
+                        data: binaryLayerData,
+                        _pathType: "open",
```
rubenthoms, line 1077: Since `_pathType` is an experimental property and might undergo changes, we might want to add a comment here to keep track?

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1136 @@
...
+                        updateTriggers: {
+                            getFilterValue: [hiddenGroups],
+                        },
+                        getColor: undefined, // color comes from attribute
+                        getWidth: undefined, // width comes from attribute
```
rubenthoms, line 1079: Consider explicitly setting `getFilterValue` to `undefined` as well instead of relying on the internal precedence of the binary attribute data over this accessor. If you change this, you can also remove the `updateTriggers` statement.

The same applies to `getDashArray` (less relevant due to the missing `PathStyleExtension`) for consistency.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1136 @@
...
+    });
+}
+
+/**
+ * Allocates combined typed-array buffers and fills them from all raw binary
+ * groups. Positions and indices are concatenated; color/width are replicated
+ * per vertex so the GPU can read them without per-vertex JS work.
+ */
+function buildBinaryData(
+    binaryRaw: { group: PolylineGroup; polylines: BinaryPolylines }[],
```
rubenthoms, line 483: I think a short comment could be useful here
``` `polylines` is a sanitized copy of group.polylines with invalid colors/widths buffers stripped out; do not read from group.polylines directly.```

File: typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts [hunk 12]
```diff
@@ -0,0 +1,1136 @@
...
+            // remaining vertices in one call — no bit-shifting required.
+            colors.set(
+                [color[0], color[1], color[2], color[3] ?? 255],
+                vOffset * 4
+            );
+            new Uint32Array(colors.buffer).fill(
+                new Uint32Array(colors.buffer)[vOffset],
+                vOffset + 1,
+                vOffset + srcVerts
+            );
```
rubenthoms, line 555: This is a nice trick with treating one 32-bit entry as four 8-bit values (32/8 = 4) and letting the `TypedArray.fill` function do the heavy lifting. However, consider adding a more detailed comment as this might not be easily comprehensible at first glance.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1136 @@
...
+    const pathGroup: PolylineGroup[] = new Array(totalPaths);
+    const groups: PolylineGroup[] = binaryRaw.map((b) => b.group);
+
+    let vOffset = 0;
+    let pOffset = 0;
+
+    for (let gIdx = 0; gIdx < binaryRaw.length; gIdx++) {
+        const { group, polylines } = binaryRaw[gIdx];
+        const src = polylines.positions;
+        const srcVerts = src.length / 3;
```
rubenthoms: Consider checking that the polylines are well formatted, i.e. that `src.length % 3 === 0` (if not - console warning or similar).

rubenthoms (changes requested):
Great improvements :+1: A couple comments regarding readability and minor improvements.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1153 @@
...
+    group: PolylineGroup,
+    props: PolylineGroupLayerProps
+): Color {
+    const fromPolylineAccessor = props.getPolylineColor?.(polyline, group);
+    if (fromPolylineAccessor != null) return fromPolylineAccessor;
+    if (polyline.color != null) return polyline.color;
+    const fromGroupAccessor = props.getGroupColor?.(group);
+    if (fromGroupAccessor != null) return fromGroupAccessor;
+    if (group.color != null) return group.color;
+    return props.defaultGroupStyle?.color ?? [0, 128, 255, 255];
```
rubenthoms: For easier maintainability/accessibility we could consider moving all constants into a separate const object at the top of the file.

```typescript
const DEFAULT_GROUP_STYLE: PolylineStyle = { color..., width..., dashArray... }
```

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1153 @@
...
+                    `length (${src.length}) is not a multiple of 3. Trailing ` +
+                    `incomplete vertex will be ignored.`
+            );
+        }
+        const srcVerts = Math.floor(src.length / 3);
+
+        // Copy positions (optionally flipping Z)
+        const base = vOffset * 3;
+        if (ZIncreasingDownwards) {
+            for (let i = 0; i < src.length; i += 3) {
```
rubenthoms: Considering the guard you introduced on line 500, you should use `srcVerts * 3` instead of `src.length` as the bound here.

typescript/packages/subsurface-viewer/src/layers/polyline_group/polylineGroupLayer.ts
```diff
@@ -0,0 +1,1153 @@
...
+        if (isBinaryPolylines(polylines)) {
+            if (
+                group.dashArray != null ||
+                props.getGroupDashArray?.(group) != null
+            ) {
+                console.warn(
+                    "PolylineGroupLayer: dashArray is not supported on BinaryPolylines groups; ignoring."
+                );
+            }
+            const srcVerts = polylines.positions.length / 3;
```
rubenthoms: Consider implementing the same guard here as in `buildBinaryData` - you could potentially get a fractional `totalVerts` value.

rubenthoms (approved):
One suggestions, one bug, and one missing guard. Otherwise, LGTM :+1: I will already approve it.

PR #2769, opened 2026-06-02
build: move deck.gl packages to peerDependencies

rubenthoms:
Moves `@deck.gl/*` and `@deck.gl-community/editable-layers` from dependencies to peerDependencies in subsurface-viewer, allowing consumers to control the deck.gl version. Also bumps the npm engine requirement to >=7, which is required for automatic peer dependency installation.


hkfb:
Consider that the `chore:` prefix doesn't build a new npm package. Perhaps `build:` is more appropriate.

hkfb (approved):

PR #2770, opened 2026-06-03
fix: add publishConfig for public access in multiple packages

hkfb:
access is "private" by default for scoped packages


nilsbon (dismissed):

w1nklr (approved):

w1nklr:
What does that change ?

hkfb:
> What does that change ?

The package failed to publish. Hopefully this will work. If not perhaps the NPM_TOKEN has expired.

PR #2775, opened 2026-06-05
fix: bump the deckgl group in /typescript to 9.3

w1nklr:
Bump deck.gl to 9.3.3 in /typescript.

Note that it is not compatible with deck.gl 9.2


hkfb (approved):

Anders2303:
I had some issues getting my dev environment set up, so I was unable to install the branch locally test it properly inside our Webviz project, but I looked over the changelogs for deckgl and lumagl, and there shouldn't be anything stopping us from going to 9.3, so we should be fine with this update.

However, I tested the branch a little bit with the storybook, and generally everything seems to be in order, but it seems that something is causing the `WellMarkers` story (?path=/story/subsurfaceviewer-well-markers-layer--well-markers) to have some sort infinite loop going on, and things freezes if you go there (interestingly enough, it only happens if you go to the explicit story, and not in the docs-overview page)

Anders2303:
Oh, the slowdown seems to happen in the PolyLines layer aswell, so might a problem in multiple layers 
?path=/story/subsurfaceviewer-polylines-layer--small-polylines-layer

Anders2303:
This was observed in Firefox, to be precise:

The browser prints this when I force-stop the scripts on the poly layer:
image

And this on the Map layer
image

w1nklr:
@Anders2303 
The (huge) slowdown you're experiencing is coming from the bump to storybook 9, and is completely unrelated to deck.gl.
It has been fixed by #2792, and the deck.gl PR branch does not yet have this fix :/

> After bump to storybook 9, most WellLogViewer stories became slow and froze. This was caused by well log sets/collections to be given as story arguments. With storybook 9, parsing huge arguments leads to freeze.
>
> Note: it does not make really sense to pack some huge data structure into a storybook argument; user will never be able to read/edit it.
>
> This has been fixed by removing the huge data from arguments and passed directly to the components, outside of the storybook arguments or rendering function.

I just checked the Polyline layer stories also, and don't have any issue :/

PR #2776, opened 2026-06-05
chore: bump storybook to version 9

w1nklr:
Bump storybook to version 9

Downgraded from storybook 10 which implies too much change.
Comes with a slight tsconfig change (mainly using ES202 instead of ES2015)


typescript/tsconfig.json
hkfb, line 1: Any potential breaking changes for package consumers due to these changes?

w1nklr: In don't think:
- node10 is the new name for node => no change
- all major browser support es2022 (https://www.w3schools.com/js/js_2022.asp)

hkfb (approved):

PR #2794, opened 2026-06-22
feat: fix WLV initial domain handling

w1nklr:
Use the domain prop if given to initialize the base content domain (previous approach forced it to the log range).
Introduce a visibleRange prop to initialize the visible domain range (previous approach forced it to the log range).

Prevent crashes if:
- empty log or log set array is provided
- if no template is specified
- if log/log set is undefined

Update the stories to specify the visibleRange instead of domain.


typescript/packages/well-log-viewer/src/components/WellLogView.tsx
```diff
@@ -1144,6 +1155,9 @@ export const argTypesWellLogViewProp = {
     },
 
     domain: {
+        description: "Initial base domain range",
```
hkfb, line 1164: Not clear what "base domain" means.

typescript/packages/well-log-viewer/src/components/DefaultSyncLogViewerLayout.tsx
```diff
@@ -19,14 +19,25 @@ export function defaultSidePanel(parent: SyncLogViewer): JSX.Element {
...
+                                readoutOptions={parent.props.readoutOptions}
+                                callbackManager={
+                                    parent.callbackManagers[iWellLog]
+                                }
+                            />
+                        );
+                    }
+                    return null;
+                }
             )}
```
hkfb, line 41: Is this potentially a breaking change? For example, could someone previously use `undefined` to mean en empty track?

typescript/packages/well-log-viewer/src/components/DefaultSyncLogViewerLayout.tsx
```diff
@@ -19,14 +19,25 @@ export function defaultSidePanel(parent: SyncLogViewer): JSX.Element {
...
-                        header={"Readout " + wellLogSets[0]?.header.well}
-                        readoutOptions={parent.props.readoutOptions}
-                        callbackManager={parent.callbackManagers[iWellLog]}
-                    />
-                )
+                (collection: WellLogSet[], iWellLog: number) => {
+                    // Handle only defined and non-empty collections
+                    const col = collection?.filter(
+                        (item) => item !== undefined
+                    );
```
hkfb, line 26: lodash's `_.compact` could replace this

typescript/packages/well-log-viewer/src/components/DefaultSyncLogViewerLayout.tsx
w1nklr: Until now, using `undefined` or `[]` led to a crash.

hkfb (approved):

PR #2805, opened 2026-07-08
fix: lighting issues

w1nklr:
Update local @luma.gl lighting code copy modified to support double side lighting.
The original code changed in @luma.gl 9.3.

### DIFF
--- a/typescript/packages/subsurface-viewer/src/layers/shader_modules/phong-lighting/phong-shaders-glsl.ts
+++ b/typescript/packages/subsurface-viewer/src/layers/shader_modules/phong-lighting/phong-shaders-glsl.ts
@@ -2,8 +2,14 @@
 // SPDX-License-Identifier: MIT
 // Copyright (c) vis.gl contributors
 
+// Webviz note: Copied this module from luma.gl.
+// https://github.com/visgl/luma.gl branch 9.3-release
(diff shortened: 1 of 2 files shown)


typescript/packages/subsurface-viewer/src/layers/shader_modules/phong-lighting/phong-shaders-glsl.ts
```diff
@@ -23,18 +28,32 @@ uniform phongMaterialUniforms {
 
 // Note modified for two sided lighting.
 vec3 lighting_getLightColor(vec3 surfaceColor, vec3 light_direction, vec3 view_direction, vec3 normal_worldspace, vec3 color) {
-  vec3 halfway_direction = normalize(light_direction + view_direction); 
+  vec3 halfway_direction = normalize(light_direction + view_direction);
 
+  // initial code
```
amrovic90, line 33: Any reason why this commented out code is kept?

w1nklr: To see the changes compared to original code.
This makes it easier when there is a new version of the code. Otherwise you have to find the old original code, check the diff with here to see the modifications, switch to the new original code and apply the changes.

amrovic90 (approved):

PR #2813, opened 2026-07-27
chore: add new story in SyncLogViewer with wellpicks

amrovic90:


typescript/packages/well-log-viewer/src/SyncLogViewer.stories.tsx
hkfb: Screenshot tests are disabled for all stories in this file, which means that your new story isn't covered by the screenshot tests. Could you selectively enable the test for your new story by moving the "no-screenshot-test" to the stories themselves?

You'll then also need to generate the screenshot by running the test update script, which is described in the readme files.

hkfb (approved):

w1nklr (commented):
Can you add also a story with pick where no pick is on all the wells to demonstrate the issue we have in PPA ?

w1nklr (changes requested):

w1nklr (approved):
We tested manually that it works with no pick existing on each well.

PR #2816, opened 2026-08-14
feat(WellLogViewer): add optional formatWellPickLabel callback for we…

hkfb:
Adds an opt-in `formatWellPickLabel` callback to `WellPickProps` that lets consumers customize how well-pick labels are rendered (e.g. two decimal places instead of the default whole-unit rounding). When the callback is not provided, the existing default behaviour is preserved exactly, so no other consumer of the shared component is affected.

### API

```ts
formatWellPickLabel?: (input: {
    horizon: string;                 // raw boundary (marker) name
    vPrimary: number | undefined;    // full-precision primary axis depth
    vSecondary: number | undefined;  // full-precision secondary axis depth
}) => Partial<{
    primary: string;                 // primary depth label
    secondary: string;               // secondary depth label
    horizon: string;                 // boundary label
}> | undefined | null;
```

The callback is invoked once per well pick boundary. It may return any subset of `{ primary, secondary, horizon }`; omitted (or `undefined`) keys keep the default formatting — `toFixed(0)` for the depths, the raw name for the boundary — and returning nothing keeps all defaults. The returned `horizon` affects the rendered text only: element names, colors, patterns and callbacks continue to use the raw boundary name. (The field is still called `horizon` for consistency with the existing `WellPick` fields it mirrors — it is not restricted to stratigraphic horizons.)

An object is used rather than positional `(markerName, depth)` arguments so that both depths are available and new fields can be added without a breaking change.

### Label escaping

Well-pick labels are interpolated into an HTML string passed to d3's `.html()`. Since the callback can now return arbitrary text, all three labels are escaped via a new `htmlEscape` utility before concatenation. This prevents broken markup for characters such as `<` and `&`, and script injection when label text originates from untrusted data. Plain labels are unaffected.

### Tests

- Unit tests for `defaultWellPickLabels`, `applyWellPickLabelFormatting`, `htmlEscape` and `generateColor`/`createColorGenerator`.
- Rendering tests asserting default labels, formatter-provided labels, and that formatter output is rendered as text and never as markup.
- A `WellPickLabelWithCustomFormatter` storybook story with its DOM snapshot.

Related to internal tracking ticket EP Connect #554887.

### Screenshot: `WellPickLabelWithCustomFormatter` story

Well pick labels rendered with the custom `formatWellPickLabel` callback: depths at 2 decimal places (e.g. `2788.00`) instead of the default whole-unit rounding, and upper-cased horizon names (e.g. `HOR_2`).


typescript/packages/well-log-viewer/src/components/WellLogView.tsx
```diff
@@ -329,6 +352,24 @@ export interface WellPickProps {
      */
     colorMapFunctions: ColormapFunction[];
     colorMapFunctionName: string; // "Stratigraphy" ..., "Step func", ...
+    /**
+     * Optional callback to override the default well-pick label formatting.
+     *
+     * It is called exactly once per well pick and receives the horizon name
+     * together with the full-precision primary and secondary depth values
+     * (either of which may be `undefined` or non-finite).
```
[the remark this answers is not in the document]
hkfb: Resolved: the PR description now documents the object-based signature (`{ horizon, vPrimary, vSecondary } -> Partial<{ primary, secondary, horizon }>`) instead of the mismatched `(markerName, depth)` wording. The object shape is kept — it exposes both depths in one call and lets us add fields later without a breaking change.

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
```diff
@@ -597,7 +684,7 @@ function addWellPickOverlay(instance: LogViewer, parent: WellLogView) {
                           "<span " +
                           styleText +
                           ">" +
-                          horizon +
+                          labels.horizon +
                           "</span>" +
```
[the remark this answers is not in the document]
hkfb: Fixed in 56126de0: added an `escapeWellPickLabel` helper and applied it to `txtPrimary`, `txtSecondary` and the horizon label before they're concatenated into the `.html(...)` strings (both the horizontal and vertical layout branches). Plain labels render identically to before; labels containing `<`, `>`, `&`, quotes, etc. are now rendered as literal text instead of being interpreted as markup. Added rendering tests asserting formatter output such as `<b>x</b>` and `A & B` shows up as text with no injected DOM elements.

hkfb:
**Follow-up on the Storybook Smoke Tests failures.**

I previously (in the PR description update) characterized the 3 failing story
snapshots (`Horizontal`, `OnInfoFilledEvent`, `Discrete`) as unrelated,
pre-existing flakiness. That was wrong — they **were** caused by this branch,
and I want to correct the record and explain the fix pushed in
89dcee71.

**Root cause:** `utils/generateColor.ts` hands out plot colors from an
8-entry palette via a module-level counter that is **never reset**:

```ts
let __iPlotColor = 0;
export function generateColor(): string {
    return __colors[__iPlotColor++ % __colors.length];
}
```

`utils/template.ts` calls it for every plot lacking an explicit color.
`@storybook/test-runner` does a single `page.goto(iframe.html)` per test
*file*, then renders every story in that file client-side in the same page —
so the counter accumulates across all stories in `WellLogViewer.stories.tsx`.
Adding `WellPickLabelWithCustomFormatter` (which uses `template1`, ~20
auto-colored plots) shifted the palette by +1 for every later story that
also relies on auto-generated colors — exactly the 3 `template2`-based
stories that come after it in file order.

**Fix:** the new story now resolves its own template colors up front with an
independent `createColorGenerator()` instance (new export alongside
`generateColor()`), so visiting it never advances the shared counter.
`getStyledTemplateTracks` gained an optional `nextColor` parameter
(defaulting to `generateColor`) to support this — existing callers are
unaffected. Verified locally: full `WellLogViewer.stories.tsx` (12 stories)
and the previously-failing 3 stories now pass unchanged, and no other
story's snapshot was touched.

The underlying bug (the unreset counter making the whole suite
order-dependent) is pre-existing and out of scope for this PR — any future
story insertion could re-trigger it elsewhere. Worth a separate issue if
maintainers agree; happy to file one.

hkfb:
**Update:** the color-order fix (89dcee71) worked as expected —
`WellLogViewer.stories.tsx` passed in this run's Storybook Smoke Tests job
(28.1s, no snapshot changes needed), and all 201 DOM/image snapshots in the
suite passed.

The job is still red, but for an unrelated reason: `GroupTreePlot/Demo ›
Default` and `› Resizable` failed with `page.screenshot: Protocol error
(Page.captureScreenshot): Unable to capture screenshot` — a Playwright/CDP
infra flake in the screenshot-stabilization retry loop
(`.storybook/test-runner.ts:21`), not a snapshot mismatch. This matches
flakiness I've also seen on `master` unrelated to this branch. Re-running
the job should clear it.

hkfb:
**Update after merging `upstream/master` (7f09e342):**

Merge was clean (only `.storybook/preview.tsx`, `.storybook/test-runner.ts`,
`package.json`/lockfile changed via upstream's own
`test(storybook): stabilize flaky GroupTreePlot smoke tests (#2829)` — none
of our files). Lint, typecheck, and the well-log-viewer Jest suite (15/15)
all pass post-merge.

`WellLogViewer.stories.tsx` has now passed cleanly in **3 separate CI runs**
(pre-merge and twice post-merge), confirming the color-order fix holds.

Storybook Smoke Tests is still red, but on two **different, unrelated**
stories each time — `SyncLogViewer` (layout-height snapshot diff) and
`CameraControlExamples` (image-snapshot diff), neither of which this PR
touches. Checking recent workflow runs across the repo: `master` passes
cleanly on 5/5 recent runs (including at this exact merge base), while several
unrelated branches (dependabot bumps, an unrelated bugfix branch) show the
same kind of intermittent single/double-story failures. That points to
environment-level flakiness in the Storybook Smoke Tests job rather than
anything introduced here — I won't keep re-triggering the workflow to avoid
burning CI minutes on it, but happy to re-run once more if a maintainer wants
another data point.

typescript/packages/well-log-viewer/src/utils/generateColor.ts
```diff
@@ -12,3 +12,12 @@ let __iPlotColor = 0;
 export function generateColor(): string {
     return __colors[__iPlotColor++ % __colors.length];
 }
+
+// Creates an independent color generator with its own counter, starting from
```
w1nklr: Could be a typescript comment to document the functions.
BTW, any tests ?
YCR

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
```diff
@@ -321,6 +321,29 @@ function addPinnedValueOverlay(instance: LogViewer, parent: WellLogView) {
...
+    primary: string;
+    secondary: string;
+    horizon: string;
+}
+
+/**
+ * Source values for a single well pick, passed to `formatWellPickLabel`.
+ * An object is used so that new fields can be added without a breaking change.
+ */
+export interface WellPickLabelInput {
```
w1nklr, line 344: Consider renaming the fields, to be more explicit:
- pickLabel or pickName (names come generally from boundaries, not limited to horizons)
- primaryDepth
- secondaryDepth

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
```diff
@@ -330,6 +353,24 @@ export interface WellPickProps {
      */
     colorMapFunctions: ColormapFunction[];
     colorMapFunctionName: string; // "Stratigraphy" ..., "Step func", ...
+    /**
+     * Optional callback to override the default well-pick label formatting.
+     *
+     * It is called exactly once per well pick and receives the horizon name
```
w1nklr: replace horizon by boundary in the doc

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
```diff
@@ -511,6 +553,63 @@ function posWellPickTitles(instance: LogViewer, parent: WellLogView) {
...
+    return Number.isFinite(value) ? (value as number).toFixed(0) : "";
+}
+
+/**
+ * Escapes a well-pick label so that it can be safely embedded in the HTML
+ * string used to render the label table. Labels may come from log data or from
+ * a consumer supplied `formatWellPickLabel` callback and must never be
+ * interpreted as markup.
+ */
+export function escapeWellPickLabel(label: string): string {
```
w1nklr: Rename to htmlEscape (could be in an utility function).

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
```diff
@@ -511,6 +553,63 @@ function posWellPickTitles(instance: LogViewer, parent: WellLogView) {
...
+    };
+}
+
+/**
+ * Resolves the labels of a single well pick: computes the defaults, calls
+ * `formatWellPickLabel` once (when given) and merges the returned partial over
+ * the defaults. A callback returning nothing, or fields that are `undefined`,
+ * leaves the corresponding defaults in place.
+ */
+export function resolveWellPickLabels(
```
w1nklr: resolve is too vague.

Why not formatWellPickLabel (ok, it collides with param) or applyWellPickFormatting ?

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
w1nklr: I realize these are coming from the poor WellPick field names :/

Let's postpone that for a follow-up (if we ever address it).

typescript/packages/well-log-viewer/src/utils/generateColor.ts
hkfb: Good call — done in 0ccfa16f. Both `generateColor()` and `createColorGenerator()` now have TSDoc explaining the palette, the shared never-reset counter (and the story-ordering hazard that motivated the factory), and how the factory differs. Added `utils/generateColor.test.ts` covering cycling/wrap-around, independence of generator instances, and that consuming from a generator doesn't advance the shared counter.

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
hkfb: Sounds good, postponing the WellPick field renames to a follow-up as agreed. No code change here.

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
hkfb: Done in 0ccfa16f — reworded the prose in `WellPickLabels`, `WellPickLabelInput` and `formatWellPickLabel` to use 'boundary' instead of 'horizon'. Kept the actual field name as `horizon` (per the note above about postponing field renames) and added a one-line explanation of why.

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
hkfb: Done in 0ccfa16f — moved it to a new `utils/htmlEscape.ts` exporting `htmlEscape()`, generalized the doc since it's no longer well-pick specific. Kept it out of `utils/index.ts` so it stays internal rather than becoming public API.

typescript/packages/well-log-viewer/src/components/WellLogView.tsx
hkfb: Good point, went with `applyWellPickLabelFormatting` — makes it explicit that it's the label being formatted (rather than something vaguer like 'resolve'). Done in 0ccfa16f.

w1nklr (approved):

PR #2822, opened 2026-08-26
feat: Make WellMarkersLayer better in scaled views.

Anders2303:
* Made the layer avoid stretching discs when using vertical scale. Made disc angles match scaled trajectory
* Added storybook example of markers along a wellbore path
* Added two wells layer utils to index export, since they would be library consumers
  * `getAziAndInclForSegment` – computes the incl/azi for a given segment
  * `getSegmentIndicesForMd` – computes the start and end indices, along with fraction length for a segment that contains a given MD

| Before | After | 
|--- | --- |
image | image

### Note:
The storybook example shows a MD-marker when hovering the trajectory, but note that the marker position is currently wrong due to issues within the wells-layer (see #2823 )


typescript/packages/subsurface-viewer/src/storybook/examples/miscExamples.stories.tsx
```diff
@@ -399,3 +417,224 @@ export const MixedLayerDefinitions: StoryObj<typeof SubsurfaceViewer> = {
     },
     parameters: parameters,
 };
+
+type MarkerDataWithMd = WellMarkerDataT & { md: number };
+type MarkersAlongWellPath = SubsurfaceViewerProps & {
+    wellIdx: number;
+    applyModelMatrix: boolean;
+};
+export const MarkersAlongWellPath: StoryObj<MarkersAlongWellPath> = {
```
hkfb: Perhaps this story should be in `WellMarkersLayer.stories.tsx`?

hkfb (approved):

typescript/packages/subsurface-viewer/src/storybook/examples/miscExamples.stories.tsx
Anders2303: I'll move it, sure. I was thinking it was in "Examples" since it used multiple layers; for future reference, what kind of stories **should** be in this file?

hkfb: I think we've used the Examples section for things that don't fit other places. Perhaps "Misc" is a more appropriate name than "Examples". I think your example fits in the WellMarkerLayer section because it demonstrates well the marker along the well trajectory use case.

PR #2829, opened 2026-09-01
test(storybook): stabilize flaky GroupTreePlot smoke tests

### LINKED ISSUE #2828: Flaky Storybook smoke tests: GroupTreePlot screenshot capture fails intermittently
## Summary

The `Storybook Smoke Tests` job (`typescript.yml`) is intermittently flaky. Across the 20 most recent
failing runs of the `typescript.yml` workflow, **24 of 28** individual test failures are:

```
page.screenshot: Protocol error (Page.captureScreenshot): Unable to capture screenshot
Call log:
  - taking page screenshot
  - waiting for fonts to load...
  - fonts loaded

  at screenshot (typescript/.storybook/test-runner.ts:21:46)
  at screenshotTest (typescript/.storybook/test-runner.ts:85:19)
```

thrown from the `while (!stable)` screenshot-stability loop in `typescript/.storybook/test-runner.ts`.

**All 24 of these failures are one of three stories:**
- `GroupTreePlot/Demo › Default`
- `GroupTreePlot/Demo › Resizable`
- `GroupTreePlot/Interactive Demo › InteractiveWithTreeTypeSwitch`

These are the only stories in the repo backed by continuously-running `motion` (Framer Motion) spring
animations on SVG geometry (`TREE_MOTION_TRANSITION = { type: "spring", bounce: 0, duration: 0.5 }` in
`packages/group-tree-plot/src/hooks/useCollapseMotionProps.ts`, applied via `motion.g` / `motion.path`
/ `motion.text` inside `AnimatePresence`). Because the animation drives SVG attributes on the main
thread (not WAAPI), the Chromium compositor is repeatedly non-idle while
`test-runner.ts`'s stability loop polls `page.screenshot()`, and this appears to intermittently trigger
a CDP `Page.captureScreenshot` protocol error rather than a normal image mismatch. These stories are
tagged `no-dom-test`, so the screenshot is their only assertion — a thrown protocol error fails the
test outright with no useful diff to inspect.

The remaining 4 of 28 failures in the sample are `WellLogViewer` DOM-snapshot mismatches on an open,
unrelated PR (#2816) and appear to be a genuine intended diff rather than flake.

## Contributing weaknesses in the harness

Independent of the animation issue, `typescript/.storybook/test-runner.ts`'s `screenshotTest`:
- runs an **unbounded** `while (!stable)` loop with no max-attempts guard, so a story that never
  settles burns the entire 60s Jest `testTimeout` instead of failing fast;
- polls every **10 000 ms**, and since the initial `previousScreenshot` is an empty buffer, every
  single story pays at least one 10s sleep — this both slows the whole ~212-story suite considerably
  and leaves only a handful of samples inside the 60s timeout for a slow-settling story;
- calls `page.screenshot()` with no `animations: "disabled"`, no `caret: "hide"`, and no timeout;
- has no retry around transient CDP errors.

## Proposed fix

We'll open a PR from our fork (`hkfb/webviz-subsurface-components`) that:
1. Adds a test-only mechanism to skip `motion` animations during Storybook test-runner execution
   (via `MotionGlobalConfig.skipAnimations`, gated behind a flag the test runner sets per-page — **not**
   `prefers-reduced-motion`, so manual Storybook browsing and the shipped component are unaffected).
2. Hardens `test-runner.ts`: bounds the stability loop, shortens the poll interval, adds
   `animations: "disabled"` / `caret: "hide"` to the screenshot call, and retries transient
   `Page.captureScreenshot` protocol errors.

Happy to have this reviewed against a different design if maintainers prefer a different mechanism for
disabling animations under test.

hkfb:
## Summary

The `Storybook Smoke Tests` CI job is intermittently flaky. Fixes #2828, which documents the
evidence: across the 20 most recent failing `typescript.yml` runs, **24 of 28** individual test
failures were

```
page.screenshot: Protocol error (Page.captureScreenshot): Unable to capture screenshot
```

thrown from the screenshot-stability loop in `typescript/.storybook/test-runner.ts`, and **all 24**
were one of the three `GroupTreePlot` stories (`Default`, `Resizable`,
`InteractiveWithTreeTypeSwitch`) — the only stories driven by continuously-running `motion` (Framer
Motion) spring animations on SVG geometry. The animation keeps the Chromium compositor busy while the
harness polls `page.screenshot()` for a settled frame, which intermittently trips a transient CDP
capture error.

## Changes

- **`.storybook/preview.tsx`**: add a decorator that sets `MotionGlobalConfig.skipAnimations` from a
  `window.__WEBVIZ_SKIP_MOTION__` flag. That flag is only ever set by the test runner (see below), so
  this is deliberately **not** wired to `prefers-reduced-motion` or a build-time env var — manual
  Storybook browsing and the shipped component's real animation behaviour are unaffected.
- **`.storybook/test-runner.ts`**:
  - a new `preVisit` hook sets the flag via `page.evaluate` before each story renders (the test runner
    navigates to `iframe.html` once and reuses that page for every story, so `preview.tsx`'s module
    body runs before any hook — the flag has to be set on the live page, not read at import time);
  - the screenshot and DOM-snapshot stability loops are unified into one bounded `waitUntilStable`
    helper (replacing the previous **unbounded** `while` loop with a **10 000 ms** poll — every story
    paid at least one 10s sleep; the new poll is 500 ms with a 20-attempt ceiling);
  - `page.screenshot()` now passes `animations: "disabled"`, `caret: "hide"`, and an explicit timeout;
  - a small retry wraps the capture to tolerate a transient `Page.captureScreenshot` protocol error
    instead of failing the test outright.
- **`typescript/package.json`**: add `motion` as an explicit devDependency (it was previously only
  hoisted transitively from `@webviz/group-tree-plot`).

## Verification

- `npm run validate` (typecheck + lint) — clean.
- `npx nx run-many -t test_correctness` — 292 passed, 6 pre-existing skips, 0 failures.
- Built Storybook locally and ran `test-storybook` directly against the static build (mirroring the CI
  `smoke_tests` job): **5 consecutive clean runs** of the `GroupTreePlot` stories (previously flaky,
  now ~5s each with no capture errors), plus a full 211-story pass with **no** `GroupTreePlot`
  failures and **no** `captureScreenshot` errors. The full run had 3 unrelated `SubsurfaceViewer`
  image-snapshot diffs (`WellsLayer`/`WellLabelLayer`, ~1.5–2.5%) that reproduce against this
  devcontainer's own Chromium/font stack rather than the pinned `mcr.microsoft.com/playwright:v1.60.0-jammy`
  CI image — a known environment-fidelity limitation, not a regression from this change (these files
  are untouched here).
- No `__image_snapshots__` baselines needed regenerating: skipping the animation snaps to the same
  final frame the animation already converges to, so `GroupTreePlot`'s existing baselines matched
  exactly in all runs above.

CI's `Storybook Smoke Tests` job (Docker + the pinned Playwright image) will provide the
authoritative screenshot-comparison result for this PR.

### DIFF
--- a/typescript/.storybook/preview.tsx
+++ b/typescript/.storybook/preview.tsx
@@ -1,4 +1,35 @@
-import type { Preview } from "@storybook/react-webpack5";
+import type { Decorator, Preview } from "@storybook/react-webpack5";
+import { MotionGlobalConfig } from "motion/react";
+
+declare global {
+    interface Window {
+        /**
+         * Test-only flag set by the Storybook test-runner (see
+         * `.storybook/test-runner.ts`, `preVisit`) before each story is
+         * visited. When set, Framer Motion animations are skipped so that
+         * screenshot and DOM snapshot assertions run against a settled
+         * frame instead of racing an in-flight spring animation (this is
+         * what made the `GroupTreePlot` stories intermittently fail with
+         * `Page.captureScreenshot` protocol errors).
+         *
+         * Deliberately *not* driven by `prefers-reduced-motion`: gating on
+         * that media query would also disable animations for anyone
+         * manually browsing Storybook with that OS setting enabled, which
+         * would misrepresent the real, motion-enabled component behaviour.
+         * This flag is only ever set by Playwright, so manual Storybook use
+         * is unaffected.
+         */
+        __WEBVIZ_SKIP_MOTION__?: boolean;
+    }
+}
+
+const withMotionTestOverride: Decorator = (Story) => {
+    MotionGlobalConfig.skipAnimations = Boolean(
+        typeof window !== "undefined" && window.__WEBVIZ_SKIP_MOTION__
+    );
+
+    return Story();
+};
 
 const preview: Preview = {
     parameters: {
@@ -18,6 +49,8 @@ const preview: Preview = {
     },
 
     tags: ["autodocs"],
+
+    decorators: [withMotionTestOverride],
 };
 
 export default preview;
--- a/typescript/.storybook/test-runner.ts
+++ b/typescript/.storybook/test-runner.ts
@@ -10,27 +10,72 @@ import {
 // https://github.com/mapbox/pixelmatch#pixelmatchimg1-img2-output-width-height-options
 const customDiffConfig = {};
 
-const screenshotTest = async (page: Page, context: TestContext) => {
-    let previousScreenshot: Buffer = Buffer.from("");
+/**
+ * Polls `sample()` until two consecutive samples are equal (by `isEqual`), or
+ * until `maxAttempts` is reached, whichever comes first. Returns the last
+ * sample taken.
+ *
+ * Used to wait out stories whose rendering settles asynchronously (debounced
+ * `ResizeObserver` layout, animations, etc.) before asserting a screenshot or
+ * DOM snapshot against them. Bounded so that a story which never settles
+ * fails fast with a diff to inspect, instead of silently consuming the whole
+ * Jest `testTimeout`.
+ */
+async function waitUntilStable<T>(
+    sample: () => Promise<T>,
+    isEqual: (a: T, b: T) => boolean,
+    { maxAttempts, poll }: { maxAttempts: number; poll: number }
+): Promise<T> {
+    let previous: T = await sample();
 
-    let stable = false;
+    for (let attempt = 1; attempt < maxAttempts; attempt++) {
+        await new Promise((resolve) => setTimeout(resolve, poll));
 
-    const poll = 10000;
-
-    while (!stable) {
-        const currentScreenshot = await page.screenshot();
-        if (currentScreenshot.equals(previousScreenshot)) {
-            stable = true;
-        } else {
-            previousScreenshot = currentScreenshot;
+        const current = await sample();
+        if (isEqual(current, previous)) {
+            return current;
         }
+        previous = current;
+    }
 
-        if (!stable) {
-            await page.waitForTimeout(poll);
+    return previous;
+}
+
+/**
+ * `page.screenshot()` can intermittently throw
+ * `Protocol error (Page.captureScreenshot): Unable to capture screenshot`
+ * when Chromium's compositor is captured mid-frame (observed almost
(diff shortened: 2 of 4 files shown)


w1nklr (approved):

hkfb:
## Follow-up: CI validation results

**Update:** the first CI run on this branch (before commit 294e8d0b) hit the *same* `Page.captureScreenshot`
protocol error on 2 of the 3 GroupTreePlot stories this PR targets, despite 6/6 clean local runs during
development. Flagging that discrepancy transparently rather than hiding it.

**Root-cause investigation:** replicated the smoke-test job's exact environment locally — the pinned
`mcr.microsoft.com/playwright:v1.60.0-jammy` image, `--user 1001`, `--shm-size=1gb`, and the full
211-story suite (not just GroupTreePlot) — and ran it twice. **Zero `captureScreenshot` errors and zero
GroupTreePlot failures** in both runs. This, together with the very intermittent nature of the underlying
Chromium CDP bug (that's the whole reason this PR exists), points to the first CI run being a rare
transient recurrence rather than a gap in the animation-skip fix itself.

**Follow-up commit (294e8d0b):** added defense-in-depth anyway — a short delay between
`screenshotWithRetry` attempts (instant back-to-back retries don't reliably clear the CDP error if the
compositor is still mid-frame) and bumped attempts 3 → 4.

**Result after the follow-up commit:** the next CI run on this branch (33496744705) passed with
**zero GroupTreePlot / captureScreenshot failures**. The only failure was an unrelated, previously-unseen
flake in `SubsurfaceViewer/Examples/Camera › OverrideControllerSettings › play-test` (a 62% pixel-diff,
image-snapshot mismatch, not a CDP protocol error) — a different package, different failure mode, and not
touched by this PR's diff, so left out of scope here. Checked `master`'s recent failure history and this
specific story hasn't appeared there; it may be a separate animation-driven-screenshot flake worth its own
issue, but is not something to fix in this PR.

**Net evidence for the original bug:** 2/2 clean full-suite CI runs post-fix vs. it being ~86% of all
smoke-test failures pre-fix. Still a draft PR pending human review — not promoting to ready-for-review.

Issue #2833, opened 2026-09-02
WellLogView: ResizeObserver feedback loop causes ~2s layout settle ramp (not an animation)


hkfb:
## Summary

`WellLogView`'s layout does not converge to its final size in a single pass. A `ResizeObserver`
callback measures the container, writes a new inline `height`/`width` style, which itself triggers
another resize, which triggers the observer again — producing a visible ramp that takes roughly
2 seconds (in an unthrottled browser) to settle, in small ~5px steps every ~250ms.

This is a genuine layout bug affecting real users too (not just tests): anyone opening a
`WellLogViewer`/`SyncLogViewer` will see the track heights visibly creep to their final size for
about 2 seconds after mount/resize, rather than rendering at the correct size immediately.

## Evidence

Probed the `WellLogViewer/Demo/SyncLogViewer` → `DiscreteLogsWithIndividualVisibleRange` story
directly via Playwright (no test harness involved — plain repeated `page.evaluate` sampling of
`getBoundingClientRect()`/computed style every 250ms):

```
t=0ms     outer=624.0  scale-track=553px
t=250ms   outer=629.0  scale-track=558px
t=500ms   outer=634.0  scale-track=563px
t=750ms   outer=639.0  scale-track=568px
t=1000ms  outer=644.0  scale-track=573px
t=1250ms  outer=649.0  scale-track=578px
t=1500ms  outer=654.0  scale-track=583px
t=1750ms  outer=659.0  scale-track=588px
t=2000ms  outer=662.4  scale-track=593px
t=2250ms  outer=662.4  scale-track=596px   <-- settled (0.92 * 720 = 662.4)
```

Confirmed this is **not** an animation: `document.getAnimations()` returns an empty array
throughout, and there are no matching CSS transitions on the relevant elements. Blocking `.woff2`
font loads via route interception did not change the ramp, ruling out a font-metrics race.

## Root cause

`Scroller`'s `ResizeObserver` callback measures `offsetWidth`/`offsetHeight`, subtracts a
freshly-measured scrollbar size (`getScrollbarSizes()`, which itself creates/removes a probe
`<div>` in `document.body` on every callback), and writes new inline `width`/`height` styles onto a
child `content` div. Writing those styles re-triggers layout, which re-triggers the observer,
converging only gradually instead of settling in one measurement.

`WellLogView.tsx`'s own `ResizeObserver` (around the `onContentRescale()` / `posWellPickTitles()`
[...]

PR #1918, opened 2024-02-13
test: enhance smoke tests with screenshots

hkfb:
Adds screenshot comparison to all stories covered by the Storybook test-runner, except stories tagged with "no-test".

PR #2834, opened 2026-09-02
test(storybook): require a genuine stable window before snapshotting

hkfb:
## Summary

Phase 2 of stabilizing the `@webviz` smoke tests (follow-up to #2829, which fixed the GroupTreePlot
`Page.captureScreenshot` flake). Since #2829 merged, two other flake classes surfaced:

1. **A regression I introduced in #2829.** That PR shortened the screenshot stability poll from
   10,000ms → 500ms while keeping a "2 consecutive equal samples" exit check. Measured settle
   profiles show some stories plateau mid-render for longer than one 500ms poll (up to ~750ms on
   `SubsurfaceViewer/Examples/Camera › OverrideControllerSettings`), which can satisfy "2 equal
   samples" while the frame is still mid-render. This surfaced once on CI
   (run `33496744705`, on the #2829 branch) as a 62% pixel-diff — never seen on `master`.
2. **A pre-existing, genuine flake in `WellLogViewer/Demo/SyncLogViewer`.** Reproduces on `master`'s
   own scheduled CI runs, not just PR branches (e.g. `33573704504`). Root cause: `Scroller`/
   `WellLogView`'s `ResizeObserver` write their own measured size back as an inline style, which
   re-triggers the observer, converging gradually over ~2s in ~5px/250ms steps — not an animation
   (`document.getAnimations()` is empty throughout). Filed as #2833 with the full measurement
   series; **not fixed here** (harness-only scope — the shipped component is unchanged). The value
   CI captured (`588px`) is literally one of the measured intermediate ramp steps.

Both share the same underlying harness weakness: `waitUntilStable`'s "first 2 equal samples" exit
condition is not robust against plateaus/ramps that happen to sit still for exactly one poll
interval.

## Fix (harness-only, `typescript/.storybook/test-runner.ts`)

- `waitUntilStable` now accepts `requiredStableSamples` (default 2, unchanged for any future caller
  that doesn't opt in). `screenshotTest` and `domSnapshotTest` both now require **5 consecutive
  identical samples** at a 500ms poll — a continuous ~2s stable window, comfortably above the
  longest observed mid-render plateau (~750ms) and the WellLogViewer ramp's per-step cadence
  (~250ms).
- `domSnapshotTest` additionally waits for a cheap in-page `MutationObserver`-based quiescence gate
  (no DOM mutations for a continuous 750ms, bounded to 5s) before starting the sampling loop —
  directly targeting the `ResizeObserver`-driven style-write loop, at near-zero cost relative to
  screenshot/DOM-string sampling.

No changes to any shipped package — `Scroller`/`WellLogView`'s actual layout behavior is untouched;
see #2833 for that follow-up.

## Validation

- `npm run validate` (typecheck + lint) clean; `npx prettier --check` clean.
- Instrumented the actual harness code path (temporary diagnostic logging, not committed) inside a
  container replicating the `smoke_tests` job's exact image/user/shm-size under CPU throttling, to
  directly observe the old "2 equal samples" logic's behavior and confirm the new logic's stability
  window holds under contention.
- Ran the previously-flaky `WellLogViewer/SyncLogViewer` stories repeatedly, both throttled
  (`--cpus=0.3`, matching CI's container config) and unthrottled — all passed.
- Ran the full 211-story local suite: 208 passed, including every GroupTreePlot and
  WellLogViewer/SyncLogViewer story. The 3 remaining failures (`WellsLayer`/`WellLabelLayer` image
  diffs) are pre-existing environment-fidelity artifacts — confirmed byte-identical failures against
  the unmodified baseline (this devcontainer's Chromium/font stack vs. the pinned CI image),
  unrelated to this change.
- **Caveat, stated plainly:** despite substantial effort (CPU-throttled Docker repro at multiple
  throttle levels, direct in-page probing, and instrumenting the real harness code path), I was not
  able to force a reliable local repro of the WellLogViewer flake actually *failing* under the old
  code — uniform CPU throttling slows the whole ramp proportionally rather than reproducing the
  specific transient contention pattern needed to catch two genuinely-equal 500ms-apart samples
  mid-ramp. The fix is grounded in direct measurement of the ramp and CI's own captured value
  matching a measured intermediate step, not in a forced side-by-side repro. Flagging this
  transparently rather than overstating confidence.

## Related

- Supersedes the screenshot-poll behavior from #2829 (fixes the regression it introduced).
- Fixes the smoke-test symptom of #2833 (opened alongside this PR) without touching the underlying
  component bug described there.
- Continuation of ADO #564787.

This PR is opened as a **draft**; CI E2E/smoke-test validation is still pending.


hkfb:
## CI result: 210/211 passed — one *different*, previously-unseen failure surfaced

CI run [33631799124](https://github.com/equinor/webviz-subsurface-components/actions/runs/33631799124)
on this branch's commit (`6828ad8`):

- ✅ Every GroupTreePlot story (the #2829 flake) — 0 `captureScreenshot` errors.
- ✅ Every `SubsurfaceViewer/Examples/Camera` story, including `OverrideControllerSettings` (the
  regression this PR fixes) — 0 pixel-diff failures.
- ✅ `WellLogViewer/Demo/SyncLogViewer › DiscreteLogsWithIndividualVisibleRange` (the original,
  root-caused height-ramp flake this PR targets) — passed.
- ❌ `WellLogViewer/Demo/SyncLogViewer › DiscreteLogsWithUndefined` — failed, but with a
  **different signature**: a pure attribute-*order* mismatch (`style` before vs. after
  `transform`/`font-size` on one `<text class="scale-range">` element, same values), consistent
  across all 3 jest retries. Not a timing race this stability fix (or any "wait longer") can
  address — see the addendum on
  [#2833](https://github.com/equinor/webviz-subsurface-components/issues/2833#issuecomment-5509916075)
  for the full analysis. Did not reproduce locally across 4+ runs on the identical pinned CI Docker
  image.

Net: this PR's two targeted fixes both held under CI. The one CI failure is a distinct,
newly-discovered issue, out of scope for this harness-only stability PR (documented on #2833 rather
than fixed here, since a real fix means either normalizing snapshot attribute order or restructuring
`scale-legend.ts`'s render sequencing - more invasive than this PR's scope). Flagging plainly rather
than re-running until green: leaving this PR's CI red is more honest than hiding a real, if
unrelated, finding.

Still a draft, per policy - not promoting to ready for review.

hkfb:
## Correction + a third change: forced-remount retries

**Correction to my previous comment:** I wrote that the `DiscreteLogsWithUndefined` failure was
"consistent across all 3 jest retries". That claim wasn't actually supported by the log - the two
occurrences I cited were the failure block and the end-of-run summary recap of the **same single
execution** (both show `45.599 s`), not two separate attempts. The log doesn't show whether retries
ran at all. Investigating that gap turned up a real bug in the harness, fixed by the commit just
pushed.

### What I found

`jest.retryTimes(3)` has been enabled since #1918 - well before either flake this PR addresses. But
it's currently a **no-op** for any flake caused by mount-time non-determinism: Storybook's
`setCurrentStory` channel event (what navigates to a story) re-renders **in place**, it doesn't
remount. A Jest retry re-invokes the same test body, which re-emits `setCurrentStory` for the same
story id - reusing the identical DOM nodes from the failed attempt.

Verified directly with a read-only Playwright probe (not committed): tagged a story's DOM nodes with
a marker attribute, then re-rendered the same story exactly as a retry would - **all markers
survived**, proving the nodes are reused rather than recreated. The same probe against Storybook's
own `forceRemount` event (used by its toolbar "remount" button) showed 0 markers survived - it
genuinely tears down and rebuilds.

This plausibly explains the `DiscreteLogsWithUndefined` failure: only **1 of 198** DOM snapshots
differed in that run - a deterministic environment cause (browser/library version, serializer) would
flip every `scale-range` element the same way, not one. This looks like true mount-time
non-determinism, which no number of the existing retries could ever have cleared.

### Fix

Track attempts per story id in `postVisit`; on any retry (attempt > 1), force a remount and await
`storyRendered` before running the existing stability waits. Zero cost on green runs (retries only
execute after a failure).

**Validated that it actually fires**, not just in theory: temporarily corrupted one committed DOM
snapshot to force a guaranteed, permanent mismatch, added temporary diagnostic logging (both
reverted afterward, not committed), and confirmed `forceRemount` fired before every retry attempt
(2, 3, and 4, since `retryTimes(3)` allows up to 4 total attempts) - and that the test still
correctly failed throughout, i.e. remounting doesn't mask a genuinely persistent mismatch.

Full local suite still shows the same 208/211 passing (3 pre-existing, unrelated `WellsLayer`/
`WellLabelLayer` environment-fidelity failures).

Pushed as `40b03bf0`. This is a general harness improvement independent of the specific flake, so I'm
keeping it in this PR rather than a separate one, per your direction. Full analysis, including a
retry-probe methodology writeup, is also being added to #2833.

Still a draft, per policy.

hkfb:
**Update: the forced-remount retry had a regression, now fixed.**

The previous commit (`40b03bf0`) called `forceRemount` on *every* retry attempt (`attempts > 1`). CI on that commit then failed with a new failure: `WellLogViewer/Demo/SyncLogViewer › Default` DOM snapshot showing drastically truncated content, plus an unrelated `WellsLayer › Wells3d` hard 60s timeout in the same run — the whole suite also took notably longer (614s vs. 434s previously), pointing at heavy runner contention that run.

**Root cause:** `forceRemount` tears down and rebuilds the DOM, which resets the WellLogViewer/Scroller's `ResizeObserver`-driven layout settle ramp back to its unsettled starting point (this ramp is the same pre-existing flake the two-stage stability criterion earlier in this PR already guards against — it can take far longer to converge under CPU contention). Remounting on *every* retry meant a slow-settling story had to re-climb its entire ramp from scratch inside each retry's bounded wait budget, rather than accumulating real settle time across attempts — under sustained contention this could make every retry fail identically, for a different reason than the flake `forceRemount` exists to fix.

**Fix (`9375d208`):** only force-remount once, on the first retry (`attempts === 2`), not on every retry. This still gives mount-time non-determinism one genuine fresh, independent draw, while later retries keep observing the same already-remounted story's continued natural settling instead of resetting the clock again.

**Validated:**
- `npm run validate` + `prettier --check`: clean.
- Confirmed via a temporary corrupted-snapshot experiment that `forceRemount` now fires on attempt 2 only (not 3 or 4), and that a genuinely persistent mismatch still fails on every attempt (remounting doesn't mask real regressions).
- Full local suite (212 stories, after picking up the `master` merge): 209 passed, same 3 pre-existing unrelated `WellsLayer`/`WellLabelLayer` environment-fidelity failures, no new regressions.
- Latest CI run on `9375d208`: 210/212 passed. The **`SyncLogViewer/Default` regression did not recur.** The two remaining failures are unrelated pre-existing flake classes: a `linearGradient` id-ordering artifact on a brand-new story from the merged-in master feature (confirmed master's own CI at that commit passed cleanly, so this is intermittent/test-order-dependent, not a bug in this PR), and a `WellsLayer` image-diff (same environment-fidelity flake family already noted, just a different specific story this run).

This PR remains a draft. Given the intermittent nature of what's left, I'd treat one more clean run as reassuring but not conclusive — happy to keep monitoring if useful.

hkfb:
Pushed two more fixes:

- **Gradient ids** normalized sequentially per snapshot before `toMatchSnapshot()` — the raw counter leaks unrelated page/render history (autodocs pre-rendering) into the assertion.
- **Color counter** (`generateColor()`) reset via a test-only flag/decorator, same pattern as the motion-skip fix. Fixes intermittent `red`/`orange` mismatches on `WellLogView`.

Both harness-only, no shipped-code behavior change. Baselines regenerated, diff-verified to only touch ids/colors.

`WellsLayer / UnfoldedProjection` image diff: attempted repro under CPU throttle in the pinned CI image — didn't reproduce (got timeouts instead). Left open, undiagnosed.

Local suite: 209-212/212, remaining failures are known devcontainer image-diff noise (unrelated to CI).

hkfb:
Follow-up on the gradient ids: moved the fix to the source. `gradientfill-plot-legend.ts` now keys each gradient's id off the `GradientFillPlot` instance (WeakMap) instead of an incrementing counter, so repeated redraws of the same plot reuse the same id. Kept the harness normalization as a safety net for remaining docs-page-prerender offset. Validated: 3 snapshots regenerated (id-only diff), 2 clean re-runs, full suite 209/212 (only known devcontainer image-diff flakes).

hkfb:
The previous CI run's only failure was `SubsurfaceViewer/WellsLayer/UnfoldedProjection` (failure D from an earlier phase) — and this time the new stability diagnostics caught it directly: `never reached 5 consecutive stable samples within the poll budget`. So D is now confirmed as a genuine mid-render capture under CI contention, not an environment/font issue. Raised `screenshotTest`'s `maxAttempts` 20→40 (zero cost on stories that already stabilize early). Re-running CI now.

hkfb:
CI run 33766263390: all green, including Storybook Smoke Tests (212/212). Both remaining failures from the prior run are addressed: gradient ids now stable (source fix), and UnfoldedProjection passed with the raised stability budget. One clean run isn't proof for an intermittent contention-driven case, but it's a good sign.

hkfb:
Follow-up on the CI failure: `SyncLogViewer/Default` failed with no stability warning — the DOM was quiet/consistent but still wrong (its `wellpickFlatting` rescale is applied via an event-driven callback chain that can race and silently never re-apply). No amount of waiting fixes a stably-wrong result, only a fresh mount does — but `postVisit` only force-remounts once, so 2 of 3 retries were re-asserting the same broken DOM. Added an opt-in `remount-every-retry` tag for this story (remounts on every retry, not just the first); left the default once-per-story policy alone since it's still correct for slow-settling flakes. Also added failure-only diagnostics distinguishing 'never stabilized' vs 'stabilized but wrong' for future occurrences. This race never reproduced locally, so I can't claim it's proven fixed — only that CI (34112953510) is now all green (212/212), and the fix gives 3 independent draws instead of 1.

hkfb:
Gradient ids no longer come from a never-reset module counter, so they don't encode how many plots were built earlier in the page. They're now derived from the plot and its own stops (`grad-<plot.id>[-<suffix>]-<hash>`), which makes them independent of render order, autodocs pre-render, sharding and retries. Identical gradients share one `<defs>` rather than emitting a duplicate.

This removes the reason the harness had to renumber ids before snapshotting, so `normalizeGradientIds` is gone and the DOM snapshots now assert the real ids.

Two checks worth noting:
- Canonicalising ids in the old and new snapshots and diffing yields no differences — only id strings moved, no stop/colour/reference changed.
- `Default` and `WellPickLabelWithCustomFormatter`, at different positions in the same file, now record identical raw ids. Under the counter the later one necessarily got higher numbers.

Full local suite 209/212 — the 3 known devcontainer-only WellsLayer/WellLabelLayer image flakes, same profile as before. CI green.​‌‍​

amrovic90 (approved):

w1nklr (approved):

PR #2847, opened 2026-09-08
test(storybook): make synthetic story data deterministic

hkfb:
Synthetic story data was generated from a module-level seeded RNG, seeded once at import and never reset:

```ts
const math = create(all, { randomSeed: "1984" });
const randomFunc = math?.random ? math.random : Math.random;
```

The test runner renders every story into a single page, so that generator's state carries across stories. The wells a story generates depend on how many numbers earlier stories drew — a count that varies with story order, autodocs pre-render, worker sharding and retries. Committed image baselines therefore encoded an arbitrary offset, and broke whenever story scheduling changed.

This is why master is currently red on `SubsurfaceViewer / Wells Layer › UnfoldedProjection`. The trigger was #2841, which changed no rendering code and left the image-comparison stack byte-identical — it changed jest's test-file ordering. The same failure reproduces byte-identically (1.6495225694444444%, 15202 px) on master and on unrelated PRs, and is immune to waiting or remounting, because the captured frame is stable, just generated from different data.

## Change

Seed a fresh generator per call in `createSyntheticWellCollection` / `createSyntheticWellHeads` and thread it through the helpers, so output is a function of the arguments alone. `TriangleLayer`'s `buildTrgl` draws per render and gets the same treatment.

Audited the other seeded generators (`miscExamples`, `WellMarkersLayer`, `PolylinesLayer`, `Grid3DLayer`, `PointsLayer`) — those draw only at module scope, so they are already deterministic and are left alone.

Baselines for the 8 affected stories are regenerated. They are the only ones that changed.

## Validation

Run in `mcr.microsoft.com/playwright:v1.60.0-jammy`, matching the CI job.

- Regenerating rewrote **only** synthetic-well and triangle stories; the other 186 snapshots matched the committed baselines untouched. That both confirms the change is scoped and shows this environment reproduces CI rendering.
- Order independence: `WellLabelLayer` passes both standalone (`--maxWorkers=1`) and in the full suite. Before the change those two orders produce different geometry.
- Full suite: **212/212**. Several image stories that failed persistently in local runs before this change now pass.

Related: #2833


hkfb:
Storybook renders every story into one page, so the module-level seeded generators in
`wellSynthesis.ts` and `TriangleLayer.stories.tsx` never reset: each story's wells and geometry
depended on how many numbers earlier stories had drawn. That draw count varies with story order,
sharding and retries, which is why these image diffs were stable per run but changed between runs,
and why waiting longer or remounting never helped.

Seeding a fresh generator per call makes the output a function of the arguments alone. Baselines for
the affected synthetic-well and triangle stories are regenerated in the pinned CI image.

`wells-layer--wells-3-d` was deliberately **not** regenerated — it renders static `volve_wells.json`,
so this change cannot affect it. It is a separate pre-existing flake.

A follow-up commit extracts the seeded-generator boilerplate, which was duplicated across 7 storybook
modules, into `util/random.ts`. Each caller keeps its own seed and placement, so no generated value
changes — confirmed by the suite producing zero baseline changes.

Worth flagging separately: `jest.retryTimes` does not rescue `toMatchImageSnapshot`. An earlier run
here had all 212 tests passing but still failed on `Snapshots: 1 failed`, because failed attempts are
counted even when a retry succeeds.​‌‍​

typescript/packages/subsurface-viewer/src/storybook/util/wellSynthesis.ts
```diff
@@ -10,26 +10,40 @@ import type {
...
+ *
+ * Each generator is independent and always starts from the same seed, so the data
+ * produced by the functions below depends only on their arguments. A single shared
+ * generator would instead make output depend on how many numbers earlier callers had
+ * already drawn, which is not reproducible: Storybook renders every story into one
+ * page, so the draw count varies with story order, sharding and re-renders.
+ */
+const createRandom = (): (() => number) => {
+    const math = create(all, { randomSeed: RANDOM_SEED });
+    return math?.random ? math.random : Math.random;
```
w1nklr: return math?.random ?? Math.random;

YCR

w1nklr (approved):

typescript/packages/subsurface-viewer/src/storybook/util/wellSynthesis.ts
hkfb: Applied — extracted into `util/random.ts`, which uses `math?.random ?? Math.random`.

